PR #14, opened 2023-02-07
Address cluster state

bardielle:
In this PR I added the ability to track cluster state in ocm_cluster_rosa_classic resource - in create and destroy, by polling the cluster state. 
The user can set a new attribute called `wait_with_timeout` (in minutes) and the Create and Destroy method waits for cluster state in that same timeout.

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -45,6 +45,7 @@ const (
 	MinVersion                = "4.10"
 	nonPositiveTimeoutSummary = "Can't poll cluster state with a non-positive timeout"
 	nonPositiveTimeoutFormat  = "Can't poll state of cluster with identifier '%s', the timeout that was set is not a positive number"
+	pollingIntervalInMinutes  = 1
 )
 
 var kmsArnRE = regexp.MustCompile(
@@ -671,7 +672,7 @@ func (r *ClusterRosaClassicResource) waitForClusterStateWithTimeout(ctx context.
 	pollCtx, cancel := context.WithTimeout(ctx, timeoutInMinutes)
 	defer cancel()
 	_, err := r.collection.Cluster(object.ID()).Poll().
-		Interval(1 * time.Minute).
+		Interval(pollingIntervalInMinutes * time.Minute).
 		Predicate(func(getClusterResponse *cmv1.ClusterGetResponse) bool {
 			object = getClusterResponse.Body()
 			logger.Debug(ctx, "cluster state is %s", object.State())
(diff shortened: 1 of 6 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -616,12 +623,59 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context,
...
+				"Received error %v", err,
+			),
+		)
+		return
+	}
 	diags = response.State.Set(ctx, state)
 	response.Diagnostics.Append(diags...)
 }
 
+func (r *ClusterRosaClassicResource) waitTillClusterIsReady(ctx context.Context, WaitWithTimeout int64, object *cmv1.Cluster, logger logging.Logger) error {
```
nirarg, line 657: Need to add unit tests to validate this new functionality

provider/cluster_rosa_classic_resource.go
```diff
@@ -616,12 +623,59 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context,
...
+	ready := object.State() == cmv1.ClusterStateReady
+	timeoutInMinutes := time.Duration(WaitWithTimeout) * time.Minute
+	if WaitWithTimeout > 0 && !ready {
```
nirarg, line 660: I don't think it necessary to check the state here, it would be verified in the poll
Also, in case `WaitWithTimeout <=0`, caller might conclude the state is ready while its not

provider/cluster_rosa_classic_resource.go
```diff
@@ -616,12 +623,59 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context,
...
+		pollCtx, cancel := context.WithTimeout(ctx, timeoutInMinutes)
+		defer cancel()
+		_, err := r.collection.Cluster(object.ID()).Poll().
+			Interval(1 * time.Minute).
```
nirarg, line 664: Should this be configured, not hard coded 1 minute?

provider/cluster_rosa_classic_resource.go
```diff
@@ -773,10 +845,46 @@ func (r *ClusterRosaClassicResource) Delete(ctx context.Context, request tfsdk.D
...
+			)
+			return
+		}
+	}
+
 	// Remove the state:
 	response.State.RemoveResource(ctx)
 }
 
+func (r *ClusterRosaClassicResource) waitTillClusterIsNotFound(ctx context.Context, WaitWithTimeout int64, resource *cmv1.ClusterClient, logger logging.Logger) error {
```
nirarg, line 867: Can this and `waitTillClusterIsReady` merged to one function?
Need to add unit tests

provider/cluster_rosa_classic_resource.go
bardielle: I added subsystem testes

provider/cluster_rosa_classic_resource.go
bardielle: changed it

provider/cluster_rosa_classic_resource.go
bardielle: As we spoke - not really, those two functions use different ocm polling methods- the "create" is using a predicate and the "delete" is using status which checks the status code that returns from the "get" API

nirarg (approved):

PR #16, opened 2023-02-09
Small linting fixes

gdbranco:


provider/machine_pool_resource.go
```diff
@@ -326,7 +327,6 @@ func (r *MachinePoolResource) Update(ctx context.Context, request tfsdk.UpdateRe
 		)
 		return
 	}
-	object := get.Body()
```
gdbranco, line 329: Wasn't being in use apparently. Please correct me if I'm mistaken

bardielle: you're right
Also please remove line 348 (that it's comment out)

bardielle (approved):

PR #21, opened 2023-02-14
Add a new resource ocm cluster wait

bardielle:


provider/cluster_rosa_classic_resource.go
```diff
@@ -855,59 +811,41 @@ func (r *ClusterRosaClassicResource) Delete(ctx context.Context, request tfsdk.D
...
-				response.Diagnostics.AddError(
-					"Can't poll cluster state",
-					fmt.Sprintf(
-						"Can't poll state of cluster with identifier '%s': %v",
-						state.ID.Value, err,
-					),
-				)
-				return
+	if !state.DisableWaitingInDestroy.Unknown && !state.DisableWaitingInDestroy.Null && state.DisableWaitingInDestroy.Value {
+		r.logger.Info(ctx, "Disable waiting")
```
nirarg, line 815: This log is not clear, write something like "Wait for destroy complete is disabled"

provider/cluster_waiter_resource.go
```diff
@@ -0,0 +1,164 @@
...
+	diags = response.State.Set(ctx, state)
+	response.Diagnostics.Append(diags...)
+}
+
+func (r *ClusterWaiterResource) Read(ctx context.Context, request tfsdk.ReadResourceRequest,
+	response *tfsdk.ReadResourceResponse) {
+	// Do Nothing
+}
+
+func (r *ClusterWaiterResource) Update(ctx context.Context, request tfsdk.UpdateResourceRequest,
```
nirarg, line 130: Shouldn't update also have the same logic as Create? 
What happen if the cluster is not ready when update is performed?

provider/cluster_rosa_classic_resource.go
```diff
@@ -338,16 +335,21 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result
 					ValueCannotBeChangedModifier(t.logger),
 				},
 			},
+			"disable_waiting_in_destroy": {
```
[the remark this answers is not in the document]
bardielle: but the default is waiting...

provider/cluster_waiter_resource.go
bardielle: The waiting is only in destroy..

bardielle: I thought you wanted me to add an update support for the *new* resource, but I think you talked about cluster update - so in cluster update, it will not affect that resource at all - while updating a cluster, the resource cluster_waiter will be successfully created...

nirarg (approved):

PR #28, opened 2023-02-25
Add domain attribute to cluster_rosa_classic resource

jjulien:
The `dns.base_domain` attribute is immediately available from the API when creating a new cluster, and can be useful in further configuration of resources within the same terraform run.

This value can be queried using `rosa describe cluster -c id`. The value from the ROSA CLI is always `[name].[dns.base_domain]`, which is why the domain attribute prepends the cluster name in this proposed feature.

An example use would be configuring an OpenID IdP using the domain attribute for the callback URL.

This creates a nice workflow where the cluster creation can begin in the background (required already because of the OIDC Provider dependency), and the `ocm_identity_provider` of type OpenID can be created in the same terraform run, which will trigger the wait for cluster install completion loop on `ocm_identity_provider`.

The result is a single terraform run for a ROSA cluster build, that does not exit until the cluster is ready for use, with SSO login enabled.

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -180,6 +180,11 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result 
 				Type:        types.StringType,
 				Computed:    true,
 			},
+			"domain": {
+				Description: "DNS Domain of Cluster",
+				Type:        types.StringType,
+				Computed:    true,
+			},
 			"replicas": {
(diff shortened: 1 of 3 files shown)


jjulien:
@arendej @bardielle Please review.

bardielle:
@gdbranco @oriAdler please review it

bardielle (commented):
/lgtm

bardielle:
@zgalor Can you also review it?

zgalor:
LGTM

bardielle (approved):

PR #37, opened 2023-03-26
Bug fix [SDA-8602] Exit when the cluster encounters an error state, w…

bardielle:
…hile waiting for the cluster creation process to complete


provider/cluster_waiter_resource.go
```diff
@@ -153,7 +153,7 @@ func (r *ClusterWaiterResource) isClusterReady(clusterId string, ctx context.Con
 		Predicate(func(getClusterResponse *cmv1.ClusterGetResponse) bool {
 			object = getClusterResponse.Body()
 			r.logger.Debug(ctx, "cluster state is %s", object.State())
-			return object.State() == cmv1.ClusterStateReady
+			return object.State() == cmv1.ClusterStateReady || object.State() == cmv1.ClusterStateError
```
nirarg, line 156:
```suggestion
switch object.State() {
	case  cmv1.ClusterStateReady,
	      cmv1.ClusterStateError:
		return true;
}
return false;
```
I think this way is more clear

nirarg (approved):

PR #42, opened 2023-04-03
Add account role version compatibility validation as part of cluster …

bardielle:
…creation


provider/cluster_rosa_classic_resource.go
```diff
@@ -623,6 +643,154 @@ func createClassicClusterObject(ctx context.Context,
...
+}
+func hasCompatibleVersionTags(logger logging.Logger, ctx context.Context, iamTags []*iam.Tag, version string) (bool, error) {
+	if len(iamTags) == 0 {
+		return false, nil
+	}
+	for _, tag := range iamTags {
+		if aws.StringValue(tag.Key) == tagsOpenShiftVersion {
+			logger.Debug(ctx, "role version is %s", aws.StringValue(tag.Value))
+			if version == aws.StringValue(tag.Value) {
+				return true, nil
```
oriAdler, line 701: Why do we return here `true`?

provider/cluster_rosa_classic_resource.go
```diff
@@ -644,6 +812,64 @@ func isByoOidcSet(sts *Sts) bool {
...
+	}
+
+	if len(versionList) == 0 {
+		err = fmt.Errorf("Could not find versions")
+		return
+	}
+
+	return
+}
+func (r *ClusterRosaClassicResource) getVersions(logger logging.Logger, ctx context.Context) (versions []*cmv1.Version, err error) {
```
oriAdler, line 833: I think this is outdated.
https://github.com/openshift/rosa/blob/master/pkg/ocm/versions.go#L36 - we select now the **latest** as the default.

provider/cluster_rosa_classic_resource.go
```diff
@@ -623,6 +643,154 @@ func createClassicClusterObject(ctx context.Context,
...
+			}
+			wantedVersion, err := semver.NewVersion(version)
+			if err != nil {
+				return false, err
+			}
+			currentVersion, err := semver.NewVersion(aws.StringValue(tag.Value))
+			if err != nil {
+				return false, err
+			}
+			return currentVersion.GreaterThanOrEqual(wantedVersion), nil
```
nirarg, line 711: Having this verification, why the verification in line 700 also needed? doesn't it cover all options here?

provider/cluster_rosa_classic_resource.go
```diff
@@ -623,6 +643,154 @@ func createClassicClusterObject(ctx context.Context,
...
+			},
+		},
+	})
+
+	sess.Handlers.Build.PushBackNamed(addTerraformProviderVersionToUserAgent)
+
+	_, err = sess.Config.Credentials.Get()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to find credentials. Check your AWS configuration and try again")
+	}
```
nirarg, line 768:
```suggestion
	if _, err = sess.Config.Credentials.Get(); err != nil {
		return nil, fmt.Errorf("Failed to find credentials. Check your AWS configuration and try again")
	}
```

provider/cluster_rosa_classic_resource.go
bardielle: I updated  it 
thanks

provider/cluster_rosa_classic_resource.go
bardielle: The same as you're doing in rosa cli: https://github.com/openshift/rosa/blob/master/pkg/aws/policies.go#L413
It means that both of them are holding the same tag version

provider/cluster_rosa_classic_resource.go
```diff
@@ -44,19 +52,27 @@ const (
...
 type ClusterRosaClassicResourceType struct {
 	logger logging.Logger
 }
 
 type ClusterRosaClassicResource struct {
-	logger     logging.Logger
-	collection *cmv1.ClustersClient
+	logger            logging.Logger
+	clusterCollection *cmv1.ClustersClient
+	versionCollection *cmv1.VersionsClient
```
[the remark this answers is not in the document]
bardielle: I just wanted to be consistent with OCM terminology (and also other resources that were defined in that provider) 
you can see that  it is common to call it collection here also in rosa cli - https://github.com/openshift/rosa/blob/master/pkg/ocm/versions.go#L37

provider/cluster_rosa_classic_resource.go
bardielle: I copied it from rosa cli - https://github.com/openshift/rosa/blob/master/pkg/aws/policies.go#L434
It's more efficiant to add that if than converting it if no need

PR #53, opened 2023-05-04
OCM-174: Added GitHub IDP provider attrs

sagidayan:
- Created new `idps` package. In future PRs will move all the providers
to this package.

- Added ability to create a github idp.

- Added an attribute validator helper struct for easy validation
construction.

- Added subsytem tests

Validations are based on ROSA CLI

### DIFF
--- a/provider/identity_provider_resource.go
+++ b/provider/identity_provider_resource.go
@@ -19,8 +19,8 @@ package provider
 import (
 	"context"
 	"fmt"
-	"time"
 	"net/url"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
@@ -28,6 +28,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift-online/ocm-sdk-go/logging"
+	"github.com/terraform-redhat/terraform-provider-ocm/provider/common"
+	"github.com/terraform-redhat/terraform-provider-ocm/provider/idps"
 )
 
 type IdentityProviderResourceType struct {
@@ -66,7 +68,13 @@ func (t *IdentityProviderResourceType) GetSchema(ctx context.Context) (result tf
 			"gitlab": {
 				Description: "Details of the Gitlab identity provider.",
 				Attributes:  t.gitlabSchema(),
-				Optional:    true,				
+				Optional:    true,
+			},
+			"github": {
+				Description: "Details of the Github identity provider.",
+				Attributes:  idps.GithubSchema(),
+				Optional:    true,
+				Validators:  idps.GithubValidators(),
 			},
 			"ldap": {
 				Description: "Details of the LDAP identity provider.",
@@ -103,24 +111,24 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
 	return tfsdk.SingleNestedAttributes(map[string]tfsdk.Attribute{
 		"ca": {
 			Description: "Optional trusted certificate authority bundle.",
-			Type:     types.StringType,
-			Optional: true,
+			Type:        types.StringType,
+			Optional:    true,
 		},
 		"client_id": {
 			Description: "Client identifier of a registered Gitlab OAuth application.",
-			Type:     types.StringType,
-			Required: true,
+			Type:        types.StringType,
+			Required:    true,
 		},
 		"client_secret": {
 			Description: "Client secret issued by Gitlab.",
-			Type:      types.StringType,
-			Required:  true,
-			Sensitive: true,
+			Type:        types.StringType,
+			Required:    true,
+			Sensitive:   true,
 		},
 		"url": {
 			Description: "URL of the Gitlab instance.",
-			Type:     types.StringType,
-			Required: true,
+			Type:        types.StringType,
+			Required:    true,
 		},
 	})
 }
@@ -315,7 +323,7 @@ func (r *IdentityProviderResource) Create(ctx context.Context,
 			htpasswdBuilder.Password(state.HTPasswd.Password.Value)
 		}
 		builder.Htpasswd(htpasswdBuilder)
-    case state.Gitlab != nil:
+	case state.Gitlab != nil:
 		builder.Type(cmv1.IdentityProviderTypeGitlab)
 		gitlabBuilder := cmv1.NewGitlabIdentityProvider()
 		if !state.Gitlab.CA.Unknown && !state.Gitlab.CA.Null {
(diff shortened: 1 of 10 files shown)


sagidayan:
Missing subsystem tests. 
Lets hold it for now.

bardielle:
@Manuel Dewald Can you review that PR?

bardielle:
@NautiluX @vkareh can you review this PR?

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
+		"client_secret": {
+			Description: "Client secret issued by Github.",
+			Type:        types.StringType,
+			Required:    true,
+			Sensitive:   true,
 		},
 	})
 }
 
+func (t *IdentityProviderResourceType) githubValidators() []tfsdk.AttributeValidator {
```
bardielle, line 175: I think that the method no need to be part of IdentityProviderResourceType struct
Later we will remove it to a different package.

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
 			Description: "URL of the Gitlab instance.",
-			Type:     types.StringType,
-			Required: true,
+			Type:        types.StringType,
+			Required:    true,
+		},
+	})
+}
+
+func (t *IdentityProviderResourceType) githubSchema() tfsdk.NestedAttributes {
```
bardielle, line 135: this method need to be outside from IdentityProviderResourceType struct

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
+		},
+		"teams": {
+			Description: "Only users that are members of at least one of the listed teams will be allowed to log in. The format is <org>/<team>.",
+			Type: types.ListType{
+				ElemType: types.StringType,
+			},
+			Optional: true,
+		},
+		"ca": {
+			Type:        types.StringType,
```
bardielle, line 157: I preffer to keep the struct - first Description and then the type

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
+			Description: "Path to PEM-encoded certificate file to use when making requests to the server.",
+			Optional:    true,
+		},
+		"client_id": {
+			Description: "Client identifier of a registered Github OAuth application.",
+			Type:        types.StringType,
+			Required:    true,
```
bardielle, line 164: insert the required attributes to be on top of the struct

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
+		},
+		"client_secret": {
```
bardielle, line 166: insert the required attributes to be on top of the struct

provider/identity_provider_resource.go
```diff
@@ -103,28 +110,138 @@ func (t *IdentityProviderResourceType) gitlabSchema() tfsdk.NestedAttributes {
...
+			mdDescription: "",
+			validate: func(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+				ghState := &GithubIdentityProvider{}
+				diag := req.Config.GetAttribute(ctx, req.AttributePath, ghState)
+				if diag.HasError() {
+					// No attribute to validate
+					return
+				}
+				// Validate hostname
+				if !ghState.Hostname.Unknown && !ghState.Hostname.Null && len(ghState.Hostname.Value) > 0 {
```
bardielle, line 231: ghState.Hostname is from Type:        types.StringType, I think the validation is incorrect

provider/identity_provider_resource.go
```diff
@@ -336,6 +453,25 @@ func (r *IdentityProviderResource) Create(ctx context.Context,
...
+		ghState := state.Github
+		githubBuilder.ClientID(ghState.ClientID.Value)
+		githubBuilder.ClientSecret(ghState.ClientSecret.Value)
+		if !ghState.CA.Unknown && !ghState.CA.Null {
+			githubBuilder.CA(ghState.CA.Value)
+		}
+		if !ghState.Hostname.Unknown && !ghState.Hostname.Null && len(ghState.Hostname.Value) > 0 {
+			githubBuilder.Hostname(ghState.Hostname.Value)
+		}
+		if len(ghState.Teams) > 0 {
```
bardielle, line 468: teams is Optional: true,
you should check  if it's not unknown and not nil

provider/identity_provider_resource.go
```diff
@@ -507,6 +643,7 @@ func (r *IdentityProviderResource) Read(ctx context.Context, request tfsdk.ReadR
 	gitlabObject := object.Gitlab()
 	ldapObject := object.LDAP()
 	openidObject := object.OpenID()
+	githubObject := object.Github()
```
bardielle, line 646: Github -> GitHub

subsystem/identity_provider_resource_test.go
```diff
@@ -131,6 +131,55 @@ var _ = Describe("Identity provider creation", func() {
 		Expect(terraform.Apply()).To(BeZero())
 	})
 
+	It("Can create a 'github' identity provider", func() {
```
bardielle, line 134: please add tests that check 
* assignment for organizations and teams
* teams without the convention

provider/identity_provider_resource.go
sagidayan: This is the exported method name coming from github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1

provider/identity_provider_resource.go
sagidayan: Will be moved to the new `idps` package

provider/identity_provider_resource.go
sagidayan: Followed the same structure as the others - but you are right. Will be moved to the new `idps` package

provider/identity_provider_resource.go
sagidayan: Will have the same result, an empty string, is of length 0. I can change it if you prefer this way of checking.

provider/identity_provider_resource.go
sagidayan: Teams, and orgs are string arrays. Not a tf typed struct. It has no `.Unknown` nor `.Null`, also since it's not a pointer, the default is an empty array. [ref](https://github.com/terraform-redhat/terraform-provider-ocm/pull/53/files#diff-6d537d29d94410c13a4ae530df04cbf7ab877d0a3c1ff94a5fe86850a96e2bb6R51). In this case, Optional is only a validation. 

Tests can validate this as well.

bardielle: Please add a subsystem test that covers it.

provider/idps/attribute_validator.go
```diff
@@ -0,0 +1,24 @@
+package idps
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+)
+
+type attributeValidator struct {
```
bardielle, line 8: Maybe this struct can be shared with other resources?  
If no change the name to include IDP

bardielle (approved):

PR #56, opened 2023-05-08
OCM-177: Move all identity providers to internal pkg

sagidayan:
All IDPs are now located in /provider/idps/

this includes:
 - schemas
 - structs (states)
 - validations (if exists)
 - updated golang string slices to terraform types.List
 - Added missing descriptions + updated docs
 - provider builders

### DIFF
--- a/provider/idps/ldap.go
+++ b/provider/idps/ldap.go
@@ -1,0 +1,148 @@
+package idps
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/terraform-redhat/terraform-provider-ocm/provider/common"
+)
+
+type LDAPIdentityProvider struct {
+	BindDN       types.String                    `tfsdk:"bind_dn"`
+	BindPassword types.String                    `tfsdk:"bind_password"`
+	CA           types.String                    `tfsdk:"ca"`
+	Insecure     types.Bool                      `tfsdk:"insecure"`
+	URL          types.String                    `tfsdk:"url"`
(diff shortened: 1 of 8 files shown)


provider/idps/ldap.go
```diff
@@ -0,0 +1,123 @@
...
+	BindDN       types.String                    `tfsdk:"bind_dn"`
+	BindPassword types.String                    `tfsdk:"bind_password"`
+	CA           types.String                    `tfsdk:"ca"`
+	Insecure     types.Bool                      `tfsdk:"insecure"`
+	URL          types.String                    `tfsdk:"url"`
+	Attributes   *LDAPIdentityProviderAttributes `tfsdk:"attributes"`
+}
+
+type LDAPIdentityProviderAttributes struct {
+	EMail             []string `tfsdk:"email"`
```
bardielle, line 21: Can you update the type to be types.List?

provider/idps/ldap.go
```diff
@@ -0,0 +1,123 @@
...
+	ID                []string `tfsdk:"id"`
+	Name              []string `tfsdk:"name"`
+	PreferredUsername []string `tfsdk:"preferred_username"`
+}
+
+func LdapSchema() tfsdk.NestedAttributes {
+	return tfsdk.SingleNestedAttributes(map[string]tfsdk.Attribute{
```
bardielle, line 28: I know you just copied it 
but the description is missing here...

bardielle (commented):
Nice job! 
I had some comments

sagidayan:
/hold

Will update according to @bardielle comments

provider/idps/ldap.go
sagidayan: Done

provider/idps/ldap.go
sagidayan: Done

provider/idps/ldap.go
bardielle: I think you also need to change the relevant code that set those values

bardielle (approved):
/lgtm

PR #57, opened 2023-05-08
OCM-136: Allow specifying channel group name

JohnStrunk:
- [x] Add a new attribute
- [x] Validate the cluster version including the channel group
- [x] Validate the version of the account roles is compatible to that version
- [x] Send the ChannelGroup as part of the object NewCluster

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/terraform-redhat/terraform-provider-ocm/provider/common"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
@@ -349,6 +350,15 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result 
 					ValueCannotBeChangedModifier(t.logger),
 				},
 			},
+			"channel_group": {
+				Description: "Name of the channel group from which to select the OpenShift cluster version, for example 'stable'.",
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					ValueCannotBeChangedModifier(t.logger),
+				},
+			},
 			"version": {
 				Description: "Identifier of the version of OpenShift, for example 'openshift-v4.1.0'.",
 				Type:        types.StringType,
@@ -600,9 +610,15 @@ func createClassicClusterObject(ctx context.Context,
 	if !network.Empty() {
 		builder.Network(network)
 	}
+
+	channelGroup := ocm.DefaultChannelGroup
+	if !state.ChannelGroup.Unknown && !state.ChannelGroup.Null {
+		channelGroup = state.ChannelGroup.Value
+	}
+
 	if !state.Version.Unknown && !state.Version.Null {
 		// TODO: update it to support all cluster versions
-		isSupported, err := checkSupportedVersion(state.Version.Value)
+		isSupported, err := checkMinSupportedVersion(state.Version.Value)
 		if err != nil {
 			logger.Error(ctx, "Error validating required cluster version %s\", err)")
 			errDecription := fmt.Sprintf(
@@ -615,9 +631,7 @@ func createClassicClusterObject(ctx context.Context,
 			)
 			return nil, errors.New(errHeadline + "\n" + errDecription)
 		}
-		if isSupported {
-			builder.Version(cmv1.NewVersion().ID(state.Version.Value))
-		} else {
+		if !isSupported {
 			logger.Error(ctx, "Cluster version %s is not supported", state.Version.Value)
 			errDecription := fmt.Sprintf(
 				"Can't check if cluster version is supported '%s': %v",
@@ -629,6 +643,17 @@ func createClassicClusterObject(ctx context.Context,
 			)
 			return nil, errors.New(errHeadline + "\n" + errDecription)
 		}
+		vBuilder := cmv1.NewVersion()
+		versionID := state.Version.Value
+		// When using a channel group other than the default, the channel name
+		// must be appended to the version ID or the API server will return an
+		// error stating unexpected channel group.
+		if channelGroup != ocm.DefaultChannelGroup {
+			versionID = versionID + "-" + channelGroup
+		}
+		vBuilder.ID(versionID)
+		vBuilder.ChannelGroup(channelGroup)
+		builder.Version(vBuilder)
 	}
 
 	proxy := cmv1.NewProxy()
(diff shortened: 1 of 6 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -645,20 +670,40 @@ func createClassicClusterObject(ctx context.Context,
...
-		version = versionList[0]
+	}
+
+	return "", fmt.Errorf("version %s is not in the list of supported versions: %v", version, versionList)
+}
+
+func (r *ClusterRosaClassicResource) validateAccountRoles(ctx context.Context, state *ClusterRosaClassicState) error {
+	r.logger.Debug(ctx, "Validating if cluster version is compatible to account roles' version")
+	region := state.CloudRegion.Value
+	version, err := r.getAndValidateVersion(ctx, state)
```
JohnStrunk, line 704: I'm not particularly pleased at putting the version validation inside the account role validation function, but this code already made the calls to retrieve the version list, so throwing the check in here was the easiest place.
As an alternative, I could update things to memoize the version list and check it separately.

provider/cluster_rosa_classic_resource.go
```diff
@@ -813,10 +860,14 @@ func (r *ClusterRosaClassicResource) getVersionList(logger logging.Logger, ctx c
...
 }
-func (r *ClusterRosaClassicResource) getVersions(logger logging.Logger, ctx context.Context) (versions []*cmv1.Version, err error) {
+func (r *ClusterRosaClassicResource) getVersions(logger logging.Logger, ctx context.Context, channelGroup string) (versions []*cmv1.Version, err error) {
 	page := 1
 	size := 100
-	filter := "enabled = 'true' AND rosa_enabled = 'true'"
+	filter := strings.Join([]string{
+		"enabled = 'true'",
+		"rosa_enabled = 'true'",
+		fmt.Sprintf("channel_group = '%s'", channelGroup),
```
JohnStrunk, line 869: The channel group is now always provided in the query, even when on `stable`. This may clear up an issue I was having when getting started where versions that aren't actually available (in stable) can be seen.

provider/cluster_rosa_classic_resource.go
```diff
@@ -1414,7 +1465,20 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
...
+		}
+	} else {
+		state.ChannelGroup = types.String{
+			Null: true,
+		}
+	}
 	version, ok := object.Version().GetID()
+	// If we're using a non-default channel group, it will have been appended to
+	// the version ID. Remove it before saving state.
+	version = strings.TrimSuffix(version, fmt.Sprintf("-%s", channel_group))
```
JohnStrunk, line 1481: I found the version naming scheme of the provider to be a bit confusing... We require versions to be specified as `openshift-vX.Y.Z`, but rosa cli and ocm cli don't. Additionally, when requesting a version from a non-default channel, the channel name needs to be appended.
I chose to make that behavior transparent in this patch by auto-appending and removing as necessary instead of requiring `openshift-vX.Y.Z-channel` to be provided.

provider/cluster_rosa_classic_resource.go
```diff
@@ -645,20 +670,40 @@ func createClassicClusterObject(ctx context.Context,
 	return object, err
 }
 
-func (r *ClusterRosaClassicResource) validateAccountRoles(ctx context.Context, state *ClusterRosaClassicState) error {
-	r.logger.Debug(ctx, "Validating if cluster version is compatible to account roles' version")
-	region := state.CloudRegion.Value
-	version := ""
+// getAndValidateVersion ensures that the cluster version is available in the
+// channel group
+func (r *ClusterRosaClassicResource) getAndValidateVersion(ctx context.Context, state *ClusterRosaClassicState) (string, error) {
```
bardielle, line 675: The function name might be confusing
Maybe change the function name to something like validateClusterVersionCompatibleWithChannelGroup... what do you think?

provider/cluster_rosa_classic_resource.go
```diff
@@ -629,6 +643,17 @@ func createClassicClusterObject(ctx context.Context,
 			)
 			return nil, errors.New(errHeadline + "\n" + errDecription)
 		}
+		vBuilder := cmv1.NewVersion()
```
bardielle, line 646: I'm not sure - but maybe we need to validate that the selected version is compatible with the channel group.

provider/cluster_rosa_classic_resource.go
bardielle: This function could also help us if the cluster version and channel group were set. (in the create)

provider/cluster_rosa_classic_resource.go
bardielle: So I think better we will be aligned with ROSA CLI and removing that suffix from the version

provider/cluster_rosa_classic_resource_test.go
```diff
@@ -204,6 +214,20 @@ var _ = Describe("Rosa Classic Sts cluster", func() {
 		Expect(err).ToNot(BeNil())
 	})
 
+	It("appends the non-default channel name to the requested version", func() {
```
bardielle, line 217: Please add a subsystem test

bardielle:
@gdbranco can you review that PR?

provider/cluster_rosa_classic_resource.go
JohnStrunk: Unless I'm missing it, we don't have access to the api server from this function... I actually wanted to do the validation here, but couldn't.
However, `createClassicClusterObject` is only called from Create(), right after the verification of account roles... and the account role verification flow is where we check the version/channel. In effect, it does get checked, just not right here.

provider/cluster_rosa_classic_resource.go
```diff
@@ -349,6 +350,15 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result
 					ValueCannotBeChangedModifier(t.logger),
 				},
 			},
+			"channel_group": {
+				Description: "Name of the channel group from which to select the OpenShift cluster version, for example 'stable'.",
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					ValueCannotBeChangedModifier(t.logger),
```
gdbranco, line 359: There are ways to change the clusters channel group btw, I have never done it though

provider/cluster_rosa_classic_resource.go
JohnStrunk: renamed

provider/cluster_rosa_classic_resource.go
JohnStrunk: Should I allow it to be modified?

provider/cluster_rosa_classic_resource.go
bardielle: you're right
the validations are done in the function "Create"- https://github.com/terraform-redhat/terraform-provider-ocm/blob/8c32bac577336a1d2ca868740c72656000b4bda3/provider/cluster_rosa_classic_resource.go#L903

bardielle: That it's because the UTs are running against the `createClassicClusterObject` function

provider/cluster_rosa_classic_resource.go
bardielle: I think better not to support it...

bardielle (approved):

PR #59, opened 2023-05-09
OCM-201: Add support for mapping_method

sagidayan:
Added mapping_method support.
 - Validation
 - idp create
 - idp read
 - subsytem tests

### DIFF
--- a/provider/identity_provider_resource.go
+++ b/provider/identity_provider_resource.go
@@ -59,6 +59,13 @@ func (t *IdentityProviderResourceType) GetSchema(ctx context.Context) (result tf
 				Type:        types.StringType,
 				Required:    true,
 			},
+			"mapping_method": {
+				Description: "Specifies how new identities are mapped to users when they log in. Options are [add claim generate lookup] (default 'claim')",
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				Validators:  idps.MappingMethodValidators(),
+			},
 			"htpasswd": {
 				Description: "Details of the 'htpasswd' identity provider.",
 				Attributes:  idps.HtpasswdSchema(),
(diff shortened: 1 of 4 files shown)


gdbranco:
Considering ROSA CLI flow SGTM

provider/identity_provider_resource.go
```diff
@@ -59,6 +62,36 @@ func (t *IdentityProviderResourceType) GetSchema(ctx context.Context) (result tf
...
 				Required:    true,
 			},
+			"mapping_method": {
+				Description: "Specifies how new identities are mapped to users when they log in. Options are [add claim generate lookup] (default 'claim')",
+				Type:        types.StringType,
+				Optional:    true,
+				Validators: []tfsdk.AttributeValidator{
+					&common.AttributeValidator{
+						Desc: "Validate mapping_method",
+						Validator: func(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
```
bardielle, line 72: I think it's better to move that function into idp package....

provider/identity_provider_resource.go
```diff
@@ -139,10 +172,16 @@ func (r *IdentityProviderResource) Create(ctx context.Context,
 		)
 		return
 	}
+	// handle mapping_method
+	if state.MappingMethod.Unknown || state.MappingMethod.Null {
+		// Set default Value
+		state.MappingMethod.Value = defaultMappingMethod
```
bardielle, line 178: you can't do it
this is an optional only attribute
If you will set it here you change the value of that attribute.... you have 2 options: 
 1. change it to be also computed
 2. no setting the value if it wasn't set

provider/identity_provider_resource.go
```diff
@@ -270,6 +309,10 @@ func (r *IdentityProviderResource) Read(ctx context.Context, request tfsdk.ReadR
 	state.Name = types.String{
 		Value: object.Name(),
 	}
+	state.MappingMethod = types.String{
```
bardielle, line 312: you can't do it (the same comment as before....)

oriAdler:
LGTM

PR #66, opened 2023-05-15
OCM-141: Terraform for IMDSv2 support

tsorya:
Adding support for http_tokens_state field. This field can be used only
    from 4.11.0 and have 2 options: optional/required

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -53,10 +53,11 @@ import (
 const (
 	awsCloudProvider     = "aws"
 	rosaProduct          = "rosa"
-	MinVersion           = "4.10"
+	MinVersion           = "4.10.0"
 	maxClusterNameLength = 15
 	tagsPrefix           = "rosa_"
 	tagsOpenShiftVersion = tagsPrefix + "openshift_version"
+	lowestHttpTokensVer  = "4.11.0"
 )
 
 var kmsArnRE = regexp.MustCompile(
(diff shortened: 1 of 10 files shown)


sagidayan:
SGTM, @bardielle Can you also take a look?

provider/cluster_rosa_classic_resource.go
```diff
@@ -57,6 +57,7 @@ const (
 	maxClusterNameLength = 15
 	tagsPrefix           = "rosa_"
 	tagsOpenShiftVersion = tagsPrefix + "openshift_version"
+	lowestHttpTokensVer  = "4.11.0"
```
bardielle, line 60: What if the cluster version will be set to 4.10? (the MinVersion of cluster version is "4.10")

provider/cluster_rosa_classic_resource.go
```diff
@@ -698,14 +715,31 @@ func (r *ClusterRosaClassicResource) getAndValidateVersionInChannelGroup(ctx con
...
+		return nil
+	}
+
+	version = getOcmVersionMinor(version)
+	a, err := ver.NewVersion(version)
 	if err != nil {
-		return fmt.Errorf("Could not get cluster version: %v", err)
+		return fmt.Errorf("version '%s' is not supported: %v", version, err)
+	}
+	b, _ := ver.NewVersion(lowestHttpTokensVer)
```
bardielle, line 728: we already have a common logic in the function checkMinSupportedVersion
Can you reuse it? or adapting that function to be compatible with your validation?

provider/cluster_rosa_classic_resource.go
bardielle: I just thought we would like to be aligned with the minimum version 
But I think it's fine to handle it like that

bardielle (approved):
/lgtm

PR #67, opened 2023-05-15
Fix e2e flow missing variables

osherdp:
Some variables were missing or given with the wrong names.
In addition, remove junit.xml reporting as we don't really use it right now (and it's causing some permission problems in prow)


osherdp:
/cc @AsherShoshan 
let's test it with your PR in openshift/release

PR #70, opened 2023-05-16
Fixing issue with empty aws_http_tokens in ocm response

tsorya:
It fixed e2e test error
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to ocm_cluster_rosa_classic.rosa_sts_cluster,
│ provider "provider[\"terraform.local/local/ocm\"]" produced an unexpected
│ new value: .aws_http_tokens_state: was null, but now cty.StringVal("").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵


osherdp:
/retitle Fixing issue with empty aws_http_tokens in ocm response

osherdp (approved):

PR #72, opened 2023-05-17
OCM-335: Added provider version and commit to cluster properties

sagidayan:
- Added new computed attribute, `ocm_properties` containing provider version/commit
 - when creating a cluster we merge `properties` and `ocm_properties` into one map that is sent to OCM. This is so we keep `properties` backwards compatible.

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -34,6 +34,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/terraform-redhat/terraform-provider-ocm/build"
 	"github.com/terraform-redhat/terraform-provider-ocm/provider/common"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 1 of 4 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -451,13 +452,17 @@ func createClassicClusterObject(ctx context.Context,
 	if !state.MultiAZ.Unknown && !state.MultiAZ.Null {
 		builder.MultiAZ(state.MultiAZ.Value)
 	}
+	// Set default properties
+	properties := map[string]string{
+		"rosa_tf_version": build.Version,
```
bardielle, line 457: Please use costs for those strings

provider/cluster_rosa_classic_resource_test.go
```diff
@@ -245,6 +248,8 @@ var _ = Describe("Rosa Classic Sts cluster", func() {
 			Expect(clusterState.CloudRegion.Value).To(Equal(regionId))
 			Expect(clusterState.MultiAZ.Value).To(Equal(multiAz))
 			Expect(clusterState.Properties.Elems["rosa_creator_arn"].Equal(types.String{Value: rosaCreatorArn})).To(Equal(true))
+			Expect(clusterState.Properties.Elems["rosa_tf_version"].Equal(types.String{Value: build.Version})).To(Equal(true))
```
bardielle, line 251: Please add it also in the subsystem test - check that the resource holds those properties `Expect(resource).To(MatchJQ(attributes.properties....) ) `

subsystem/cluster_resource_rosa_test.go
```diff
@@ -349,6 +356,151 @@ var _ = Describe("Cluster creation", func() {
...
+            properties = { 
+             rosa_tf_version = "bob"
+			}
+			sts = {
+				operator_role_prefix = "test"
+				role_arn = "",
+				support_role_arn = "",
+				instance_iam_roles = {
+					master_role_arn = "",
+					worker_role_arn = "",
```
osherdp, line 497: is it just me or something is weird with the indentations here?

subsystem/cluster_resource_rosa_test.go
```diff
@@ -349,6 +356,151 @@ var _ = Describe("Cluster creation", func() {
...
+				}
+			}
+		  }
+		`)
+		Expect(terraform.Apply()).ToNot(BeZero())
```
osherdp, line 502: can we validate the returned message as well?
just to differentiate it from other possible errors/failures

provider/cluster_rosa_classic_resource.go
```diff
@@ -51,15 +52,22 @@ import (
...
+	MinVersion            = "4.10.0"
+	maxClusterNameLength  = 15
+	tagsPrefix            = "rosa_"
+	tagsOpenShiftVersion  = tagsPrefix + "openshift_version"
+	lowestHttpTokensVer   = "4.11.0"
+	propertyRosaTfVersion = tagsPrefix + "tf_version"
+	propertyRosaTfCommit  = tagsPrefix + "tf_commit"
 )
 
+var ocm_properties = map[string]string{
```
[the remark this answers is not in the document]
sagidayan: You are right

subsystem/cluster_resource_rosa_test.go
sagidayan: This Apply returns an int. Did not see any other useful function that will return the error of the `Apply`

subsystem/cluster_resource_rosa_test.go
sagidayan: It is, fixed.

osherdp (approved):

subsystem/cluster_resource_rosa_test.go
sagidayan: Still seems off. Will fix in part 2 of OCM-335

subsystem/cluster_resource_rosa_test.go
bardielle: I also tried to find a way. even if I created a new function for that...

PR #73, opened 2023-05-17
fix account_roles to extract version from extended openshift_version

AsherShoshan:
1. In case openshift_Version is passed to cluster creation (as 4.12.15, 4.13.rc-7, etc) , the version for account_roles creation (rosa_openshift_version) can be extracted from this parameter
2. added channel variable (stable is the default)

### DIFF
--- a/ci/e2e/terraform_provider_ocm_files/main.tf
+++ b/ci/e2e/terraform_provider_ocm_files/main.tf
@@ -33,6 +33,7 @@ data "aws_caller_identity" "current" {
 
 resource "ocm_cluster_rosa_classic" "rosa_sts_cluster" {
   name               = var.cluster_name
+  version            = "openshift-v${var.openshift_version}"
   cloud_region       = var.aws_region
   aws_account_id     = data.aws_caller_identity.current.account_id
   availability_zones = var.aws_availability_zones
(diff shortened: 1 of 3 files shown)


ci/e2e/terraform_provider_ocm_files/main.tf
```diff
@@ -33,6 +33,8 @@ data "aws_caller_identity" "current" {
 
 resource "ocm_cluster_rosa_classic" "rosa_sts_cluster" {
   name               = var.cluster_name
+  version            = "openshift-v${var.openshift_version}"
```
osherdp, line 36: reading the code, you don't need ``"openshift-v${var.openshift_version}"``, just ``var.openshift_version``
see: https://github.com/terraform-redhat/terraform-provider-ocm/blob/1bc6fa6f4259897967199a497488a2194e641a46/provider/cluster_rosa_classic_resource.go#L705

osherdp (changes requested):

ci/e2e/terraform_provider_ocm_files/main.tf
AsherShoshan: ok.  hope it stays...

osherdp (approved):

PR #81, opened 2023-05-22
OCM-240  |  day2 proxy settings on cluster_rosa_classic

bardielle:


provider/cluster_rosa_classic_resource.go
```diff
@@ -693,18 +691,54 @@ func createClassicClusterObject(ctx context.Context,
 		builder.Version(vBuilder)
 	}
 
+	builder, err = buildProxy(state, builder)
```
sagidayan, line 694: nit. It might be simpler if the function would return the proxy and not update the builder. We can then do `builder.Proxy(proxy)` here. This will make the code more readable, IMO. but it works :)

provider/cluster_rosa_classic_resource.go
```diff
@@ -1178,10 +1203,72 @@ func (r *ClusterRosaClassicResource) Update(ctx context.Context, request tfsdk.U
...
+	diags = response.State.Set(ctx, plan)
 	response.Diagnostics.Append(diags...)
 }
 
+func updateProxy(state, plan *ClusterRosaClassicState, clusterBuilder *cmv1.ClusterBuilder) (*cmv1.ClusterBuilder, bool, error) {
+	shouldUpdateProxy := false
+	if (state.Proxy == nil && plan.Proxy != nil) || (state.Proxy != nil && plan.Proxy == nil) {
+		shouldUpdateProxy = true
+	}
+	if state.Proxy != nil && plan.Proxy != nil {
```
sagidayan, line 1215: We can skip this if statement when `shouldUpdateProxy` is true (from line 1213). right?

provider/cluster_rosa_classic_resource.go
bardielle: I'm doing it because not all the properties are part of the "proxy", the AdditionalTrustBundle is property of the builder

PR #86, opened 2023-05-25
Automatically prefix resources with the aws username

osherdp:
It has come to our attention that some of the resources are untagged. This makes it hard to correlate them with the initiating user, especially for cases where it's being run against a local CS env.

Here we first query the username from ``aws`` client (a preconfigured ``aws`` client is either way a requirement for using the provider) and prefixes the resources with this value.


osherdp:
/test e2e-presubmit
--


<br class="Apple-interchange-newline">

osherdp:
will get re-implemented with the new e2e code, so closing for now

PR #87, opened 2023-05-25
OCM-274: Added 2 make targets

sagidayan:
- apply_folder (./ci/apply_folder.sh)
- destroy_folder (./ci/destroy_folder.sh)

Targets can be run locally or from PROW CI.

Running locally will target the (gitignored) folder `./playground`. You
can use this to debug and play with local builds of the provider. To
change the default playground location, set `$WORK_DIR` before running
make.

If running in OpenShift CI, `$WORK_DIR` is set to
`${SHARED_DIR}/work`. And will handle archiving/extracting the
state to/from the `SHARED_DIR`. Also passing kubeconfig/cluster_id to the SHARED_DIR.

No Parameters are needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,3 +113,12 @@ e2e_test: tools install
 		--offline-token=$(test_token) \
 		--openshift-version=$(openshift_version) \
 		$(NULL)
+
+.PHONY: apply_folder
+apply_folder: install
+	bash ./ci/apply_folder.sh
+
+.PHONY: destroy_folder
+destroy_folder: install
+	bash ./ci/destroy_folder.sh
+
--- a/ci/apply_folder.sh
+++ b/ci/apply_folder.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+function error_exit() {
+    msg=${1}
+    >&2 echo "[Error] ${msg}"
+    exit 1
+}
+
+function prow_archive_state() {
+    if [ -n "${OPENSHIFT_CI:-}" ]; then
+        echo
+        echo "[INFO] Running in OpenShift CI (Prow)."
+        echo "[INFO] Archiving state for PROW."
+        echo
+        set -o xtrace
+        tar cvfz "${STATE_ARCHIVE}" ./*.tf*  # For next steps to use
+        set +o xtrace
+        echo "[INFO] ${STATE_ARCHIVE} created."
+    fi
+
+    if [ -n "${1:-}" ]; then
+        error_exit "TF command failed."
+    fi
+}
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+BASE_DIR=$(realpath "${THIS_DIR}/..")
+
+echo
+echo ">> Running apply_folder script"
+echo "   ---------------------------"
+echo
+if [ -n "${OPENSHIFT_CI:-}" ]; then
+    source "${THIS_DIR}/setup_prow_env.sh"
+fi
+
+# Setup defaults
+TERRAFORM_D_DIR=${TERRAFORM_D_DIR:-"${HOME}"}
+WORK_DIR=${WORK_DIR:-"${BASE_DIR}/playground"}
+
+echo "[INFO] Will apply folder."
+echo "[INFO] - WORK_DIR: ${WORK_DIR}"
+echo "[INFO] - TERRAFORM_D_DIR: ${TERRAFORM_D_DIR}"
+echo "[INFO] - THIS_DIR: ${THIS_DIR}"
+echo "[INFO] - BASE_DIR: ${BASE_DIR}"
+
+if [[ ! -d ${WORK_DIR} ]]; then
+    error_exit "can't load WORK_DIR"
+fi
+
+set -o xtrace
+
+cd "${WORK_DIR}"
+HOME=${TERRAFORM_D_DIR} terraform init
+HOME=${TERRAFORM_D_DIR} terraform apply -auto-approve || prow_archive_state "true"
+
+set +o xtrace
+
+prow_archive_state
+
+if [ -n "${OPENSHIFT_CI:-}" ]; then
+    source "${THIS_DIR}/post_apply_prow_setup.sh"
+fi
+
+echo
+echo "[INFO] Finished applying the terraform folder ${WORK_DIR} successfully"
+echo
--- a/ci/destroy_folder.sh
+++ b/ci/destroy_folder.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+function error_exit() {
+    msg=${1}
+    >&2 echo "[Error] ${msg}"
+    exit 1
+}
+
+function prow_archive_state() {
+    if [ -n "${OPENSHIFT_CI:-}" ]; then
+        echo
+        echo "[INFO] Running in OpenShift CI (Prow)."
+        echo "[INFO] Archiving state for PROW."
+        echo
+        set -o xtrace
+        tar cvfz "${STATE_ARCHIVE}" ./*.tf*  # For next steps to use
+        set +o xtrace
+        echo "[INFO] ${STATE_ARCHIVE} created."
+    fi
+
+    if [ -n "${1:-}" ]; then
+        error_exit "TF command failed."
+    fi
+}
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+BASE_DIR=$(realpath "${THIS_DIR}/..")
+
+echo
+echo ">> Running destroy_folder script"
+echo "   -----------------------------"
+echo
+if [ -n "${OPENSHIFT_CI:-}" ]; then
+    source "${THIS_DIR}/setup_prow_env.sh"
+fi
+
+# Setup defaults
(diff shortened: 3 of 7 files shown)


hack/apply_folder.sh
```diff
@@ -0,0 +1,62 @@
...
+
+if [[ ! -d ${WORK_DIR} ]]; then
+    error_exit "can't load WORK_DIR"
+fi
+
+set -x
+
+cd "${WORK_DIR}"
+HOME=${TERRAFORM_D_DIR} terraform init
+HOME=${TERRAFORM_D_DIR} terraform apply -auto-approve
```
osherdp, line 44: if ``terraform apply`` fails I think we won't get to the archiving part (because we're still with ``set -o errexit``)
I suggest adding this part as a ``trap``

hack/apply_folder.sh
```diff
@@ -0,0 +1,62 @@
...
+echo "[INFO] - WORK_DIR: ${WORK_DIR}"
+echo "[INFO] - TERRAFORM_D_DIR: ${TERRAFORM_D_DIR}"
+echo "[INFO] - THIS_DIR: ${THIS_DIR}"
+echo "[INFO] - BASE_DIR: ${BASE_DIR}"
+
+if [[ ! -d ${WORK_DIR} ]]; then
+    error_exit "can't load WORK_DIR"
+fi
+
+set -x
```
osherdp, line 40: for consistency with other declerations, here and elsewhere:
```suggestion
set -o xtrace
```

hack/apply_folder.sh
osherdp, line 1: let's move it under ``ci/e2e`` folder? the term ``hack`` is usually for much smaller processes

hack/apply_folder.sh
```diff
@@ -0,0 +1,62 @@
...
+
+set +x
+
+
+if [ -n "${OPENSHIFT_CI:-}" ]; then
+    echo
+    echo "[INFO] Running in OpenShift CI (Prow)."
+    echo "[INFO] Archiving state for PROW."
+    echo
+    set -x
```
osherdp, line 54: why we need this?

hack/destroy_folder.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "[INFO] Archiving state for PROW."
+    echo
+    set -x
+    tar cvfz "${STATE_ARCHIVE}" ./*.tfstate*  #save for later terraform destroy
+    set +x
+    echo "[INFO] ${STATE_ARCHIVE} created."
+fi
+
+echo
+echo "[INFO] Hack Script finished successfully"
```
osherdp, line 61:
```suggestion
echo "[INFO] Finished destroying the terraform folder successfully"
```

hack/apply_folder.sh
osherdp: also remove the word ``hack`` everywhere. it feels dirty :)

osherdp (changes requested):

hack/setup_prow_env.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# This file should be sourced in a wrapper script (i.e apply_folder.sh)
+
+echo
+echo "[INFO] Running in OpenShift CI (Prow)."
+echo "[INFO] Setting up for PROW"
+echo
+
+# Validate some mandatory variables
+if [ -z "${ART_NAME:-}" ]; then
```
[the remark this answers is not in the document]
sagidayan: In @AsherShoshan 's Prow implementation its basically the archive name, (cluster/account-roles etc.)
(I assume art->artifact)

We can change it here.

hack/setup_prow_env.sh
```diff
@@ -0,0 +1,65 @@
...
+
+rm -rf "${WORK_DIR}"
+mkdir "${WORK_DIR}"
+
+if [[ -f "${STATE_ARCHIVE}" ]]; then
+    echo "[INFO] Found TF state archive for '${ART_NAME}' @ ${STATE_ARCHIVE}"
+    tar xvfz "${SHARED_DIR}/${ART_NAME}.tar.gz" -C "${WORK_DIR}"
+fi
+
+export TERRAFORM_D_DIR=/root # location of .terraform.d folder
```
AsherShoshan, line 53: how this envar comes to play in terraform init/apply.. you need to change the $HOME in order to take the .terraform.d contents (created by docker build install)

hack/setup_prow_env.sh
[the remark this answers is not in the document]
osherdp: yes, let's call it ``ARTIFACT_NAME``

hack/setup_prow_env.sh
```diff
@@ -0,0 +1,65 @@
...
+# Validate some mandatory variables
+if [ -z "${ART_NAME:-}" ]; then
+    error_exit "missing mandatory variable \$ART_NAME"
+fi
+
+if [ -z "${TF_FOLDER:-}" ]; then
+    error_exit "missing mandatory variable \$TF_FOLDER"
+fi
+
+if [ -z "${TF_VARS:-}" ]; then
```
AsherShoshan, line 19: Why? you can apply with using default values (in variables.tf)

Makefile
```diff
@@ -113,3 +113,12 @@ e2e_test: tools install
 		--offline-token=$(test_token) \
 		--openshift-version=$(openshift_version) \
 		$(NULL)
+
+.PHONY: apply_folder
+apply_folder: install
```
AsherShoshan, line 118: Will this run the 'install' of terraform-provider?   on each apply?  it's already in image.

hack/apply_folder.sh
sagidayan: Thought about it, and found that it's confusing since now you can use this locally. So not just for CI. Maybe place these under `scripts` ?

osherdp: maybe under ``e2e``? we already have ``subsystem`` under root dir

Makefile
sagidayan: Correct. 
These targets are not just for running in a container, I prefer not to assume it has been invoked before. Also, I think it's the right thing to do here, since this target is depending on a build/install. It won't hurt.

hack/setup_prow_env.sh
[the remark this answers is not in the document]
sagidayan: Or better: `ARCHIVE_NAME`

hack/setup_prow_env.sh
sagidayan: Look at the line that executes terraform. It is passed to it.

hack/apply_folder.sh
sagidayan: We decided to place these into `/ci`

hack/setup_prow_env.sh
sagidayan: Correct. Ill update

Makefile
AsherShoshan: But every apply will recompile the ocm-provider binary?  why can't u check it's there?

ci/apply_folder.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
```
bardielle, line 2: Those bash scripts should be in a dedicated directory - for example, prow_commands or something like that

ci/apply_folder.sh
```diff
@@ -0,0 +1,69 @@
...
+function error_exit() {
+    msg=${1}
+    >&2 echo "[Error] ${msg}"
+    exit 1
+}
+
+function prow_archive_state() {
+    if [ -n "${OPENSHIFT_CI:-}" ]; then
+        echo
+        echo "[INFO] Running in OpenShift CI (Prow)."
```
bardielle, line 18: Does it mean it will work only on prow's env? 
There is an assumption/ requirements before running those make commands?

ci/apply_folder.sh
```diff
@@ -0,0 +1,69 @@
...
+cd "${WORK_DIR}"
+HOME=${TERRAFORM_D_DIR} terraform init
+HOME=${TERRAFORM_D_DIR} terraform apply -auto-approve || prow_archive_state "true"
+
+set +o xtrace
+
+prow_archive_state
+
+echo
+echo "[INFO] Finished applying the terraform folder successfully"
```
bardielle, line 68: which folder? 
if we will have a lot of folders in the future? it should be an input parameter

ci/destroy_folder.sh
```diff
@@ -0,0 +1,69 @@
...
+
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+function error_exit() {
+    msg=${1}
+    >&2 echo "[Error] ${msg}"
+    exit 1
+}
+
+function prow_archive_state() {
```
bardielle, line 15: It seems that it's a common function 
Can you use a common file for that?

ci/destroy_folder.sh
```diff
@@ -0,0 +1,69 @@
...
+        set +o xtrace
+        echo "[INFO] ${STATE_ARCHIVE} created."
+    fi
+
+    if [ -n "${1:-}" ]; then
+        error_exit "TF command failed."
+    fi
+}
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
bardielle, line 32: Also, can all those parameter initialization be in a common function on a different file?

ci/destroy_folder.sh
```diff
@@ -0,0 +1,69 @@
...
+
+if [[ ! -d ${WORK_DIR} ]]; then
+    error_exit "can't load WORK_DIR"
+fi
+
+set -o xtrace
+
+cd "${WORK_DIR}"
+HOME=${TERRAFORM_D_DIR} terraform init
+HOME=${TERRAFORM_D_DIR} terraform destroy -auto-approve || prow_archive_state "true"
```
bardielle, line 61: I think you can make a refactoring -
use the command from outside with the same shell script file

Makefile
sagidayan: Im fine with that.

I prefer a stable atomic "flow". Re-compiling is seconds.

ci/apply_folder.sh
bardielle: We decided to add an instruction to the README file

bardielle (approved):
/lgtm

bardielle (approved):
/lgtm

PR #90, opened 2023-05-30
OCM-759: add error to "rosa_operator_roles_data_source" in case get stsOperatorRolesList failure

nirarg:
In case failing to list the STS Operator Roles List "terraform apply" should fail and not continue to next step This is done in order to make it more clear for user what is the error In addition, added hint to user for checking the credentials (token)


bardielle (approved):
/lgtm

PR #95, opened 2023-05-31
OCM-891: OIDC managed and unmanaged ci/e2e folders added

amalykhi:
https://issues.redhat.com/browse/OCM-891
New tf folders will allow to build managed and unmanaged oidc ROSA STS cluster via terraform-provider


ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
```diff
@@ -0,0 +1,91 @@
...
+#
+
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.20.0"
+    }
+    ocm = {
+      version = ">=1.0.1"
```
nirarg, line 24: Do we want to keep the version in environment variable and let the ci manage this?
Does this makes sense?

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
```diff
@@ -0,0 +1,91 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.20.0"
```
nirarg, line 21: We always get the latest version, don't we want to lock on specific version and manually increase versions?

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
```diff
@@ -0,0 +1,91 @@
...
+}
+
+data "ocm_rosa_operator_roles" "operator_roles" {
+  operator_role_prefix = var.operator_role_prefix
+  account_role_prefix = var.account_role_prefix
+}
+
+module operator_roles_and_oidc_provider {
+  source = "terraform-redhat/rosa-sts/aws"
+  version = "0.0.5"
```
nirarg, line 45: This must be managed by the CI, as the version might change (when using `=` and not `>=`)

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+}
+
+variable replicas {
+    type = string
+    default = "3"
+}
+
+variable openshift_version {
+    type = string
+    default = "4.13.0"
```
nirarg, line 41: Also need to be configured by the CI, we would like to prevent changing the code for this

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+}
+
+variable channel_group {
+    default = "stable"
```
nirarg, line 45: Also need to be configured by the CI

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+    default = "https://api.stage.openshift.com"
+}
+
+variable operator_role_prefix {
+    type = string
+}
+
+variable account_role_prefix {
+    type = string
+    default = ""
```
nirarg, line 17: Shouldn't we enforce the CI to give the prefix? shouldn't have default empty string as it might cause to bad behavior
The prefix is used for resources names and we need the CI to create the resources with known prefixes for better handling of undeleted resources, am I wrong?

nirarg (commented):
Few comments
Especially about the versioning which I think that need to be determined by the caller (from the CI)
notice that I added the comments only under `cluster_with_managed_oidc_config` but it relevant to similar changes in the next modules

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
amalykhi: I used the example files as template for this tffile. I think it is good idea, to define the version by CI. I would add it in different PR

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
amalykhi: Added >=

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
amalykhi: It is configured by tfvars.
Do you mean remove the default value?

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
amalykhi: It is configured by tfvars.
Do you mean remove the default value?

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
amalykhi: correct, it is required field , that should include account-role created before by the user.

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
nirarg: Yes, you right, it configurable.
The CI should always set the version? if yes, I think it better to remove it to have better indication if the CI missed it.
Same for "channel group" below

amalykhi: Good point, so I would prefer validate all the variables. WDYT @AsherShoshan if I remove all defaults from variables.tf?

amalykhi: the pre-submit check fails without the defaults(

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/main.tf
amalykhi: I think we could modify it later if decided to upgrade manually

ci/e2e/oidc_configuration/cluster_with_managed_oidc_config/variables.tf
amalykhi: As decided - we would try to not using defaults values

nirarg:
Thanks @amalykhi ,
the pre-submit is failing now because the CI doesn't add the openshift version and channel?
Is it something can be fixed before merging this PR?

PR #97, opened 2023-06-04
OCM-1018: Change provider logger to tflog

nirarg:
Implement new internal logger uses "github.com/hashicorp/terraform-plugin-log/tflog" This is done in order to have better logging
With previous logger implementation, all log levels are printed in DEBUG log level

Also found two issues in Terraform:

- If user not setting the `TF_LOG` environment variable, the logs are not printed
- The log prints the caller, in that case it line in the logging file

Need to check if those issues are already fixed in later versions


logging/logging.go
```diff
@@ -0,0 +1,107 @@
...
+	warnEnabled  bool
+	errorEnabled bool
+}
+
+// New creates the provider.
+func New() ocmlogging.Logger {
+	tfLogger := &TfLogger{}
+	logLevel := tflogging.LogLevel()
+	switch logLevel {
+	case "TRACE", "DEBUG":
```
osherdp, line 26: Am I missing something or is this struct re-implements standard logging functionality?
I mean, there got to be an existing logging implementation that allows to set a log-level and write lines based on it
(also for the "panic on fatal" behavior. it should be pretty common)

osherdp (approved):

PR #99, opened 2023-06-07
Allow users to set cluster_name, cloud_region and hosted_zones

nirarg:
In order to let the user control on the resources names and region, change those parameters to variables for managed and unmanaged clusters

Please notice: First commit is "make fmt_tf" auto results

### DIFF
--- a/examples/create_rosa_cluster/create_rosa_sts_cluster/classic_sts/account_roles/variables.tf
+++ b/examples/create_rosa_cluster/create_rosa_sts_cluster/classic_sts/account_roles/variables.tf
@@ -1,21 +1,21 @@
-variable ocm_environment {
-    type = string
-    default = "production"
+variable "ocm_environment" {
+  type    = string
+  default = "production"
 }
 
-variable openshift_version {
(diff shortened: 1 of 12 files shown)


examples/create_rosa_cluster/create_rosa_sts_cluster/classic_sts/account_roles/variables.tf
```diff
@@ -1,21 +1,21 @@
...
+  type    = string
+  default = "production"
 }
 
-variable openshift_version {
-    type = string
-    default = "4.12"
+variable "openshift_version" {
+  type    = string
+  default = "4.12"
```
amalykhi, line 8: I suggest to upgrade version to the latest GA 4.13

nirarg: So maybe better to remove this default, in order to enforce the user to add the version?
What do you say?

osherdp (approved):

PR #111, opened 2023-06-14
OCM-2299: IDP resource: Validate cluster exists before wait to ready state

nirarg:
When user try to create IDP, it expected that the cluster is already exists Avoid starting the poll loops in case the cluster doesn't exist


provider/identity_provider_resource.go
```diff
@@ -129,8 +130,20 @@ func (r *IdentityProviderResource) Create(ctx context.Context,
 		return
 	}
 
-	// Wait till the cluster is ready:
 	resource := r.collection.Cluster(state.Cluster.Value)
+	// We expect the cluster to be already exist
+	// Try to get it and if result with NotFound error, return error to user
+	if resp, err := resource.Get().SendContext(ctx); err != nil && resp.Status() == 404 {
```
bardielle, line 136: Please change the value 404 to be: http.StatusNotFound

provider/identity_provider_resource.go
```diff
@@ -129,8 +130,20 @@ func (r *IdentityProviderResource) Create(ctx context.Context,
...
+		message := fmt.Sprintf("Cluster %s not found, error: %v", state.Cluster.Value, err)
+		tflog.Error(ctx, message)
+		response.Diagnostics.AddError(
+			"Can't poll cluster state",
+			message,
+		)
+		return
+	}
```
bardielle, line 144:
```suggestion
	if resp, err := resource.Get().SendContext(ctx); err != nil {
	        if resp.Status() == 404 {
	        		message := fmt.Sprintf("Cluster %s not found, error: %v", state.Cluster.Value, err)
			        tflog.Error(ctx, message)
	        		
	        } else {
	        	        message := fmt.Sprintf("Can't poll cluster with ID %s, error %v", state.ID.Value, err)
	        }
	        response.Diagnostics.AddError(
	        	        "Can't poll cluster state",
	        	        message,
	        	        )
		return
	}
```

bardielle:
Please add a test case

provider/identity_provider_resource.go
nirarg: I think it has to fail only if the cluster is not found, for any other error (like possible connectivity issue) better to wait, no?
The point here is to fail in case the cluster doesn't exist

subsystem/identity_provider_resource_test.go
```diff
@@ -39,7 +39,43 @@ var _ = Describe("Identity provider creation", func() {
...
+				  "name": "my-cluster",
+				  "state": "ready"
+				}`),
+			),
+		)
+	})
+
+	It("Fail if cluster_id not found", func() {
+		// Prepare the server:
+		server.Reset()
```
bardielle, line 55: I prefer to use a context for the other test cases (and removing the before each)

nirarg: Done

bardielle (approved):

PR #114, opened 2023-06-15
Add an options to set version in oidc clusters

nirarg:


examples/create_rosa_cluster/create_rosa_sts_cluster/oidc_configuration/cluster_with_managed_oidc_config/main.tf
```diff
@@ -63,6 +63,8 @@ resource "ocm_cluster_rosa_classic" "rosa_sts_cluster" {
   cloud_region       = var.cloud_region
   aws_account_id     = data.aws_caller_identity.current.account_id
   availability_zones = var.availability_zones
+  version            = var.openshift_version
+  channel_group      = var.openshift_channel_group
```
bardielle, line 67: Channel group is not enabled in rosa cli. 
If I'm not wrong @arendej  asked that it will not be available to the public

nirarg: Ok, not mandatory to push this
But anyway the channel is already public, as it included in the resource attribute
Avoiding from adding it to the example is not the way to make it private

nirarg:
/hold
Don't merge, keeping this PR to have an example how to change version and channel in the example

examples/create_rosa_cluster/create_rosa_sts_cluster/oidc_configuration/cluster_with_managed_oidc_config/main.tf
bardielle: I know just one more way avoiding customer to use it

nirarg: I removed the channel group, now the PR includes only the version

nirarg:
@amalykhi

PR #119, opened 2023-06-20
OCM-2254: Add Dockerfiles for ARM and AMD64 binaries

enriquebelarte:
Build images for both architectures and use them for testing binary.

### DIFF
--- a/build/ci-tf-bin.Dockerfile
+++ b/build/ci-tf-bin.Dockerfile
@@ -1,0 +1,32 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS builder
+WORKDIR /root
+
+# Update the base image and install necessary packages
+RUN ARCH=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) && \
+    microdnf update -y && \
+    microdnf install -y git yum-utils python3-pyyaml python3-jinja2 make jq tar gzip go-toolset && \
+    yum-config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo && \
+    microdnf install -y terraform && \
+    microdnf clean all && \
+    #Install oc CLI
+    curl -sS https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp/stable/openshift-client-linux.tar.gz | tar -C /usr/local/bin -xzf - oc && \
+    curl -Ls https://github.com/openshift-online/ocm-cli/releases/download/v0.1.66/ocm-linux-$ARCH --output  /usr/local/bin/ocm && \
+    chmod +x /usr/local/bin/ocm 
+
+ENV PATH="/usr/local/go/bin:${PATH}"
+ENV GOPATH=/usr/local/go
+ENV TEST_OFFLINE_TOKEN=""
+
+# terraform-provider-ocm repo
+COPY . ./terraform-provider-ocm
+
+RUN go env -w GO111MODULE=on &&\
+    go install github.com/onsi/ginkgo/v2/ginkgo@latest &&\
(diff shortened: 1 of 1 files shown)


build/ci-tf-bin-arm64.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM --platform=linux/arm64 registry.access.redhat.com/ubi8/ubi-minimal:latest AS builder
```
osherdp, line 1: we really need two duplicated dockerfiles?
the base image ``registry.access.redhat.com/ubi8/ubi-minimal:latest`` is multi-arch

enriquebelarte: Did not know that. I'm working on a unified Dockerfile then.

Thanks.

build/ci-tf-bin.Dockerfile
osherdp, line 1: As discussed privately, let's trim-out everything that is not needed: the oc/ocm/terraform clients, golang test dependencies, etc.

build/ci-tf-bin.Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS builder
```
osherdp, line 1:
```suggestion
FROM registry.access.redhat.com/ubi8/go-toolset:1.19 AS builder
```
it can save us the golang installation and also set us to a specific version (instead of being unpinned)

build/ci-tf-bin.Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+ENV GOPATH=/usr/local/go
+ENV TEST_OFFLINE_TOKEN=""
+
+# terraform-provider-ocm repo
+COPY . ./terraform-provider-ocm
+
+RUN go env -w GO111MODULE=on &&\
+    go install github.com/onsi/ginkgo/v2/ginkgo@latest &&\
+    go install github.com/golang/mock/mockgen@v1.6.0 &&\
+    cd terraform-provider-ocm && go mod tidy && go mod vendor && make install &&\
```
osherdp, line 26: ``make build`` can be convenient because it requires less dependencies

build/ci-tf-bin.Dockerfile
enriquebelarte: go-toolset:1.19 is 1.21GB and ubi-minimal is 94.5MB. Installed size of go-toolset package and its dependencies is 593MB so I don't know how important is to set the specific golang version vs double the image size.

osherdp: Ack. If it's that much more space then I agree with you. Let's keep using the original image

build/ci-tf-bin.Dockerfile
```diff
@@ -5,28 +5,15 @@ WORKDIR /root
 RUN ARCH=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) && \
```
osherdp, line 5: I think we don't use ``ARCH`` anymore

build/ci-tf-bin.Dockerfile
```diff
@@ -5,28 +5,15 @@ WORKDIR /root
...
     microdnf update -y && \
     microdnf install -y git yum-utils python3-pyyaml python3-jinja2 make jq tar gzip go-toolset && \
```
osherdp, line 7:
```suggestion
    microdnf install -y git make go-toolset && \
```

osherdp (changes requested):

build/ci-tf-bin.Dockerfile
enriquebelarte: Right.

osherdp (approved):

PR #118, opened 2023-06-20
OCM-2408: Improve upgrade agreements acknowlegement process

nirarg:
* Add an option to approve - add boolean attribute "upgrade_admin_ack" (default=false)
* In case acknolege is missing, build elaborated description based on "upgradePoliciesClient" dry run

This is the error message displayed when try to update from 4.12.19 to 4.13.1 (candidate channel):
image

PR #121, opened 2023-06-20
OCM-2408: Improve upgrade agreements acknowledgement process

JohnStrunk:
Builds on #118

- Changes the `upgrade_admin_ack` boolean to be a string `upgrade_acknowledgements_for` that should be set to the openshift major.minor (i.e., "4.12")
- Adds acknowledgement for gates that don't need user intervention (i.e., STSOnly gates)

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -413,6 +413,11 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result 
 					ValueCannotBeChangedModifier(),
 				},
 			},
+			"upgrade_admin_ack": {
+				Description: ".",
+				Type:        types.BoolType,
+				Optional:    true,
+			},
 		},
 	}
 	return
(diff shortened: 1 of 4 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -1377,21 +1379,30 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
...
+		}
+		targetMinorVersion := getOcmVersionMinor(desiredVersion.String())
+		userAcksGates := !plan.UpgradeAcksFor.Unknown && !plan.UpgradeAcksFor.Null && plan.UpgradeAcksFor.Value == targetMinorVersion
+		if len(ackGateIDs) > 0 && !userAcksGates { // User has not acknowledged mandatory gates
+			return fmt.Errorf("%s\nTo acknowledge these items, please add \"upgrade_acknowledgements_for = %s\""+
+				" and re-apply the changes", description, targetMinorVersion)
+		}
+
+		// Ack all gates to OCM; including ones that don't require _user_ acks
+		allGates, _, err := upgrade.CheckMissingAgreements(desiredVersion.String(), state.ID.Value, upgradePoliciesClient, false)
```
nirarg, line 1395: Isn't it better to call this function only once and return all gates
In this case, the function can return in adition another bollean indicates if there are ackGates which require showing message to user, as it won't be possible to use `len(ackGateIDs)`  for that

PR #123, opened 2023-06-21
OCM-2342: Adding IDP examples and docs phase 2

sagidayan:
- LDAP


tshefi:
Looks good, tested successfully logged into my OCP cluster with an LDAP user/password.

PR #124, opened 2023-06-21
OCM-2428: Remove openshift-v prefix from create cluster version attribute

nirarg:


provider/cluster_rosa_classic_resource.go
```diff
@@ -686,7 +686,7 @@ func createClassicClusterObject(ctx context.Context,
 			return nil, errors.New(errHeadline + "\n" + description)
 		}
 		vBuilder := cmv1.NewVersion()
-		versionID := state.Version.Value
+		versionID := fmt.Sprintf("openshift-v%s", state.Version.Value)
```
bardielle, line 689: I think better moving the format string to a const

provider/cluster_rosa_classic_resource.go
```diff
@@ -1321,7 +1318,7 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
 	}
 
 	// Make sure the desired version is available
-	versionId := state.CurrentVersion.Value
+	versionId := fmt.Sprintf("openshift-v%s", state.CurrentVersion.Value)
```
bardielle, line 1321: Use the same const as I mentioned in my previous comment

subsystem/cluster_resource_rosa_test.go
```diff
@@ -2563,7 +2563,7 @@ var _ = Describe("ocm_cluster_rosa_classic - upgrade", func() {
 					worker_role_arn = "",
 				}
 			}
-			version = "openshift-v4.10.1"
+			version = "4.10.1"
```
bardielle, line 2566: Can you add a failure test that uses an incorrect version string?

provider/cluster_rosa_classic_resource.go
nirarg: Please notice that here removing the prefix from `state.Version.Value` and in the next comment removing it from `CurrentVersion.Value`
I don't think const is required here

subsystem/cluster_resource_rosa_test.go
nirarg: Please see new comment added https://github.com/terraform-redhat/terraform-provider-ocm/pull/124/commits/530b0881efb093ebfe052f6fd001164cba419894

bardielle (approved):

PR #138, opened 2023-06-26
OCM-2237: Fix path also to be used for the operator roles creation

nirarg:


nirarg:
/hold
This PR is blocked by 2 other PRs
https://github.com/terraform-redhat/terraform-aws-rosa-sts/pull/20 - after this merged need to create tag version 0.0.8 in the module
https://github.com/terraform-redhat/terraform-provider-ocm/pull/137 - Rebased on this PR, after rebase need to rebase on main

docs/upgrading-cluster.md
```diff
@@ -15,12 +15,12 @@ To upgrade your ROSA cluster to another version, export the following variable t
     ```
     export TF_VAR_openshift_version=<version_number>
     ```
-1. When upgrading from between major Y-Streams, you may need to provide an administrative acknowledgement that there are major changes for your cluster. For example, if you are upgrading from 4.11.43 to 4.12.21, you must use `4.12` as the value for this variable.
+1. If you choose to upgrade to a new version that necessitates approval, particularly when transitioning between major Y-Streams, you may be asked to provide administrative confirmation regarding significant modifications for your cluster. In such a scenario, during the initial upgrade attempt, you will receive an error message that provides guidance on the necessary changes. It is crucial to diligently follow those instructions and only acknowledge completion of the requirements by adding the "upgrade_acknowledgements_for" attribute to your resource, specifying the target version. For instance, if you are upgrading from 4.11.43 to 4.12.21, you should use '4.12' as the value for this variable.
```
bardielle, line 18:
```suggestion
1. When upgrading to a new cluster version, especially when transitioning between major Y-Streams, it requires approval. You might be requested to provide administrative confirmation regarding significant modifications for your cluster. In this case, when you first attempt the upgrade, you will receive an error message that provides guidance on the necessary modifications. It is essential to follow those instructions carefully and indicate completion of the requirements by adding the "upgrade_acknowledgements_for" attribute to your resource, specifying the target version. For example, if you are upgrading from 4.11.43 to 4.12.21, you should use '4.12' as the value for this variable.
```

bardielle (approved):
/lgtm

PR #142, opened 2023-06-26
OCM-2493: Add support for 1AZ pool in mAZ cluster

JohnStrunk:
Support creating single AZ machine pool in a multi-az cluster. AZ can be designated by either specifying the AZ name or by providing a subnet id for byo VPC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /dist/
 terraform.tfstate*
 /playground/*
+terraform-provider-rhcs
--- a/provider/machine_pool_resource.go
+++ b/provider/machine_pool_resource.go
@@ -53,16 +53,25 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
 				Description: "Identifier of the cluster.",
 				Type:        types.StringType,
 				Required:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					ValueCannotBeChangedModifier(),
+				},
 			},
 			"id": {
 				Description: "Unique identifier of the machine pool.",
 				Type:        types.StringType,
 				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+				},
 			},
 			"name": {
 				Description: "Name of the machine pool.Must consist of lower-case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character.",
 				Type:        types.StringType,
 				Required:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					ValueCannotBeChangedModifier(),
+				},
 			},
 			"machine_type": {
 				Description: "Identifier of the machine type used by the nodes, " +
@@ -142,6 +151,36 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
 				},
 				Optional: true,
 			},
+			"multi_availability_zone": {
+				Description: "Create a multi-AZ machine pool for a multi-AZ cluster (default true)",
+				Type:        types.BoolType,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+					ValueCannotBeChangedModifier(),
+				},
+			},
+			"availability_zone": {
+				Description: "Select availability zone to create a single AZ machine pool for a multi-AZ cluster",
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+					ValueCannotBeChangedModifier(),
+				},
+			},
+			"subnet_id": {
+				Description: "Select subnet to create a single AZ machine pool for BYOVPC cluster",
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+					ValueCannotBeChangedModifier(),
+				},
+			},
 		},
 	}
 	return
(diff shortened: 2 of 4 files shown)


provider/machine_pool_resource.go
```diff
@@ -142,6 +151,29 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
 				},
 				Optional: true,
 			},
+			"availability_zones": {
+				Description: "The availability zones in which to place the machine pool",
```
bardielle, line 155: The description in rosa cli is 
"Select availability zone to create a single AZ machine pool for a multi-AZ cluster"
I think we would like to be align to rosa cli with the purpose of that attribute

provider/machine_pool_resource.go
```diff
@@ -142,6 +151,29 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
...
+				},
+				Optional: true,
+				Computed: true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+					ValueCannotBeChangedModifier(),
+				},
+			},
+			"subnet_ids": {
+				Description: "The subnet IDs in which to place the machine pool",
```
bardielle, line 167: I think we would like to be align with rosa cli. The description in rosa cli is 
"Select subnet to create a single AZ machine pool for BYOVPC cluster"

provider/machine_pool_resource.go
```diff
@@ -219,6 +251,46 @@ func (r *MachinePoolResource) Create(ctx context.Context,
...
+				"Can't build machine pool for cluster '%s': %v",
+				state.Cluster.Value, err,
+			),
+		)
+		return
+	}
+	if len(subnetList) > 0 {
+		builder.Subnets(subnetList...)
+	}
+
```
bardielle, line 293: What if `len(subnetList) > 0 && len(azList) > 0` ? Would we like to block it?

provider/machine_pool_resource.go
```diff
@@ -455,6 +560,71 @@ func (r *MachinePoolResource) Update(ctx context.Context, request tfsdk.UpdateRe
...
+			}
+			if !poolInClusterSubnet {
+				return false, fmt.Errorf("Setting \"subnet_ids\" is not compatible with the cluster's subnets")
+			}
+		}
+	}
+
+	// It's a multi-az pool if the cluster is multi-az and they didn't tell us
+	// which az to put the pool into
+	isMultiAZPool := isMultiAZCluster && len(poolAZs) != 1 && len(poolSubnets) != 1
```
bardielle, line 623: As I see in rosa cli - you can't see AZs and submets together in one machine pool  - I think it should be changed to :
```suggestion
	isMultiAZPool := isMultiAZCluster && (len(poolAZs) > 1 || len(poolSubnets) > 1)
```

JohnStrunk: It is not necessary to specify subnets or pool AZs in the case where you want the same "multi"-ness as the cluster.
In that case, len of both of those could be 0. For a multiAZ cluster, that would lead us to mistakenly think it's a single AZ pool, when it's not.

provider/machine_pool_resource.go
JohnStrunk: I decided not to because once it has been created, OCM populates the az list in the create response. If the user had supplied subnets, now we'd have both in the state file, causing the "both" test to fail when an update is applied (e.g., scaling).

provider/machine_pool_resource.go
```diff
@@ -455,6 +560,71 @@ func (r *MachinePoolResource) Update(ctx context.Context, request tfsdk.UpdateRe
 	response.Diagnostics.Append(diags...)
 }
 
+// Validate the machine pool's settings that pertain to availability zones.
+// Returns whether the machine pool is/will be multi-AZ.
+func (r *MachinePoolResource) validateAZConfig(state *MachinePoolState) (bool, error) {
```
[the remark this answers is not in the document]
JohnStrunk: We need information about the cluster in order to do the validation. I don't think that is available during the validation phase.
Specifically, `validateAZConfig` makes use of the cluster client to grab the az and subnet data for the specific cluster in order to determine how long the lists are allowed to be, and what values may be in each list.

.gitignore
```diff
@@ -3,3 +3,4 @@
 /dist/
 terraform.tfstate*
 /playground/*
+terraform-provider-ocm
```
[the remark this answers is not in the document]
sagidayan: Right, let's add `terraform-provider-rhcs` then.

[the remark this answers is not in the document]
sagidayan: the name confused me...

provider/machine_pool_resource.go
nirarg: I understand the issue in update scenario, but this validation is for create scenario. we can have this validation only for create, not for update. am I right?

provider/machine_pool_resource.go
```diff
@@ -142,6 +151,29 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
 				},
 				Optional: true,
 			},
+			"availability_zones": {
```
nirarg, line 154: In Rosa CLI used single availability zone, need to be aligned

provider/machine_pool_resource.go
```diff
@@ -142,6 +151,29 @@ func (t *MachinePoolResourceType) GetSchema(ctx context.Context) (result tfsdk.S
...
+					ElemType: types.StringType,
+				},
+				Optional: true,
+				Computed: true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
+					ValueCannotBeChangedModifier(),
+				},
+			},
+			"subnet_ids": {
```
nirarg, line 166: In Rosa CLI used single subnet ID, need to be aligned

JohnStrunk:
@nirarg @bardielle @amalykhi 

Some thoughts:
- Both AZ and subnet cannot be modified once the pool is created. This means we don't need to do anything w/ them during Update(), only Create().
- OCM sends back the actual AZ/subnet(s), and that has led to a fair bit of the complication in validation.
- The `--multi-availability-zone` of rosa cli is redundant based on the AZ/subnet options & type of cluster.

Proposal:
- Switch to a single string for each of AZ and subnet options
- Auto deduce "multi-availability-zone" based on (1) multi-zone cluster and (2) whether AZ or subnet are provided.
- Don't update AZ/subnet in the state file based on OCM response, just preserve the user's config. This will allow us to prevent modification while also avoiding needing a list and complicated validation.

provider/machine_pool_resource.go
```diff
@@ -464,6 +554,86 @@ func (r *MachinePoolResource) Update(ctx context.Context, request tfsdk.UpdateRe
...
+		if !inClusterSubnet {
+			return false, fmt.Errorf("subnet_id %s is not valid for cluster %s", state.SubnetID.Value, state.Cluster.Value)
+		}
+	} else {
+		if len(clusterSubnets) > 0 && !state.MultiAvailabilityZone.Value && isMultiAZCluster {
+			state.SubnetID = types.String{Value: clusterSubnets[0]}
+		} else {
+			state.SubnetID = types.String{Null: true}
+		}
+	}
```
nirarg, line 633: If I understand the code right here, there might be situation in which you set both az and subnet, am I right?
So it possible that user sets az and code use `clusterSubnets[0]`?
If I understand it right from the cli code, if user ask to singleAZ but doesn't set either AZ or subnet, it takes only `clusterAZs[0]` and keep subnet empty

JohnStrunk: done

nirarg:
Please squash all commits before merge

bardielle:
@JohnStrunk greate job! 
Please squash all the commits before I'm approving the PR

JohnStrunk:
> Please squash all commits before merge

@nirarg done.

PR #146, opened 2023-06-27
OCM-2501: OCM-2512: Improve upgrade semantics

JohnStrunk:
- Error if the requested version set to a value < current cluster version
- Cancel pending upgrades if requested version is set to current version

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -1352,42 +1352,8 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
 	cancelingUpgradeOnly := desiredVersion.Equal(currentVersion)
 
 	if !cancelingUpgradeOnly {
-		// Make sure the desired version is available
-		versionId := fmt.Sprintf("openshift-v%s", state.CurrentVersion.Value)
-		if !state.ChannelGroup.Unknown && !state.ChannelGroup.Null && state.ChannelGroup.Value != ocm.DefaultChannelGroup {
-			versionId += "-" + state.ChannelGroup.Value
-		}
-		availableVersions, err := upgrade.GetAvailableUpgradeVersions(ctx, r.versionCollection, versionId)
-		if err != nil {
-			return fmt.Errorf("failed to get available upgrades: %v", err)
-		}
-		found := false
-		for _, v := range availableVersions {
-			sem, err := semver.NewVersion(v.RawID())
-			if err != nil {
-				return fmt.Errorf("failed to parse available upgrade version: %v", err)
-			}
-			if desiredVersion.Equal(sem) {
-				found = true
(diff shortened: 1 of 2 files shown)


JohnStrunk:
/cc @nirarg
FYI @amalykhi

JohnStrunk:
/retest
Looks like CI is broken... related to rename?

provider/cluster_rosa_classic_resource.go
```diff
@@ -1326,47 +1340,55 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
...
+	if currentVersion.GreaterThan(desiredVersion) {
 		tflog.Debug(ctx, "No cluster version upgrade needed.")
+		if requestedVersionChanged {
+			// User changed the version they want, but actual is higher. We
+			// don't support downgrades.
+			return fmt.Errorf("cluster version is already above the requested version")
+		}
 		return nil
 	}
+	cancelingUpgradeOnly := desiredVersion.Equal(currentVersion)
```
bardielle, line 1352: correct me if I'm wrong, but it seems that if cancelingUpgradeOnly is true - you can return from that function

provider/cluster_rosa_classic_resource.go
```diff
@@ -1326,47 +1340,55 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
...
-	availableVersions, err := upgrade.GetAvailableUpgradeVersions(ctx, r.versionCollection, versionId)
-	if err != nil {
-		return fmt.Errorf("failed to get available upgrades: %v", err)
-	}
-	found := false
-	for _, v := range availableVersions {
-		sem, err := semver.NewVersion(v.RawID())
-		if err != nil {
-			return fmt.Errorf("failed to parse available upgrade version: %v", err)
+	if !cancelingUpgradeOnly {
```
bardielle, line 1354: That function become to be complicated - Can you refactor that code and move it to another function?

provider/cluster_rosa_classic_resource.go
```diff
@@ -1381,7 +1403,7 @@ func (r *ClusterRosaClassicResource) upgradeClusterIfNeeded(ctx context.Context,
 		return err
 	}
 
-	if !correctUpgradePending {
+	if !correctUpgradePending && !cancelingUpgradeOnly {
 		// Gate agreements are checked when the upgrade is scheduled, resulting
```
bardielle, line 1407: Also I'm prefer to refactor that logic and move it to another function (lines 1407- 1460) if it possible

JohnStrunk:
I think the flow of upgradeClusterIfNeeded is better now that the 2 if-clauses were factored out.
/cc @bardielle

JohnStrunk:
/retest
maybe there will be quota today

PR #172, opened 2023-07-04
OCM-2549: Add notice to user that operator policies are created as part of the account roles example

nirarg:


sagidayan (approved):
/lgtm

PR #183, opened 2023-07-07
OCM-2360: Add support for importing a classic cluster

JohnStrunk:
Adds support for running `terraform import` to pull in a cluster.

**Usage:**
Add a minimal definition of the cluster to the `.tf` file:
```
resource "rhcs_cluster_rosa_classic" "mycluster" { }
```
Run import on that resource, passing the cluster's id:
```console
$ terraform import rhcs_cluster_rosa_classic.mycluster 24q7p8vthmm5erql10nn3qovq45juhsk
rhcs_cluster_rosa_classic.mycluster: Importing from ID "24q7p8vthmm5erql10nn3qovq45juhsk"...
data.rhcs_policies.all_policies: Reading...
rhcs_cluster_rosa_classic.mycluster: Import prepared!
  Prepared rhcs_cluster_rosa_classic for import
rhcs_cluster_rosa_classic.mycluster: Refreshing state... [id=24q7p8vthmm5erql10nn3qovq45juhsk]
data.rhcs_policies.all_policies: Read complete after 1s
data.aws_caller_identity.current: Reading...
data.aws_caller_identity.current: Read complete after 0s [id=765374464689]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -30,7 +30,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/properties"
 
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/terraform-redhat/terraform-provider-rhcs/build"
@@ -43,6 +45,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	ocm_errors "github.com/openshift-online/ocm-sdk-go/errors"
@@ -98,6 +101,7 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result 
 				Optional:    true,
 				Computed:    true,
 				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
 					ValueCannotBeChangedModifier(),
 				},
 			},
@@ -126,6 +130,7 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result 
 				Optional: true,
 				Computed: true,
 				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
 					ValueCannotBeChangedModifier(),
 				},
 			},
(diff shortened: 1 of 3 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -109,21 +111,31 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result
...
 				PlanModifiers: []tfsdk.AttributePlanModifier{
+					tfsdk.UseStateForUnknown(),
 					ValueCannotBeChangedModifier(),
 				},
 			},
 			"name": {
 				Description: "Name of the cluster. Cannot exceed 15 characters in length.",
 				Type:        types.StringType,
-				Required:    true,
+				Optional:    true,
```
bardielle, line 121: Cluster name is a required attribute (in rosa cli and the OCM API)
There is a reason you changed it to optional?

JohnStrunk: Required fields must be supplied in the config file and cannot be pulled in via Read(). Leaving this as required (and any of the others as well) would force them to be provided to the import command, in addition to the cluster_id.
To keep the import concise, I only require the cluster_id, and allow read() to populate all the other fields--- that only works if `computed` is true.

The validation now ensures that the formerly `required` fields are defined, it just doesn't differentiate how they get their value (from the user config or from querying the cluster via read()). Those fields will still need to be provided on cluster creation, and that will be enforced.

nirarg: How other providers do this?
I couldn't find any other good examples, as this provider using very old version

JohnStrunk: Regarding docs, yes, autogeneration will move these fields to the optional category. I'm not sure that's actually a bad thing though. I believe there is a mismatch between what terraform means by "required" and what we are trying to convey...

Terraform => "**User** must provide a value"
Us => "This field must have a value"

The distinction really only applies when the resource is imported.

----

It may be possible to change these fields back to required and force the user to put them into the config file prior to import, but that is both redundant and doesn't follow hashi's described flow. However, I'm happy to give it a try if that's how you want it to work.

bardielle: That's why we had the "get" call to ocm in order to supply the required params

JohnStrunk: I'm not understanding...
The only thing we need in order to do the "refresh" is the cluster id, so that's all I ask for at the import step.
Once import calls Read() to do the refresh, all the other fields get populated into the state file (including name, etc.).

The roadblock I hit was that even though `name` was in the state, I got an error that it was a "required" parameter when I ran TF apply the next time (e.g., to scale the cluster).

JohnStrunk:
/hold
Tracking down an issue w/ the STS fields after a subsequent apply()

provider/cluster_rosa_classic_resource.go
JohnStrunk: After finding a different explanation/walkthrough of the import process, this is not needed. I have removed it.

nirarg:
@JohnStrunk this is look great
Great job on investigating this
Please let me know when all validation are done so we can merge

PR #186, opened 2023-07-10
OCM-2644: Recreate machine pool if it was deleted from other client.

sagidayan:
This PR adds the ability to reconcile a machine pool that was created by TF, but then was removed via a different method, Not via terraform.

In this case, And if the user still has the machine pool resource in his `.tf` files. Upon an `apply` the provider will re-create the pool. 

Before this PR, the provider would have error out, since it could not find the machine pool defined in the state.

### DIFF
--- a/provider/machine_pool_resource.go
+++ b/provider/machine_pool_resource.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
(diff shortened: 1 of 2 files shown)


nirarg:
/approve
@bardielle please take a look

sagidayan:
/hold

refactoring

PR #190, opened 2023-07-11
OCM-2666: Remove AWS validations from rosa_cluster resource

nirarg:
This change is done in order to avoid the requirement to have valid AWS credentials in the local machine, in which the terraform executed. This change include:
* Remove the relevant code from the provider
* changing the subsystem accordingly to the change
* Removing limitation from docs

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -32,17 +32,11 @@ import (
 
 	"github.com/openshift/rosa/pkg/helper"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/client"
-	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/terraform-redhat/terraform-provider-rhcs/build"
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/upgrade"
 
-	"github.com/aws/aws-sdk-go/aws/arn"
-	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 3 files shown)


provider/cluster_rosa_classic_resource.go
```diff
@@ -1126,6 +900,7 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context,
 		)
 		return
 	}
+	tflog.Info(ctx, fmt.Sprintf("************************ 1"))
```
bardielle, line 903: please remove the log prints

provider/cluster_rosa_classic_resource.go
```diff
@@ -1138,6 +913,7 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context,
 		)
 		return
 	}
+	tflog.Info(ctx, fmt.Sprintf("************************ 2"))
```
bardielle, line 916: please remove the log prints

provider/common/helpers.go
```diff
@@ -151,3 +151,12 @@ func HandleErr(res *ocmerrors.Error, err error) error {
 	errType := weberr.ErrorType(res.Status())
 	return errType.Set(errors.Errorf("%s", msg))
 }
+
+// StringValue returns the value of the string pointer passed in or
+// "" if the pointer is nil.
```
bardielle, line 156: If I'm not wrong - that function is not needed anymore...

nirarg: You right, removed it

provider/cluster_rosa_classic_resource.go
nirarg: Oops ... removed

provider/cluster_rosa_classic_resource.go
nirarg: Oops ... removed

PR #196, opened 2023-07-13
OCM-2645: Recreate cluster if it was deleted from other client

sagidayan:
This PR adds the ability to reconcile a cluster that was created by TF, but then was removed via a different method, Not via terraform.

In this case, And if the user still has the cluster resource in his `.tf` files. Upon an `apply`, the provider will re-create the cluster.

Before this PR, the provider would have error out, since it could not find the cluster defined in the state.

Note: This PR also modifies the `cluster_waiter`. This is because we now do need to update the resource if the cluster id has changed. Also moved the timeout validation into a `validator` **this will now throw an error** if the user has set the `timeout` to a negative number. Unlike it worked before, printing a warning and using the default value.

The polling (`cluster_waiter`) was also moved to a new function, so both `Create` and `Update` can call.

### DIFF
--- a/provider/cluster_waiter_resource.go
+++ b/provider/cluster_waiter_resource.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
 )
 
 type Waiter interface {
@@ -43,8 +44,9 @@ func (t *ClusterWaiterResourceType) GetSchema(ctx context.Context) (result tfsdk
 			"timeout": {
 				Description: "An optional timeout till the cluster is ready. The timeout value should be in minutes." +
 					" the default value is 60 minutes",
-				Type:     types.Int64Type,
-				Optional: true,
+				Type:       types.Int64Type,
+				Optional:   true,
+				Validators: timeoutValidators(),
 			},
 			"ready": {
 				Description: "Whether the cluster is ready",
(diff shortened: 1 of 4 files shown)


provider/cluster_waiter_resource.go
```diff
@@ -81,40 +83,12 @@ func (r *ClusterWaiterResource) Create(ctx context.Context,
...
 	if err != nil {
-
-		response.Diagnostics.AddError(
-			"Can't poll cluster state",
-			fmt.Sprintf(
-				"Can't poll state of cluster with identifier '%s': %v",
-				state.Cluster.Value, err,
-			),
-		)
+		response.Diagnostics.AddError("Can't poll cluster state", err.Error())
```
nirarg, line 89: Please add here information if this is during create or update
Something like:
```suggestion
		response.Diagnostics.AddError("Can't poll cluster state (Create resource)", err.Error())
```

provider/cluster_waiter_resource.go
```diff
@@ -123,12 +97,43 @@ func (r *ClusterWaiterResource) Create(ctx context.Context,
...
+	state := &ClusterWaiterState{}
+	diags := request.State.Get(ctx, state)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+	state, err := r.startPolling(ctx, state)
+
+	if err != nil {
+		response.Diagnostics.AddError("Can't poll cluster state", err.Error())
```
nirarg, line 109: Please add here information if this is during create or update
Something like:
```suggestion
		response.Diagnostics.AddError("Can't poll cluster state (Read resource)", err.Error())
```

provider/cluster_waiter_resource.go
```diff
@@ -123,12 +97,43 @@ func (r *ClusterWaiterResource) Create(ctx context.Context,
...
+	diags := request.Plan.Get(ctx, plan)
+	_ = request.Plan.Get(ctx, plan)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+	state, err := r.startPolling(ctx, plan)
+
+	if err != nil {
+		response.Diagnostics.AddError("Can't poll cluster state", err.Error())
```
nirarg, line 130: Please add here information if this is during create or update
Something like:
```suggestion
		response.Diagnostics.AddError("Can't poll cluster state (Update resource)", err.Error())
```

provider/cluster_waiter_resource.go
```diff
@@ -179,3 +213,23 @@ func (r *ClusterWaiterResource) retryClusterReadiness(attempts int, sleep time.D
...
+		&common.AttributeValidator{
+			Desc: "Timeout must be a positive number",
+			Validator: func(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+				timeout := &types.Int64{}
+				diag := req.Config.GetAttribute(ctx, req.AttributePath, timeout)
+				if diag.HasError() {
+					// No attribute to validate
+					return
+				}
+				if !timeout.Null && !timeout.Unknown && timeout.Value < 0 {
```
nirarg, line 229: Can it be zero?
The error message below indicates that the answer is no :-)
```suggestion
				if !timeout.Null && !timeout.Unknown && timeout.Value <= 0 {
```

provider/cluster_waiter_resource.go
```diff
@@ -123,12 +97,43 @@ func (r *ClusterWaiterResource) Create(ctx context.Context,
 
 func (r *ClusterWaiterResource) Read(ctx context.Context, request tfsdk.ReadResourceRequest,
 	response *tfsdk.ReadResourceResponse) {
-	// Do Nothing
+	state := &ClusterWaiterState{}
```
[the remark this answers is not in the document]
sagidayan: Opp, changes were not committed. Let me fix that.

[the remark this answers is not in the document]
sagidayan: It should be just create/update

provider/cluster_waiter_resource.go
sagidayan: Not relevant. No read operations.

provider/cluster_waiter_resource.go
sagidayan: done

provider/cluster_waiter_resource.go
sagidayan: done

PR #198, opened 2023-07-13
OCM-2636: Enable importing machine pools

JohnStrunk:
Tested importing:
- 1AZ cluster, 1AZ pool
- mAZ cluster, 1AZ pool
- mAZ cluster, mAZ pool

Note: I removed the verification that replicas is a multiple of 3 in the case of mAZ pools. OCM provides this validation separately. The reason for the removal from the TF code is that it was causing a problem during updates after state import wrt the 1AZ optional fields for mAZ clusters. It would have required additional API queries to read the cluster object.

### DIFF
--- a/provider/machine_pool_resource.go
+++ b/provider/machine_pool_resource.go
@@ -705,19 +705,19 @@ func (r *MachinePoolResource) Delete(ctx context.Context, request tfsdk.DeleteRe
 
 func (r *MachinePoolResource) ImportState(ctx context.Context, request tfsdk.ImportResourceStateRequest,
 	response *tfsdk.ImportResourceStateResponse) {
-	// To import a machine pool, we need to know the cluster ID and the machine pool name
+	// To import a machine pool, we need to know the cluster ID and the machine pool ID
 	fields := strings.Split(request.ID, ",")
 	if len(fields) != 2 || fields[0] == "" || fields[1] == "" {
 		response.Diagnostics.AddError(
 			"Invalid import identifier",
-			"Machine pool to import should be specified as <cluster_id>,<machine_pool_name>",
+			"Machine pool to import should be specified as <cluster_id>,<machine_pool_id>",
 		)
 		return
 	}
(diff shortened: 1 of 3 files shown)


provider/machine_pool_resource.go
```diff
@@ -455,34 +456,11 @@ func (r *MachinePoolResource) Update(ctx context.Context, request tfsdk.UpdateRe
...
-		)
-		return
-	}
-
 	computeNodesEnabled := false
 	autoscalingEnabled := false
 
 	if !plan.Replicas.Unknown && !plan.Replicas.Null {
 		computeNodesEnabled = true
-		if isMultiAZPool && plan.Replicas.Value%3 != 0 {
```
nirarg, line 476: Why did you remove this validation?
In case of multiple AZ, replicas must be replica of 3
This is the relevant validation in the cli:
https://github.com/openshift/rosa/blob/4de5ba8ee9b6b9eff715d3f8c5fa9ec362047584/cmd/edit/machinepool/machinepool.go#L65

JohnStrunk: I'm the one that added the check originally: https://github.com/terraform-redhat/terraform-provider-rhcs/commit/bc3214e95efa633429ac62afceb5a4a9f80cd00a
OCM validates the replica count, so it isn't actually necessary... I just added it because (1) it was in the cli and (2) I had the info for other reasons.

I removed it because it relies on knowing `isMultiAZPool` which comes from `validateAZConfig()`, and the logic in the validate function wasn't able to cope w/ the state from import. It comes down to the validation using the user config values. However, import (Read) doesn't set them in the state, so TF believes the fields are null. Since they are locked from updating, the user can't set them (nor should they need to, honestly, since the values and validation only matter on Create, not Update).

I add a new function to determine the multi-ness of the pool and keep this validation if you'd like.

nirarg: But the validateAZConfig use the plan and not the state, so maybe it will work with the state?

PR #200, opened 2023-07-14
OCM-2671: Add import of identity provider

JohnStrunk:


provider/identity_provider_resource.go
```diff
@@ -580,10 +581,62 @@ func (r *IdentityProviderResource) Delete(ctx context.Context, request tfsdk.Del
...
-		tftypes.NewAttributePath().WithAttributeName("id"),
-		request,
-		response,
-	)
+	// To import an identity provider, we need to know the cluster ID and the provider name.
+	fields := strings.Split(request.ID, ",")
+	if len(fields) != 2 || fields[0] == "" || fields[1] == "" {
+		response.Diagnostics.AddError(
+			"Invalid import identifier",
+			"Identity provider to import should be specified as <cluster_id>,<provider_name>",
```
bardielle, line 589: why not required the provider_id? https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__identity_providers__identity_provider_id_

JohnStrunk: I chose to use the provider name because it didn't look like I could get the provider id easily from rosa cli (i.e., how would a user obtain the ID to pass in?).

PR #204, opened 2023-07-18
OCM-2630, OCM-2858: Fix Replicas/Autoscaling validations

nirarg:


nirarg:
/hold cancel
The changes in `ocm-common` where already merged

Makefile
```diff
@@ -74,6 +74,10 @@ unit-test:
 		--succinct \
 		-ldflags="$(ldflags)" \
 		-r provider
+	ginkgo run \
+		--succinct \
+		-ldflags="$(ldflags)" \
+		-r internal/...
```
osherdp, line 80:
```suggestion
		-r provider internal/...
```

internal/ocm/resource/cluster_test.go
```diff
@@ -0,0 +1,186 @@
...
+	})
+})
+
+func int64Pointer(src int64) *int64 {
+	return &src
+}
+
+func stringPointer(src string) *string {
+	return &src
+}
```
osherdp, line 186:
```suggestion
func Pointer[T ~int64 | ~string](src T) *T {
	return &src
}
```

osherdp (changes requested):

Makefile
nirarg: I tried this, it didn't work

osherdp: maybe only with a newer ginkgo I guess

nirarg: Its seems that I had a mistake before, now it works :-)
Fixed

internal/ocm/resource/cluster_test.go
nirarg: Changed it to generic of type any
This generic function enforce me to declare that the type is from type int64, otherwise its int
So I called the function like: `pointer(int64(2))`

osherdp (approved):

internal/ocm/resource/cluster_test.go
osherdp: Even better

bardielle:
/lgtm
Great job!!!

PR #205, opened 2023-07-20
OCM-2674: ability to create an admin user when creating a cluster

sagidayan:
This PR Adds an optional attribute to rosa_classic, if exists, it will create an admin user on cluster create.

On the way, added validations to HTPasswd (username/password) since they are used here, and were missing from the implementation.


provider/idps/htpasswd.go
```diff
@@ -39,3 +43,70 @@ func CreateHTPasswdIDPBuilder(ctx context.Context, state *HTPasswdIdentityProvid
 	}
 	return builder
 }
+
+func HTPasswdValidators() []tfsdk.AttributeValidator {
```
nirarg, line 47: Can't htpasswd and admincreds use the same validator?
Istead of having this duplicated?

sagidayan: As of now, it can. They are the same. But we do plan to extend the htpasswd IDP (soon) to also accept a list of users. This will change the validator, and it won’t apply to the admin case.

nirarg:
/approve
@EricPonvelle please review the document change

PR #209, opened 2023-07-24
OCM-2415: use rosa-sts v0.0.12 - openshift version validation

nirarg:


examples/create_account_roles/variables.tf
```diff
@@ -4,8 +4,7 @@ variable "ocm_environment" {
 }
 
 variable "openshift_version" {
-  type    = string
-  default = "4.13"
+  type = string
```
bardielle, line 7: I think it better to set a default here to an empty string, Otherwise, The error will be from the examples instead of the module...

bardielle: I tested it - if the user does not set it, TF asks him to supply a value in run-time... 
I changed it to have a default value with an empty string

PR #211, opened 2023-07-25
OCM-2670: Fix GitHub IDP example

sagidayan:
Could not find the cause of the array not being passed to terraform. 

This PR will simply make the example work.

### DIFF
--- a/examples/create_identity_provider/github/main.tf
+++ b/examples/create_identity_provider/github/main.tf
@@ -34,6 +34,6 @@ resource "rhcs_identity_provider" "github_idp" {
   github = {
     client_id     = var.github_client_id
     client_secret = var.github_client_secret
-    organizations = var.github_orgs
+    organizations = [var.github_org]
   }
 }
--- a/examples/create_identity_provider/github/variables.tf
+++ b/examples/create_identity_provider/github/variables.tf
@@ -24,7 +24,7 @@ variable "github_client_secret" {
   type        = string
   description = "GitHub client secret"
 }
-variable "github_orgs" {
-  type        = list(string)
-  description = "List of GitHub organisations"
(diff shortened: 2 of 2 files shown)


examples/create_identity_provider/github/variables.tf
```diff
@@ -24,7 +24,7 @@ variable "github_client_secret" {
   type        = string
   description = "GitHub client secret"
 }
-variable "github_orgs" {
-  type        = list(string)
-  description = "List of GitHub organisations"
+variable "github_org" {
+  type        = string
+  description = "GitHub organisation"
```
osherdp, line 29:
```suggestion
  description = "GitHub organization"
```

osherdp (changes requested):

examples/create_identity_provider/github/main.tf
```diff
@@ -34,6 +34,6 @@ resource "rhcs_identity_provider" "github_idp" {
   github = {
     client_id     = var.github_client_id
     client_secret = var.github_client_secret
-    organizations = var.github_orgs
+    organizations = [var.github_org]
```
osherdp, line 37: wdym by saying the array isn't being passed to terraform? we should have some examples like that in assisted I think: https://github.com/openshift/assisted-test-infra/blob/ca8ab6295b6e401508d2d164058bdb2e58e4ab93/terraform_files/baremetal/main.tf#L26

sagidayan: I think its something to do with the **(OLD)** sdk version we are using. An old one.

We could not figure it out. So we will reach out to HashiCorp. But for now, we want the example to simply work.

osherdp (approved):

PR #212, opened 2023-07-25
OSDOCS-6873: Move create account roles into docs (WIP)

nirarg:
This PR is suggestion how to move all examples to be under the docs directory, so it can be included in the terraform registry.

I will create a script that will create terraform directories from this file, so the examples can be tested by users/developers

Please let me know what you think about this suggestion


docs/examples/create_account_iam_resources.md
```diff
@@ -0,0 +1,142 @@
...
+  operator_role_policies = data.rhcs_policies.all_policies.operator_role_policies
+  all_versions           = data.rhcs_versions.all
+  path                   = var.path
+  tags                   = var.tags
+}
+```
+
+## Verification
+
+1. In the `rosa` CLI, run the following command to verify that the account roles are created:
```
amalykhi, line 123: probably users would be interested to verify also from AWS side, but I didn't find the link with the description in AWS docs.

amalykhi:
I like the idea. Will it try to generate .tfvars file for the user with some sample values(except token that can't be done)?

nirarg:
> I like the idea. Will it try to generate .tfvars file for the user with some sample values(except token that can't be done)?

@amalykhi good idea, I will think how to add also .tfvars to auto generation files
Thanks

PR #222, opened 2023-08-02
OCM-3005: Add option to set private connectivity

nirarg:
This PR enables the option to create private cluster.
Until now any cluster with private link was private cluster
After this change the following options are allowed:

- public cluster (default)
- public cluster (optionally with PrivateLink)
- private cluster (requires PrivateLink)

This PR also move all AWS and APIPrivacy related attribute usage to internal package with improved validations and unit tests


nirarg:
@bardielle & @sagidayan 
The PR is ready
Please review

internal/ocm/resource/cluster.go
```diff
@@ -1,13 +1,18 @@
...
-	"context"
 	"errors"
+	"fmt"
+	"regexp"
 
 	"github.com/openshift-online/ocm-common/pkg/cluster/validations"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
+var kmsArnRE = regexp.MustCompile(
```
bardielle, line 12: Can we move it into the common repo?

nirarg: Opened a jira ticket [OCM-3155](https://issues.redhat.com/browse/OCM-3155)
This will be added in different PR

internal/ocm/resource/cluster.go
```diff
@@ -93,3 +98,79 @@ func (c *Cluster) CreateNodes(ctx context.Context, autoScalingEnabled bool, repl
...
+func (c *Cluster) CreateAWSBuilder(awsTags map[string]string, ec2MetadataHttpTokens *string, kmsKeyARN *string,
+	isPrivateLink bool, awsAccountID *string, stsBuilder *cmv1.STSBuilder, awsSubnetIDs []string) error {
+
+	if isPrivateLink && awsSubnetIDs == nil {
+		return errors.New("Clusters with PrivateLink must have a pre-configured VPC. Make sure to specify the subnet ids.")
+	}
+
+	awsBuilder := cmv1.NewAWS()
+
+	awsBuilder.Tags(awsTags)
```
bardielle, line 111: I think it better to check if awsTags is not nil

nirarg: Done

PR #225, opened 2023-08-03
V1.2.2 fix

bardielle:


sagidayan:
If we update the tag on this commit, we should update all links accordingly. FYI. Otherwise, LGTM

PR #227, opened 2023-08-03
OCM-3042: Remove ability to manage default machinepool via cluster

JohnStrunk:
This makes the default machine pool fields in the cluster resource only used for creating the initial cluster. They cannot be used to update the default worker pool.

- Associated fields cannot be changed
- Fields are not read back from OCM into the state

NOTE: After merging this PR, it will not be possible to manage the default machinepool until the rest of [OCM-3042](https://issues.redhat.com//browse/OCM-3042) is complete.

PR #228, opened 2023-08-03
OCM-3042: Default MachinePool should be manipulated via machine_pool resource

JohnStrunk:
Supersedes #227

- [x] "worker" pool can be specified at cluster creation and it automatically imports the default worker pool
- [x] Modifying the "worker" pool adjusts the default pool
- [x] Default pool can be deleted by removing the "worker" pool
- [x] Once "worker" pool is deleted, it can't be recreated

### DIFF
--- a/provider/clusterrosaclassic/cluster_rosa_classic_resource.go
+++ b/provider/clusterrosaclassic/cluster_rosa_classic_resource.go
@@ -199,18 +199,16 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 				},
 			},
 			"autoscaling_enabled": schema.BoolAttribute{
-				Description: "Enables autoscaling.",
+				Description: "Enable autoscaling for the initial worker pool. (only valid during cluster creation)",
 				Optional:    true,
 			},
 			"min_replicas": schema.Int64Attribute{
-				Description: "Minimum replicas of worker nodes in a machine pool.",
+				Description: "Minimum replicas of worker nodes in a machine pool. (only valid during cluster creation)",
 				Optional:    true,
-				Computed:    true,
 			},
 			"max_replicas": schema.Int64Attribute{
-				Description: "Maximum replicas of worker nodes in a machine pool.",
+				Description: "Maximum replicas of worker nodes in a machine pool. (only valid during cluster creation)",
 				Optional:    true,
-				Computed:    true,
 			},
 			"api_url": schema.StringAttribute{
 				Description: "URL of the API server.",
@@ -237,24 +235,18 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 			},
 			"replicas": schema.Int64Attribute{
 				Description: "Number of worker/compute nodes to provision. Single zone clusters need at least 2 nodes, " +
-					"multizone clusters need at least 3 nodes.",
+					"multizone clusters need at least 3 nodes. (only valid during cluster creation)",
 				Optional: true,
-				Computed: true,
 			},
 			"compute_machine_type": schema.StringAttribute{
 				Description: "Identifies the machine type used by the default/initial worker nodes, " +
 					"for example `m5.xlarge`. Use the `rhcs_machine_types` data " +
-					"source to find the possible values.",
+					"source to find the possible values. (only valid during cluster creation)",
 				Optional: true,
-				Computed: true,
-				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.UseStateForUnknown(),
-					stringplanmodifier.RequiresReplace(),
-				},
 			},
 			"default_mp_labels": schema.MapAttribute{
 				Description: "This value is the default/initial machine pool labels. Format should be a comma-separated list of '{\"key1\"=\"value1\", \"key2\"=\"value2\"}'. " +
-					"This list overwrites any modifications made to node labels on an ongoing basis. ",
+					"(only valid during cluster creation)",
 				ElementType: types.StringType,
 				Optional:    true,
 			},
@@ -275,7 +267,7 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 			},
 			"kms_key_arn": schema.StringAttribute{
 				Description: "The key ARN is the Amazon Resource Name (ARN) of a AWS Key Management Service (KMS) Key. It is a unique, " +
-					"fully qualified identifier for the AWS KMS Key. A key ARN includes the AWS account, Region, and the key ID" + 
+					"fully qualified identifier for the AWS KMS Key. A key ARN includes the AWS account, Region, and the key ID" +
 					"(optional).",
 				Optional: true,
 				PlanModifiers: []planmodifier.String{
@@ -476,7 +468,6 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 			},
 		},
 	}
-	return
 }
 
 func (r *ClusterRosaClassicResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
@@ -1065,19 +1056,8 @@ func (r *ClusterRosaClassicResource) Update(ctx context.Context, request resourc
 	}
 
 	clusterBuilder := cmv1.NewCluster()
-	clusterBuilder, _, err := updateNodes(ctx, state, plan, clusterBuilder)
-	if err != nil {
-		response.Diagnostics.AddError(
-			"Can't update cluster",
-			fmt.Sprintf(
-				"Can't update cluster nodes for cluster with identifier: `%s`, %v",
-				state.ID.ValueString(), err,
-			),
-		)
-		return
-	}
 
-	clusterBuilder, err = updateProxy(state, plan, clusterBuilder)
+	clusterBuilder, err := updateProxy(state, plan, clusterBuilder)
 	if err != nil {
 		response.Diagnostics.AddError(
 			"Can't update cluster",
@@ -1138,11 +1118,6 @@ func (r *ClusterRosaClassicResource) Update(ctx context.Context, request resourc
 		return
 	}
 
-	// update the autoscaling enabled with the plan value (important for nil and false cases)
-	state.AutoScalingEnabled = plan.AutoScalingEnabled
-
-	// update the Replicas with the plan value (important for nil and zero value cases)
-	state.Replicas = plan.Replicas
 	object := update.Body()
 
 	// Update the state:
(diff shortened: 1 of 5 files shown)


JohnStrunk:
I'm (unsurprisingly) getting the same error as in #227 during the CI teardown phase:
```
Error: Value cannot be changed

  with rhcs_cluster_rosa_classic.rosa_sts_cluster,
  on main.tf line 78, in resource "rhcs_cluster_rosa_classic" "rosa_sts_cluster":
  78: resource "rhcs_cluster_rosa_classic" "rosa_sts_cluster"

AttributeName("compute_machine_type") is blocked for updating
```

I believe that the TF source being provided for the teardown doesn't match that used for cluster creation, and so the result is this error.

JohnStrunk:
/cc @bardielle
This is ready for review, though I still need to track down the issue w/ OpenShift CI

JohnStrunk:
/hold
CI debugging

JohnStrunk:
Looks like the CI changes yesterday(?) fixed my problem, but now there is a CI failure, unrelated to this code:
```
[0m[1mrhcs_cluster_wait.rosa_cluster: Creation complete after 58m2s[0m
[31m╷[0m[0m
[31m│[0m [0m[1m[31mError: [0m[0m[1mupdating S3 Bucket (oidc-u5k6) tags: listing resource tags (oidc-u5k6): AccessDenied: Access Denied
[31m│[0m [0m	status code: 403, request id: C3FRWTH754GZ5ENS, host id: uYOfUlEJZPsZePfjHqrLdVNIBh0PrOTxOWzgwXtr2bEo2SYycSUd6QYFvJlzACnaOncdebvNJLA=[0m
[31m│[0m [0m
[31m│[0m [0m[0m  with module.oidc_config_input_resources.module.rosa_oidc_config_resources[0].aws_s3_bucket.s3_bucket,
[31m│[0m [0m  on .terraform/modules/oidc_config_input_resources/oidc_config_resources/create_s3_bucket.tf line 2, in resource "aws_s3_bucket" "s3_bucket":
[31m│[0m [0m   2: resource "aws_s3_bucket" "s3_bucket" [4m{[0m[0m
[31m│[0m [0m
[31m╵[0m[0m
```

provider/cluster_rosa_classic_resource.go
```diff
@@ -1428,25 +1374,6 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
 	state.Domain = types.String{
 		Value: fmt.Sprintf("%s.%s", object.Name(), object.DNS().BaseDomain()),
 	}
-	state.Replicas = types.Int64{
```
bardielle, line 1431: I think it still need to be computed - if someone will edit it from rosa cli we would like to be aligned with that changes

provider/machine_pool_resource.go
```diff
@@ -247,6 +251,59 @@ func (r *MachinePoolResource) Create(ctx context.Context,
 		return
 	}
 
+	// The default machine pool is created automatically when the cluster is created.
+	// We want to import it instead of creating it.
+	if machinepoolName == defaultMachinePoolName {
```
bardielle, line 256: I think it will be more clear to move it to another method

subsystem/machine_pool_resource_test.go
```diff
@@ -1836,3 +1836,117 @@ var _ = Describe("Machine pool import", func() {
...
+				{
+				  "id": "worker",
+				  "href": "/api/clusters_mgmt/v1/clusters/123/machine_pools/worker",
+				  "kind": "MachinePool",
+				  "instance_type": "r5.xlarge",
+				  "replicas": 4
+				}`),
+			),
+		)
+		terraform.Source(`
```
bardielle, line 1937: Can you add a test for updating the default mp labels?

provider/machine_pool_resource.go
JohnStrunk: done

subsystem/machine_pool_resource_test.go
JohnStrunk: done

provider/machine_pool_resource.go
```diff
@@ -785,21 +785,49 @@ func (r *MachinePoolResource) Delete(ctx context.Context, request tfsdk.DeleteRe
...
-			),
-		)
-		return
+		// We can't delete the pool, see if it's the last one:
+		numPools, err2 := r.countPools(ctx, state.Cluster.Value)
+		if numPools == 1 && err2 == nil {
+			// It's the last one, issue warning instead of error
+			response.Diagnostics.AddWarning(
+				"Can't delete machine pool",
+				fmt.Sprintf(
```
JohnStrunk, line 794: Per @arendej, add: "ROSA Classic clusters must have at least one machine pool.", and change "Can't" -> "Cannot"

provider/machinepool/machine_pool_resource.go
```diff
@@ -437,13 +517,27 @@ func (r *MachinePoolResource) Update(ctx context.Context, req resource.UpdateReq
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Save the state:
+	diags = resp.State.Set(ctx, state)
+	resp.Diagnostics.Append(diags...)
+}
+
+func (r *MachinePoolResource) doUpdate(ctx context.Context, state *MachinePoolState, plan *MachinePoolState) (diags diag.Diagnostics) {
```
sagidayan, line 531: Can we be more explicit and return the value in each return state? Took me a while to understand what was returned...

JohnStrunk: done

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -237,24 +235,18 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
...
-				Computed: true,
 			},
 			"compute_machine_type": schema.StringAttribute{
 				Description: "Identifies the machine type used by the default/initial worker nodes, " +
 					"for example `m5.xlarge`. Use the `rhcs_machine_types` data " +
-					"source to find the possible values.",
+					"source to find the possible values. (only valid during cluster creation)",
 				Optional: true,
-				Computed: true,
-				PlanModifiers: []planmodifier.String{
```
bardielle, line 250: Why did you remove it? If the user updates it he will not get any error message and will think it was updated

provider/machinepool/machine_pool_resource.go
```diff
@@ -372,6 +384,59 @@ func (r *MachinePoolResource) Create(ctx context.Context, req resource.CreateReq
...
+	}
+	diags = r.doUpdate(ctx, existingState, state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// the 1AZ-related settings don't apply to the default machine pool
+	if state.MultiAvailabilityZone.IsUnknown() {
+		state.MultiAvailabilityZone = types.BoolNull()
```
bardielle, line 427: I think if this attribute is computed - it cannot be null

provider/machinepool/machine_pool_resource.go
```diff
@@ -372,6 +384,59 @@ func (r *MachinePoolResource) Create(ctx context.Context, req resource.CreateReq
...
+	}
+	if state.AvailabilityZone.IsUnknown() {
+		state.AvailabilityZone = types.StringNull()
```
bardielle, line 430: I think if this attribute is computed - it cannot be null

provider/machinepool/machine_pool_resource.go
```diff
@@ -372,6 +384,59 @@ func (r *MachinePoolResource) Create(ctx context.Context, req resource.CreateReq
...
+	}
+	if state.SubnetID.IsUnknown() {
+		state.SubnetID = types.StringNull()
```
bardielle, line 433: I think if this attribute is computed - it cannot be null

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
JohnStrunk: The CMT is just like the other fields that pertain to the default MP. They will be used to set the initial configuration for the worker MP, but any changes thereafter must be done via the MP resource.

The plan modifiers would cause the whole cluster to be deleted and recreated if the value is changed... I don't think that's what we want. To "change" the machine type, the pool must be recreated, not the cluster (I realize the worker pool can't actually be recreated, but I don't think destroying the cluster is a good solution).

provider/machinepool/machine_pool_resource.go
JohnStrunk: TF considers the values to be set, even if they are null.
I tried removing these lines, and the tests fail:

```
  ╷
  │ Error: Provider returned invalid result object after apply
  │ 
  │ After the apply operation, the provider still indicated an unknown value
  │ for rhcs_machine_pool.worker.availability_zone. All values must be known
  │ after apply, so this is always a bug in the provider and should be reported
  │ in the provider's own repository. Terraform will still save the other known
  │ object values in the state.
  ╵
  ╷
  │ Error: Provider returned invalid result object after apply
  │ 
  │ After the apply operation, the provider still indicated an unknown value
  │ for rhcs_machine_pool.worker.multi_availability_zone. All values must be
  │ known after apply, so this is always a bug in the provider and should be
  │ reported in the provider's own repository. Terraform will still save the
  │ other known object values in the state.
  ╵
  ╷
  │ Error: Provider returned invalid result object after apply
  │ 
  │ After the apply operation, the provider still indicated an unknown value
  │ for rhcs_machine_pool.worker.subnet_id. All values must be known after
  │ apply, so this is always a bug in the provider and should be reported in
  │ the provider's own repository. Terraform will still save the other known
  │ object values in the state.
```

Whereas, they pass w/ the code as written.

JohnStrunk: (pardon the delete and recreate... I thought I ran into a problem, but I just didn't rebuild after experimenting :man_facepalming:)

TF regards Null as a "set" value satisfying the "computed" requirement. When I remove these lines, the subsystem tests (and tests against a live cluster) fail, but as written, both pass.

Error generated by omitting the lines:
```
╷
│ Error: Provider returned invalid result object after apply
│ 
│ After the apply operation, the provider still indicated an unknown value for rhcs_machine_pool.worker.availability_zone. All values must be known after apply, so this is always a bug in the provider and
│ should be reported in the provider's own repository. Terraform will still save the other known object values in the state.
╵
```

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
bardielle: Can you test your binary code using importing cluster resource? 
According to planModifier - can you write a specific plan modifier for value cannot be changed?

provider/machinepool/machine_pool_resource.go
```diff
@@ -681,21 +779,50 @@ func (r *MachinePoolResource) Delete(ctx context.Context, req resource.DeleteReq
...
-		)
-		return
+		// We can't delete the pool, see if it's the last one:
+		numPools, err2 := r.countPools(ctx, state.Cluster.ValueString())
+		if numPools == 1 && err2 == nil {
+			// It's the last one, issue warning instead of error
+			resp.Diagnostics.AddWarning(
+				"Cannot delete machine pool",
+				fmt.Sprintf(
+					"Cannot delete the last machine pool for cluster '%s'. "+
```
amalykhi, line 789: May be I am missing something, but should there is check for the scenario:
-I have 2 machinepools: with 1 replica and and default - with 3 replicas (for multiaz cluster)
- I want to delete default machinepool
Will it fail as the last machinepool doesn't fit the cluster requirement?

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
JohnStrunk: > Can you test your binary code using importing cluster resource?

Will do.

> According to planModifier - can you write a specific plan modifier for value cannot be changed?

I'm not sure if it's possible... I'll take a look and see if I can create one.

provider/machinepool/machine_pool_resource.go
amalykhi: ACK. Thabk you @JohnStrunk

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
JohnStrunk: Cluster import works.
This PR still need an approved tag

PR #230, opened 2023-08-06
OCM-2637: [BREAKING] htpasswd IDP user list

sagidayan:
This commit changes the way the user is passing user credentials into htpasswd IDP resource.

Before this commit, a single username and password was passed. 
This commit/PR moves to `users`, a list of user credentials.

Updated docs and examples as well.

### DIFF
--- a/docs/resources/identity_provider.md
+++ b/docs/resources/identity_provider.md
@@ -82,8 +82,16 @@ Optional:
 
 Required:
 
+- `users` (Attributes List) A list of htpasswd user credentials (see [below for nested schema](#nestedatt--htpasswd--users))
+
+<a id="nestedatt--htpasswd--users"></a>
+### Nested Schema for `htpasswd.users`
+
+Required:
+
 - `password` (String, Sensitive) User password.
-- `username` (String) User name.
+- `username` (String) User username.
+
 
 
 <a id="nestedatt--ldap"></a>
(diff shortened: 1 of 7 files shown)


nirarg:
Thank you @sagidayan ,
That is look good
Did you check how it acts after provider update?
1. Create cluster before change
2. Create htpasswd IDP
3. update provider
4. Change manually the username/password format in the manifest
I'm aware that this can't be updated after creation but need to check if other partd in the manifest are functional.
Is it enough? does the terraform work as expected after this change?

nirarg:
@bardielle please also review this

sagidayan:
@nirarg 
Since there is no update implementation, tf will return `Error: Provider produced inconsistent result after apply`. This is to be expected, since tf will try to update the resource. 
As mentioned in the title of this PR `BREAKING`.

You can, however, destroy the old one, update the .tf file to the new schema and apply. Should result in a working IDP. Plus, so far, you could only have one user in the IDP, and we are not GA, So don't see the harm.
Worst case, you recreate an IDP. (one user will "suffer") for (less than) 10 sec.

This will also not work if we do support the old `username/password` attributes for compatibility, TF will still try to update and will return the same error.

Thoughts?

Actually, after #226, the error will be as described there (Update not supported)

nirarg:
Ok, got it
So as far as I understand:
1. As long as the user doesn't run terraform apply on the manifest, the old user work
2. In case user want to apply any change in the manifest (not related to the IDP) he must destroy the existing htpasswd IDPs and create new ones with the new username/password format

Need to document this known limitation
@EricPonvelle @arendej FYI

/hold cancel

sagidayan:
> Ok, got it So as far as I understand:
> 
> 1. As long as the user doesn't run terraform apply on the manifest, the old user work
> 2. In case user want to apply any change in the manifest (not related to the IDP) he must destroy the existing htpasswd IDPs and create new ones with the new username/password format
> 
> Need to document this known limitation @EricPonvelle @arendej FYI
> 
> /hold cancel

Correct.

PR #234, opened 2023-08-10
OCM-758: New rhcs_dns_domain resource

sagidayan:
This PR adds `rhcs_dns_domain`. This is the first of many changes and tweaks to support the shared VPC feature. 

Callbacks implemented:
- create
- read
- delete
- import

**Updates** of the resource are not supported since this is out of scope for the epic.

Resource reconcile will work, once [OCM-3208](https://issues.redhat.com//browse/OCM-3208) is fixed.

This will be used to reserve a DNS base domain for the cluster that will be created in the shared private VPC.

Added subsystem tests and doc file.


provider/dns_domain_resource.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (t *DNSDomainResourceType) GetSchema(ctx context.Context) (result tfsdk.Schema,
+	diags diag.Diagnostics) {
+
+	result = tfsdk.Schema{
+		Description: "Identity provider.",
+		Attributes: map[string]tfsdk.Attribute{
+			"id": {
+				Description: "Unique identifier of the identity provider.",
```
nirarg, line 68: Wrong (copy/paste) description

provider/dns_domain_resource.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Get the plan
+	plan := &DNSDomainState{}
+	diags := request.Plan.Get(ctx, plan)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	dnsDomain := resource.NewDNSDomain(r.dns_domains_client)
```
nirarg, line 133: why `dns_domains_client` isn't internal created inside `resource/dns_domain`?

internal/ocm/resource/dns_domain.go
```diff
@@ -0,0 +1,39 @@
...
+func (d *DNSDomain) Create() (*cmv1.DNSDomainsAddResponse, error) {
+	payload, err := d.builder.Build()
+	if err != nil {
+		return nil, err
+	}
+	return d.client.Add().Body(payload).Send()
+}
+
+func (d *DNSDomain) Delete(id string) error {
+	_, err := d.client.DNSDomain(id).Delete().Send()
```
nirarg, line 33: Is it the right convention to have the client inside the `internal/ocm/resource`?
This is not how it was done in the cluster
Lets discuss about what we think should be the right convention

nirarg (commented):
From what I understand, the dns_domain resource purpose also to provide the base domain, no?
Where is it implemented?

provider/dns_domain_resource.go
sagidayan: Correct, Fixed.

provider/dns_domain_resource.go
sagidayan: We have discussed this. I do think we should have all clients created in side `resource/*` BUT we are not passing all the parameters needed down the chain to create the client elsewhere. We create clients via the parent (root provider), This is the only place where we have the information.

internal/ocm/resource/dns_domain.go
sagidayan: It's not the same, It's simpler. Less logic in this resource. Don't see the benefit of exposing the `builder` in this case. Provider in this case simply needs some functions for CRUD. Passing minimal information (mainly id). No reason to add more lines of code for these small operations. IMO

But sure, We can talk about it :)

PR #238, opened 2023-08-14
Allow using empty default value (null) for availability_zones

nirarg:
We must allow null default values in order to allow using modules for this provider.

When created a module, which sets the `availability_zones = null`
I got an error during destroy that this attribute can't be changed


nirarg:
This is the error:
image

nirarg:
/retest
/hold
@bardielle I know you had doubts about this change
Please check it and if it ok, please remove the hold

PR #239, opened 2023-08-14
Change Provider environment prefix from "OCM_" to "RHCS_"

nirarg:


bardielle:
Please change it also in the makefile, README files and CI files

PR #240, opened 2023-08-14
OCM-214: handle cluster-autoscaler resources

osherdp:
Being able to create, delete and update cluster-autoscaler resources in all managed offerings of OCM (ROSA/OSD).

Checklist:
- [x] Manually test all functionalities except 'create' (that is already tested)
- [x] Adding all supported parameters
- [x] Adding tests
- [x] Supporting update function
- [x] Adding examples
- [ ] Waiting for the feature to be enabled on production (the feature-toggle is currently off for most users)


provider/provider.go
```diff
@@ -204,6 +204,7 @@ func (p *Provider) GetResources(ctx context.Context) (result map[string]tfsdk.Re
 		"rhcs_rosa_oidc_config_input": &RosaOidcConfigInputResourceType{},
 		"rhcs_rosa_oidc_config":       &RosaOidcConfigResourceType{},
 		"rhcs_dns_domain":             &DNSDomainResourceType{},
+		"rhcs_cluster_autoscaler":     &ClusterAutoscalerResourceType{},
```
nirarg, line 207: Having "cluster" included in the name might be confusing for user, as it might cause the user it related to cluster creation.
I understand that you added the "cluster" to make it clear that the autoscaler belongs to the cluster.
@arendej any thoughts here?

osherdp: naming it autoscaler would be ambiguous in my opinion

osherdp:
/hold
waiting for reviews

provider/cluster_autoscaler_state.go
```diff
@@ -0,0 +1,61 @@
...
+	MaxPodGracePeriod           types.Int64               `tfsdk:"max_pod_grace_period"`
+	PodPriorityThreshold        types.Int64               `tfsdk:"pod_priority_threshold"`
+	IgnoreDaemonsetsUtilization types.Bool                `tfsdk:"ignore_daemonsets_utilization"`
+	MaxNodeProvisionTime        types.String              `tfsdk:"max_node_provision_time"`
+	BalancingIgnoredLabels      types.List                `tfsdk:"balancing_ignored_labels"`
+	ResourceLimits              *AutoscalerResourceLimits `tfsdk:"resource_limits"`
+	ScaleDown                   *ScaleDownConfig          `tfsdk:"scale_down"`
+}
+
+type ResourceRange struct {
```
sagidayan, line 37: Maybe we should add a `Autoscaler` prefix for these structs (like you did in `AutoscalerResourceLimits`)? To avoid confusion. I know we have a major refactoring soon and it might help.

provider/provider.go
sagidayan: I agree with @osherdp here. But ill wait for @arendej response.

sagidayan (commented):
I would also maybe add a validator to all ranges, making sure `min <= max` before hitting the API

provider/cluster_autoscaler_state.go
osherdp: Sure, done!

provider/provider.go
arendej: per discussion, we can leave the name as-is. 
@EricPonvelle please point out in documentation about rhcs_cluster_autoscaler about how it is scoped to worker machinepools and not the control plane/infra.

osherdp:
> I would also maybe add a validator to all ranges, making sure `min <= max` before hitting the API

@sagidayan ptal, I added a bunch of client-side validations

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,704 @@
...
+
+				steps := []string{}
+				for _, step := range req.AttributePath.Steps() {
+					steps = append(steps, fmt.Sprintf("%s", step))
+				}
+				if resourceRange.Min.Value > resourceRange.Max.Value {
+					resp.Diagnostics.AddAttributeError(
+						req.AttributePath,
+						"Invalid resource range",
+						// fmt.Sprintf("In '%s.%s' attribute, max value must be greater or equal to min value", req.AttributePath.WithoutLastStep().LastStep(), req.AttributePath.LastStep()),
```
sagidayan, line 246: Did you forget to change to the commented line, or forgot to delete? I assume delete. :)

sagidayan (changes requested):
nit comment. But besides that, looks good to me

provider/cluster_autoscaler_resource.go
osherdp: oops, forgot that
thanks!

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+// populateAutoscalerState copies the data from the API object to the Terraform state.
+func populateAutoscalerState(object *cmv1.ClusterAutoscaler, clusterId string, state *ClusterAutoscalerState) {
+	state.Cluster = types.String{
+		Value: clusterId,
+	}
+
+	if value, exists := object.GetBalanceSimilarNodeGroups(); exists {
+		state.BalanceSimilarNodeGroups = types.Bool{Value: value}
+	} else {
+		state.BalanceSimilarNodeGroups = types.Bool{Null: true}
```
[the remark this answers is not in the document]
nirarg: Ok, I guess we can ignore this for now as this how it was done in other places and it going to be changed anyway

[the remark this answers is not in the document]
osherdp: It looks like by default it is created with ``Null: false`` so we need it (for the current sdk at least)

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+		Predicate(func(get *cmv1.ClusterGetResponse) bool {
+			return get.Body().State() == cmv1.ClusterStateReady
+		}).
+		StartContext(pollCtx)
+
+	return err
+}
+
+// populateAutoscalerState copies the data from the API object to the Terraform state.
+func populateAutoscalerState(object *cmv1.ClusterAutoscaler, clusterId string, state *ClusterAutoscalerState) {
```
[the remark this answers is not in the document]
osherdp: Basically speaking OCM doesn't come up with values that weren't set by the client. Meaning there are no defaults or something alike. In this function we take an API object and set the state based on it

Should it still be changed to ``comptued = true``? Because I think the usage for it is a bit different (IDs being generated, defaults being set by the backend, etc.)

provider/cluster_autoscaler_resource.go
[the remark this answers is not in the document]
nirarg: In tests, where you create the object manually, you right
But when terraform creates the object from the manifest, it takes care to assign null=true, in case no value is assigned

provider/cluster_autoscaler_resource.go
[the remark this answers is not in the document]
nirarg: We already had similar issue in which the backend was changed and added default value, any future change like ths might cause to break the terraform, so we must take care for it also here.
If there is no chance for updating the value, so just lets remove it from the populate, otherwise need to declare it can be computed

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+		Attributes: map[string]tfsdk.Attribute{
+			"cluster": {
+				Description: "Identifier of the cluster.",
+				Type:        types.StringType,
+				Required:    true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{
+					ValueCannotBeChangedModifier(),
+				},
+			},
+			"balance_similar_node_groups": {
```
bardielle, line 55: Can you add a default value?

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+				Type:     types.BoolType,
+				Optional: true,
+			},
+			"log_verbosity": {
+				Description: "Sets the autoscaler log level. " +
+					"Default value is 1, level 4 is recommended for DEBUGGING and " +
+					"level 6 will enable almost everything.",
+				Type:     types.Int64Type,
+				Optional: true,
+			},
```
bardielle, line 75: Can you add the default value? 
```suggestion
			"log_verbosity": {
				Description: "Sets the autoscaler log level. " +
					"Default value is 1, level 4 is recommended for DEBUGGING and " +
					"level 6 will enable almost everything.",
				Type:     types.Int64Type,
				Optional: true,
				Computed: true,
				Default:     int64default.StaticInt64(1)
			},
```

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+			"max_pod_grace_period": {
+				Description: "Gives pods graceful termination time before scaling down.",
```
bardielle, line 77: What is the default value?

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+				Description: "To allow users to schedule 'best-effort' pods, which shouldn't trigger " +
+					"Cluster Autoscaler actions, but only run when there are spare resources available.",
+				Type:     types.Int64Type,
+				Optional: true,
+			},
+			"ignore_daemonsets_utilization": {
+				Description: "Should cluster-autoscaler ignore DaemonSet pods when calculating resource utilization " +
+					"for scaling down. false by default",
+				Type:     types.BoolType,
+				Optional: true,
```
bardielle, line 91: If we will get the value from the backend it should be computed

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+	plan := &ClusterAutoscalerState{}
+	diags := request.Plan.Get(ctx, plan)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	resource := r.collection.Cluster(plan.Cluster.Value)
+	// We expect the cluster to exist. Return an error if it's not the case
+	if resp, err := resource.Get().SendContext(ctx); err != nil && resp.Status() == http.StatusNotFound {
```
bardielle, line 278: Last version we changed the `WaitTillClusterReady` function to support that case

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+			err.Error(),
+		)
+		return
+	}
+
+	state := &ClusterAutoscalerState{}
+	populateAutoscalerState(getResponse.Body(), clusterId, state)
+	diags := response.State.Set(ctx, state)
+	response.Diagnostics.Append(diags...)
+}
```
bardielle, line 468: Can be more than one autoscaler for a cluster? 
I guess it's only one per cluster (otherwise you should set an identifier), and if this is the case you can use the "simple" import state 
```suggestion
	func (r *ClusterRosaClassicResource) ImportState(ctx context.Context, request resource.ImportStateRequest,
	response *resource.ImportStateResponse) {
	tflog.Debug(ctx, "begin importstate()")

	resource.ImportStatePassthroughID(ctx, path.Root("cluster"), request, response)
}

```

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+
+func (r *ClusterAutoscalerResource) waitForClusterToGetReady(ctx context.Context, clusterId string) error {
```
bardielle, line 470: As I mentioned before - please use the common function 
`common.WaitTillClusterReady`

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+	state.MaxNodeProvisionTime = common.EmptiableStringToStringType(object.MaxNodeProvisionTime())
+
+	if value, exists := object.GetBalancingIgnoredLabels(); exists {
+		state.BalancingIgnoredLabels = common.StringArrayToList(value)
+	} else {
+		state.BalancingIgnoredLabels = types.List{Null: true}
+	}
+
+	if object.ResourceLimits() != nil {
+		state.ResourceLimits = &AutoscalerResourceLimits{}
```
bardielle, line 537: can you refactor it to a new function?

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+	} else {
+		state.BalanceSimilarNodeGroups = types.Bool{Null: true}
+	}
+
+	if value, exists := object.GetSkipNodesWithLocalStorage(); exists {
+		state.SkipNodesWithLocalStorage = types.Bool{Value: value}
+	} else {
+		state.SkipNodesWithLocalStorage = types.Bool{Null: true}
+	}
+
```
bardielle, line 503: can you export it to a common function? 

```suggestion
       state.SkipNodesWithLocalStorage =common.setBoolOptionalValue(object.GetSkipNodesWithLocalStorage())
       ......
func setBoolOptionalValue(value *bool, exists bool) types.Bool{      
       if exists {
              return types.Bool{Value: value}
       }
       return types.Bool{Null: true}
```

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+	if value, exists := object.GetLogVerbosity(); exists {
+		state.LogVerbosity = types.Int64{Value: int64(value)}
+	} else {
+		state.LogVerbosity = types.Int64{Null: true}
+	}
```
bardielle, line 508: can you export it to a common function?

 state.LogVerbosity =common.setBoolOptionalValue(object.GetSkipNodesWithLocalStorage())
       ......
func setBoolOptionalValue(value *bool, exists bool) types.Bool{      
       if exists {
              return types.Bool{Value: value}
       }
       return types.Bool{Null: true}

```suggestion
	state.LogVerbosity = common.GetInt64OptionalValue( object.GetLogVerbosity())
....
```

provider/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,703 @@
...
+						},
+					},
+				)
+			}
+		}
+	}
+
+	if object.ScaleDown() != nil {
+		state.ScaleDown = &AutoscalerScaleDownConfig{}
+
```
bardielle, line 582: Can you refactor it an move it to a different function?

provider/cluster_autoscaler_resource_test.go
```diff
@@ -0,0 +1,264 @@
...
+					DelayAfterAdd("3h").
+					DelayAfterDelete("4h").
+					DelayAfterFailure("5h")).Build()
+
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(autoscaler).To(Equal(expectedAutoscaler))
+		})
+
+		It("successfully converts when all state fields are null", func() {
```
bardielle, line 240: add more test cases - import, creating multiple autoscaler to one cluster and failed, 
cluster not exist, or not ready 
update test

subsystem/cluster_autoscaler_resource_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+			terraform.Source(`
+				resource "rhcs_cluster_autoscaler" "cluster_autoscaler" {
+					cluster = "123"
+				}
+			`)
+			Expect(terraform.Apply()).ToNot(BeZero())
+		})
+
+		It("fails if OCM backend fails to create the object", func() {
```
bardielle, line 92: what about creating more than one autoscale object ?

sagidayan:
Ill take over. Once I create a new PR from @osherdp 's branch, Ill close this one.

sagidayan:
Closing this PR, Code has moved to #372 instead

PR #242, opened 2023-08-16
OCM-2643: Allow default ingress configurations

tsorya:
Tested with 4.14.0-ec.2 in candidate channel


provider/cluster_rosa_classic_resource.go
```diff
@@ -1033,6 +1054,18 @@ func (r *ClusterRosaClassicResource) Update(ctx context.Context, request tfsdk.U
 		return
 	}
 
+	err = r.updateIngress(ctx, state, plan)
```
bardielle, line 1057: Prefet to call it before calling populate for 2 reasons: 
1. I'm working on migration to TF SDK V2 and there we will not have a struct (we will have a similar function that will get the connection )
2. I don't think this is the mission of apopulate_state...

tsorya: Actually populate here populates state before update and according to it decides if we need to update or not. I don't think i can run this before first populate

provider/cluster_rosa_classic_resource.go
```diff
@@ -1792,6 +1898,30 @@ func getThumbprint(oidcEndpointURL string, httpClient HttpClient) (thumbprint st
 	return sha1Hash(cert.Raw)
 }
 
+func (r *ClusterRosaClassicResource) getDefaultIngress(ctx context.Context, state *ClusterRosaClassicState) error {
+	ingresses, err := r.clusterCollection.Cluster(state.ID.Value).Ingresses().List().SendContext(ctx)
+	if err != nil {
+		return err
+	}
+	if ingresses.Size() > 0 {
```
gdbranco, line 1906: It is not possible to create a second ingress anymore, but if for some reason a customer imported or created a cluster using TF when it was still possible they might have more than one.

Might need to filter if the ingress being retrieved is actually the default one

provider/cluster_rosa_classic_state.go
```diff
@@ -85,3 +86,11 @@ type AdminCredentials struct {
 	Username types.String `tfsdk:"username"`
 	Password types.String `tfsdk:"password"`
 }
+
+type DefaultIngress struct {
+	RouteSelectors           types.Map    `tfsdk:"route_selectors"`
```
gdbranco, line 91: Not sure how it is going to work with TF, but there is also 2 extra attributes only for day 2 operations. Which are the components route hostname and components route tls secret ref

tsorya: Wanted to start with easy ones, will do it as second pr

tsorya: It will be following PR

provider/cluster_rosa_classic_resource.go
```diff
@@ -347,6 +347,11 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result
 				Optional:   true,
 				Validators: proxyValidators(),
 			},
+			"default_ingress": {
```
nirarg, line 350: Please generate this attribute to the resource docs, by using [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs)

provider/cluster_rosa_classic_resource.go
```diff
@@ -347,6 +347,11 @@ func (t *ClusterRosaClassicResourceType) GetSchema(ctx context.Context) (result
...
+				Description: "default_ingress",
```
nirarg, line 351: Please add more informative description here 
something similar to the CLI: https://github.com/openshift/rosa/blob/b694d0272c9a0c77c564ee0987a5b7fad8b41ccc/cmd/create/ingress/cmd.go#L45

nirarg (commented):
Great work, just few comments
@sagidayan can you please also take a look?

provider/cluster_rosa_classic_resource.go
```diff
@@ -1739,6 +1772,11 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
 		Value: object.Region().ID(),
 	}
 
+	err := populateDefaultIngress(ctx, state, clusterCollection.Cluster(state.ID.Value).Ingresses())
```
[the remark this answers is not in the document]
tsorya: In order to have same logic for default ingress param both for create and update. In create you must bring it after cluster is created so i set it in populate.
Another way i must add getIngress for update/get/create, i assume it doesn't worth one api call on update

provider/cluster_rosa_classic_resource.go
tsorya: It seems generate docs creates many files and changes their names, not relevant to my changes
`deleted:    docs/resources/identity_provider.md`
`docs/resources/rhcs_identity_provider.md`

nirarg: So please just add the relevant lines for your change, it need to be done semi manual

bardielle:
/hold 
would be merged after framework changes

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -330,6 +331,11 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 				Optional:    true,
 				Validators:  []validator.Object{proxy.ProxyValidator()},
 			},
+			"default_ingress": schema.SingleNestedAttribute{
+				Description: "default_ingress",
```
bardielle, line 335: please change the description

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common/attrvalidators"
+)
+
+const lowestDefaultIngressVer = "4.14.0-0"
```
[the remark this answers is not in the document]
tsorya: Not sure what you mean by "at least move it to be defined in the sdk library" ?

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,269 @@
...
+			Description: "Components route TLS secret reference for oauth, console, download.",
+			Optional:    true,
+		},
+	}
+}
+
+func PopulateDefaultIngress(ctx context.Context, state *DefaultIngress,
+	client *cmv1.IngressesClient) (*DefaultIngress, error) {
+
+	// in case default ingress is not part of state no need to populate it
```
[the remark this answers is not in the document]
tsorya: Bringing default ingress requires one more api call that has no point if you don't need it.
Another problem will be computed/optional that doesn't work good on custom structs and creates issues actually. So my idea was just to not bring it in case it was not required

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,269 @@
...
+			} else {
+				state.WildcardPolicy = types.StringNull()
+			}
+			rnmop, ok := ingress.GetRouteNamespaceOwnershipPolicy()
+			if ok {
+				state.NamespaceOwnershipPolicy = types.StringValue(string(rnmop))
+			} else {
+				state.NamespaceOwnershipPolicy = types.StringNull()
+			}
+			hostname, ok := ingress.GetClusterRoutesHostname()
```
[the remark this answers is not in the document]
tsorya: Did it before we talked about optional/computed fields, let me fix it

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
nirarg: Is this value can be changed in the future?
If yes, it need to be managed from one place, for both cli and tf provider, no?

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
nirarg: But what about the "import" case, what happen if you want to import the cluster with this value?

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
tsorya: Can  you point me to the right place? In anycase lets do it with another PR but need link to the place as i have no idea about the repo

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
tsorya: If you will need to change anything in default ingress, you will add this value in order to update it.

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
tsorya: Actually i am not sure it will work removing null, how we are going to remove variable in that case? 
Lets say we want to remove hostname so we are removing it and it should be nil how we are gonna get it without setting nil?

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
tsorya: The issue is that you can't set custom types as optional/computed as you can't have them unknown in case you know how to do it i will be glad to change it

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
gdbranco: This is managed by the api, if the user supplies an older version he'll face an error. So it should be ok to not have this in TF. It's there in the CLI mostly for the interactive mode and alerting the user, for TF he'll be alerted through the api error

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,269 @@
...
+			},
+		},
+		"route_selectors": schema.MapAttribute{
+			Description: "Route Selectors for ingress. Format should be a comma-separated list of 'key=value'. " +
+				"If no label is specified, all routes will be exposed on both routers." +
+				"For legacy ingress support these are inclusion labels, otherwise they are treated as exclusion label.",
+
+			ElementType: types.StringType,
+			Optional:    true,
+			Validators:  []validator.Map{NotEmptyMapValidator()},
```
gdbranco, line 51: If this is optional shouldn't it be able to be empty? What does the not empty map validator actually does does?

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		"excluded_namespaces": schema.ListAttribute{
+			Description: "Excluded namespaces for ingress. Format should be a comma-separated list 'value1, value2...'. " +
+				"If no values are specified, all namespaces will be exposed.",
+			ElementType: types.StringType,
+			Optional:    true,
+			Validators: []validator.List{
+				listvalidator.SizeAtLeast(1),
```
gdbranco, line 59: Similar comment, if it is optional does this validate that it needs at least one element?

provider/defaultingress/ingress.go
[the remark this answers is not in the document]
gdbranco: We are accepting the values as part of the cluster creation request body, which means that for TF it needs to be in place for the state. Prior to this there was no possibility of changing the default values, so it was available similarly to the default machine pool (worker) where it would be a sub resource.

provider/defaultingress/ingress.go
tsorya: it verifies that if you provide this param it should not be clean list

provider/defaultingress/ingress.go
tsorya: it verifies that if you provide this param it should not be empty map

gdbranco: Oh if you provide it, gotcha

provider/clusterrosaclassic/validators.go
```diff
@@ -63,7 +63,7 @@ var propertiesValidator = attrvalidators.NewMapValidator("properties map should
 			return
 		}
 
-		for k, _ := range propertiesElements {
+		for k := range propertiesElements {
```
bardielle, line 66: I think it's a mistake

tsorya: was changed by make fmt

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,453 @@
...
+	// it we need to bring it first as we need to set specific id
+	if common.IsStringAttributeUnknownOrEmpty(state.Id) {
+		err := r.populateDefaultIngress(ctx, state)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !reflect.DeepEqual(state, plan) {
+		err := validateDefaultIngress(ctx, plan)
```
nirarg, line 312: Is there any reason why not adding this validation in the plan stage?

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,453 @@
...
+		ingressBuilder.RouteNamespaceOwnershipPolicy(cmv1.NamespaceOwnershipPolicy(state.NamespaceOwnershipPolicy.ValueString()))
+	}
+	// LoadBalancer type can't be empty
+	if !common.IsStringAttributeUnknownOrEmpty(state.LoadBalancerType) {
+		ingressBuilder.LoadBalancerType(cmv1.LoadBalancerFlavor(state.LoadBalancerType.ValueString()))
+	}
+	return ingressBuilder
+}
+
+func ingressResource() map[string]schema.Attribute {
```
nirarg, line 377: Please merge this function into the Schema function, to keep the same convention as in other resources

provider/defaultingress/notemptymap_validator.go
nirarg, line 1: Please move this validator under common directory
This validator can be re-used in other resorces

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,453 @@
+package defaultingress
```
gdbranco, line 1: Would you mind if we call it default application ingress?

Today we don't expose, but there's also an API ingress. So if we ever end up opening that it will be more complicated to adjust the terms :thinking:

provider/defaultingress/ingress.go
```diff
@@ -0,0 +1,453 @@
...
+		if err != nil {
+			return err
+		}
+		if plan == nil {
+			plan = &DefaultIngress{}
+		}
+
+		ingressBuilder := getDefaultIngressBuilder(ctx, plan)
+
+		if !reflect.DeepEqual(state.ClusterRoutesHostname, plan.ClusterRoutesHostname) {
```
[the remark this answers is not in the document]
tsorya: I don't want to handle things like sending empty string, prefer not to send data if it was not required

provider/defaultingress/ingress.go
tsorya: not sure i understand, i assume it can be optimization but currently not sure how

nirarg: All validations that can be done in the plan mode (before calling the backend) need to be done than
I don't think this critical for now, lets open ticket to change it later

provider/defaultingress/ingress.go
nirarg: I agree that `ingress` might be better package name here
The `defaultingress` might confuse
@tsorya any thoughts here?

tsorya: As we are working only with default one I think we should stick with default in name. When we will have an option to work with not only default one we can change the package and add default to variables that can be set

nirarg: Ok, I'm fine with pushing like this
Anyway, the package name is internal detail and can be changed without any issue

nirarg:
@tsorya thank you for the great work here
/approve
/lgtm

PR #253, opened 2023-08-24
OCM-3102: Ocm 3102 main

bardielle:
Fixing a case of updating the proxy attribute this the first time it was set (and it's null)


osherdp (approved):

sagidayan:
Jira ticket? :)

PR #254, opened 2023-08-24
Sdk v2

bardielle:


build/ci-tf-e2e.Dockerfile
```diff
@@ -15,7 +15,7 @@ ENV PATH="/usr/local/go/bin:${PATH}"
 ENV GOPATH=/usr/local/go
 ENV TEST_OFFLINE_TOKEN=""
 
-# terraform-provider-rhcs repo
+# terraform-cprovider-rhcs repo
```
sagidayan, line 18: Why this change? (this affects our CI as well)

bardielle: By a mistake - thanks I fixed it

internal/rhcs/cluster/clusterschema/proxy_schema.go
```diff
@@ -0,0 +1,144 @@
+package clusterschema
+
+import (
+	"fmt"
```
tsorya, line 4: I think imports should be formatted as :
go imports
external imports
internal import
I think imports should be formatted as :
go imports
external imports
internal import

at all places

internal/rhcs/cluster/clusterschema/sts.go
```diff
@@ -0,0 +1,163 @@
...
+func ExpandStsFromResourceData(resourceData *schema.ResourceData) *Sts {
+	list, ok := resourceData.GetOk("sts")
+	if !ok {
+		return nil
+	}
+
+	return ExpandStsFromInterface(list)
+}
+
+func ExpandStsFromInterface(i interface{}) *Sts {
```
tsorya, line 112: maybe we can changes all expands to use https://github.com/mitchellh/mapstructure ?
I assume it always has map[string]interface? so it can be useful and pretty easy to use something more generic?

PR #256, opened 2023-09-03
OCM-3124: fix for additional_trust_bundle resulting in an error

### LINKED ISSUE #229: Provider produced inconsistent result after apply - additional_trust_bundle
When providing additional trusted certificates via the `proxy.additional_trust_bundle` attribute the provider attempts to store a different value in the state than what Terraform expects. Causing Terraform to exit with the error (I've replaced the certificate bodies with ...)

```shell
rhcs_cluster_rosa_classic.rosa_sts_cluster: Creating...

 Error: Provider produced inconsistent result after apply
 
 When applying changes to rhcs_cluster_rosa_classic.rosa_sts_cluster, provider "provider[\"registry.terraform.io/terraform-redhat/rhcs\"]" produced an unexpected new value: .proxy.additional_trust_bundle: was cty.StringVal("-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n"), but now cty.StringVal("REDACTED").

 This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

I'm using Terraform v1.5.3 with rhcs version 1.2.2.

sagidayan:
Since OCM returns `REDUCTED` as values of sensitive content, the provider was populating `REDUCTED` as the new state for `additional_trust_bundle`. 

This PR ensures that if there is a value in the state, do not override value from backend.

Should fix #229

### DIFF
--- a/provider/cluster_rosa_classic_resource.go
+++ b/provider/cluster_rosa_classic_resource.go
@@ -1660,7 +1660,7 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster, 
 	}
 
 	trustBundle, ok := object.GetAdditionalTrustBundle()
-	if ok {
+	if ok && common.IsStringAttributeEmpty(state.Proxy.AdditionalTrustBundle) {
 		if state.Proxy == nil {
 			state.Proxy = &Proxy{}
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.4 (Sep 4, 2023)
+-ENHANCEMENTS:
+* Fix for "Provider produced inconsistent result after apply" error when setting proxy.additional_trust_bundle
+
 ## 1.2.3 (Aug 24, 2023)
 -ENHANCEMENTS:
 * Fixed a bug in cluster_rosa_resource -Terraform provider panic after adding additional CA bundle to ROSA cluster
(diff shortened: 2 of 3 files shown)


nirarg (commented):
Can we have subsystem to validate this?

provider/cluster_rosa_classic_resource.go
```diff
@@ -1657,7 +1657,7 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
 	}
 
 	trustBundle, ok := object.GetAdditionalTrustBundle()
-	if ok {
+	if ok && common.IsStringAttributeEmpty(state.Proxy.AdditionalTrustBundle) {
```
bardielle, line 1660: Please also open a PR for 1.2-prerelease branch (we would like to publish a fix)

sagidayan: Cant we cherry-pick after the merge?

osherdp: you can type right now ``<slash>cherry-pick v1.2.0-prerelease.1-branch`` and it will get applied after the merge

bardielle: Also, please add a commit with the changelog changes...

PR #257, opened 2023-09-04
OCM-3005: Fix allow private cluster without private link

nirarg:
Change the terraform provider logic to align with the logic in the backend

The following cases are allowed:
1. private cluster with private link
2. public cluster without private link
3. private cluster without private link

Public cluster with private link is not allowed

### DIFF
--- a/subsystem/cluster_resource_rosa_test.go
+++ b/subsystem/cluster_resource_rosa_test.go
@@ -1947,7 +1947,7 @@ var _ = Describe("rhcs_cluster_rosa_classic - create", func() {
 		Expect(terraform.Apply()).NotTo(BeZero())
 
 	})
-	It("Creates cluster with aws subnet ids & private link", func() {
+	It("Creates private cluster with aws subnet ids without private link", func() {
 		// Prepare the server:
 		server.AppendHandlers(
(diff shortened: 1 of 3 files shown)


arendej:
Critical correction:

The following cases are allowed:

private cluster with private link
public cluster without private link
Public cluster with private link
Private cluster without private link is not allowed

subsystem/cluster_resource_rosa_test.go
```diff
@@ -2026,7 +2035,7 @@ var _ = Describe("rhcs_cluster_rosa_classic - create", func() {
 		`)
 		Expect(terraform.Apply()).To(BeZero())
 	})
-	It("Creates cluster with aws subnet ids & private & private link", func() {
+	It("Creates private cluster with aws subnet ids & private link", func() {
```
bardielle, line 2038: I think we missed here the test (that should be failed) not Private with private link

jharrington22:
> Critical correction:
> 
> The following cases are allowed:
> 
> private cluster with private link public cluster without private link Public cluster with private link Private cluster without private link is not allowed

We do not support a  "Public cluster with private link"

This PR looks good to me.

PR #263, opened 2023-09-07
Fix bad text replacements

osherdp:
* reverting ``cprovider -> provider``
* reverting ``clusterservice -> provider``

### DIFF
--- a/internal/rhcs/idps/gitlab.go
+++ b/internal/rhcs/idps/gitlab.go
@@ -2,8 +2,9 @@ package idps
 
 import (
 	"fmt"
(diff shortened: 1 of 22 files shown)


internal/rhcs/idps/gitlab.go
```diff
@@ -2,9 +2,10 @@ package idps
 
 import (
 	"fmt"
-	common2 "github.com/terraform-redhat/terraform-provider-rhcs/internal/rhcs/common"
 	"net/url"
 
+	common2 "github.com/terraform-redhat/terraform-provider-rhcs/internal/rhcs/common"
```
bardielle, line 7: Please also replace common2 to be common

PR #268, opened 2023-09-10
Move provider to new "provider" package

nirarg:
This is done to allow testing only part of the resources 

Include also moving resources into their own package
* clusterwaiter
* dnsdomain
* clusterrosaclassic 

Include also moving data sources into their own package
* cloudprovider


fwprovider/provider.go
```diff
@@ -193,9 +195,9 @@ func (p *Provider) Configure(ctx context.Context, req tfprovider.ConfigureReques
 // Resources returns the resources supported by the provider.
 func (p *Provider) Resources(ctx context.Context) []func() resource.Resource {
 	return []func() resource.Resource{
-		NewClusterWaiterResource,
-		NewDNSDomainResource,
-		NewClusterRosaClassicResource,
+		clusterwaiter.NewClusterWaiterResource,
+		dnsdomain.NewDNSDomainResource,
+		// clusterwaiter.NewClusterRosaClassicResource,
```
osherdp, line 200: why is it commented-out?

fwprovider/cloudprovider/cloud_provider_state.go
osherdp, line 1: what is 'fw'?

osherdp (changes requested):

fwprovider/provider.go
nirarg: Because its still not working
It need to be uncomment after @bardielle fix

fwprovider/cloudprovider/cloud_provider_state.go
nirarg: Ok, I also started thinking about it
fw is coming to mark framework, but now I also saw there was another place it was marked with "tfr", so it might be confusing
What do you say about renaming the old provider package to oldprovider, as a temporary name untl all its content will be removed? than we can use the provider package here

fwprovider/provider.go
osherdp: coolz
can you add a TODO at least?

nirarg: Done

osherdp (approved):

osherdp:
@nirarg please rename the PR title

osherdp (approved):

PR #274, opened 2023-09-11
Common changes and added validators

bardielle:
1. Added validators ( admin_creds, properties, proxy )- to share the concept
2. Update the common with common functionality 

The common tests are passing

### DIFF
--- a/provider/clusterrosaclassic/admin_creds_validator.go
+++ b/provider/clusterrosaclassic/admin_creds_validator.go
@@ -1,0 +1,63 @@
+package clusterrosaclassic
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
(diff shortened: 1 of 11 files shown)


provider/clusterrosaclassic/admin_creds_validator.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
+)
+
+// atLeastValidator validates that an integer Attribute's value is at least a certain value.
+type adminCredsValidator struct {
+}
+
+// Description describes the validation in plain text formatting.
+func (v adminCredsValidator) Description(_ context.Context) string {
+	return fmt.Sprintf("proxy map should not include an hard coded OCM proxy")
```
bardielle, line 18:
```suggestion
	return fmt.Sprintf("admin creds validators")
```

PR #276, opened 2023-09-12
Moving oidc_config to new framework

tsorya:
Moving oidc_config to new framework

subsystem passes

### DIFF
--- a/provider/common/helpers.go
+++ b/provider/common/helpers.go
@@ -17,6 +17,12 @@ limitations under the License.
 package common
 
 import (
+	"bytes"
+	"crypto/sha1"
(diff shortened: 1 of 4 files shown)


provider/groupmembership/group_membership_state.go
osherdp, line 1: Can you clarify what happens here? Are we deleting oidc-config resource and github mistakenly considers it a rename to group-membership resource?

tsorya: It is mess , sec i am fixing it

tsorya: fixed

PR #280, opened 2023-09-12
Moving groups to new framework

tsorya:


nirarg:
@tsorya  please rebase it now and I will merge it immediately after

PR #286, opened 2023-09-13
IDP changes to new framework

sagidayan:


provider/identityprovider/identity_provider_resource.go
```diff
@@ -23,108 +23,122 @@ import (
...
 
-	// Create the resource:
-	result = &IdentityProviderResource{
-		collection: collection,
+	if diag.HasError() {
+		return
 	}
 
-	return
+	//TODO: add validations
```
bardielle, line 137: Please add the validation

sagidayan: I am uncertain if we need in this level. We have validations for all the rest. The only things without validations are `cluster` `id` and `name`...  Ill remove this function

sagidayan: On second thought, maybe we want to be sure just one of the idp "providers/methods" are passed. (this validation was not present in the previous implementation.

JohnStrunk: I don't want to hold up this PR, as the main focus is to get the code ported over.
This resource has many validations in it, and I think a fair number of them can be converted to use some of the framework's built-in validators.
Examples:
- htpassword username and password checks can use `stringvalidator.RegexMatches()` and the min/max length validators.
- The mappingmethod validator can be a `stringvalidator.OneOf()`

I suspect we can also cut a fair bit of logic in this and other resources by adding ConflictsWith() and/or ExactlyOneOf()

bardielle: Added a ticket https://issues.redhat.com/browse/OCM-3735

provider/identityprovider/identity_provider_resource.go
```diff
@@ -391,13 +390,17 @@ func (r *IdentityProviderResource) Read(ctx context.Context, request resource.Re
...
+			state.Github.Teams = types.ListUnknown(types.StringType)
 		}
 		orgs, ok := githubObject.GetOrganizations()
 		if ok {
 			state.Github.Organizations, err = common.StringArrayToList(orgs)
 			if err != nil {
 				response.Diagnostics.AddError("failed to convert string slice to tf list", "GitHub Organizations conversion failed")
 			}
+		} else {
+			state.Github.Organizations = types.ListUnknown(types.StringType)
```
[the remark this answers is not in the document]
bardielle: Added a ticket for that: https://issues.redhat.com/browse/OCM-3734

PR #288, opened 2023-09-14
Fix machine pool resource attributes

bardielle:
@JohnStrunk


bardielle:
Closing it since we need a refactoring in that area

PR #291, opened 2023-09-14
Fix machine pool validation

JohnStrunk:
- This fixes the problem of TF deciding the machine pool needs to be replaced unnecessarily.
  - There is a change here where az and subnet will now have empty ("") values instead of null. This is necessary in order for UseStateForUnknown to allow us to avoid replacing due to (known after apply).
- It adds some static validation of parameters, but isn't really able to remove much of the run-time validation because it depends on the state of the cluster.
- Added some static validation for a couple other parameters

### DIFF
--- a/provider/machinepool/machine_pool_resource.go
+++ b/provider/machinepool/machine_pool_resource.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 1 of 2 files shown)


provider/machinepool/machine_pool_resource.go
```diff
@@ -778,6 +789,25 @@ func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *Mac
...
+	// Due to RequiresReplace(), we need to ensure these fields always have a
+	// value, even if it's empty. It will be empty if the cluster is multi-AZ or
+	// if the other (AZ/subnet) value is set. We don't need to set
+	// MultiAvailibilityZone here, because it's set in the validation function
+	// during create.
+	azs := object.AvailabilityZones()
+	if len(azs) == 1 {
+		state.AvailabilityZone = types.StringValue(azs[0])
+	} else {
+		state.AvailabilityZone = types.StringValue("")
```
bardielle, line 802: What about setting a default string in the attribute scop also?

bardielle:
Looks good overall 
Had one comment about the defaults

provider/machinepool/machine_pool_resource.go
JohnStrunk: TIL... defaults. :smile:

Trying to set defaults is causing problems... specifically, it treats them like the user specified the value (i.e., we can't set a different one later in processing). For the single AZ cases where we do need to assign it, we can't.

The docs also say that defaults are applied after validation, so they don't really simplify much there either. I looked at assigning default False to the optional bools (autoscaling, use_spot), but it didn't end up simplifying the runtime validation logic.

bardielle: OK I understand :) 
thank you for your clarification

PR #303, opened 2023-09-20
OCM-3795: Adding http tokens default to terraform state

tsorya:


amalykhi:
Verified with the build from the branch. Works as expected
/lgtm

PR #309, opened 2023-09-21
Fix update proxy in cluster rosa

tsorya:
Currently if you try to remove param from proxy you will fail as nothing actually is sent to ocm,
same if you try to remove full proxy object. 
This change will send proxy params as empty string in case they were deleted


osherdp:
@tsorya can you add to PR description what's exactly is broken?

PR #312, opened 2023-09-26
OCM-3828: HTPasswd IDP unique username validation

sagidayan:


osherdp (approved):

provider/identityprovider/htpasswd.go
```diff
@@ -86,3 +89,24 @@ func CreateHTPasswdIDPBuilder(ctx context.Context, state *HTPasswdIdentityProvid
...
+func uniqueUsernameValidator() validator.List {
+	return attrvalidators.NewListValidator("userlist unique username", func(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) {
+		usersList := req.ConfigValue
+		htusers := []HTPasswdUser{}
+		err := usersList.ElementsAs(ctx, &htusers, true)
+		if err != nil {
+			resp.Diagnostics.AddAttributeError(req.Path, "Invalid list conversion", "Failed to parse userlist")
+			return
+		}
+		usernames := make(map[string]bool)
```
nirarg, line 102: I think better to use here array (not map) and for each name check if already exists in the array

nirarg: Ok, never mind, this is good

sagidayan: :)
I do think that O(1) in finding in this case is "less expensive" than O(n) of memory...

PR #315, opened 2023-09-26
OCM-3909: Fix availability zones to ListNull in case nothing returned

nirarg:
In case the user doesn't provide the availability zone, cluster create returnes before it ready and with empty azs list.
This change is done in order to avoid from the following error:
Screenshot from 2023-09-26 11-01-31


JohnStrunk (approved):
/lgtm

PR #317, opened 2023-09-27
Handling populate relavant values in machine pool

tsorya:


sagidayan:
I think this is not the behavior we want. I'll take this offline 😉 .

/hold

provider/machinepool/machine_pool_resource.go
```diff
@@ -544,14 +544,9 @@ func (r *MachinePoolResource) Update(ctx context.Context, req resource.UpdateReq
 
 	object := update.Body()
 
-	// update the autoscaling enabled with the plan value (important for nil and false cases)
-	state.AutoScalingEnabled = plan.AutoScalingEnabled
```
nirarg, line 548: While AutoScalingEnabled can be not computed, I think Replicas need to be computed because when user doesn't declare value we put default =2
Anyway this should be done in populate and not here.

@JohnStrunk @bardielle @sagidayan what do you think?

I hope we will have here one convention how those cases should be handled

provider/machinepool/machine_pool_resource.go
```diff
@@ -544,14 +544,9 @@ func (r *MachinePoolResource) Update(ctx context.Context, req resource.UpdateReq
...
-	// update the Replicas with the plan value (important for nil and zero value cases)
-	state.Replicas = plan.Replicas
-
 	// Save the state:
-	r.populateState(object, state)
-	diags = resp.State.Set(ctx, state)
+	r.populateState(object, plan)
+	diags = resp.State.Set(ctx, plan)
```
nirarg, line 549: Good catch.

I think its a bit confusing that in some places we assign the Plan into parameter called state.
(for example in Create and populate functions)
We should refactor the code to make it more consistent and clear

PR #319, opened 2023-09-28
Add terraform cluster destroy step

shellymiron:
Just added a function for destroying the account roles resources and the cluster itself after it deployed by profile.


xueli181114:
Thanks Shelly, it's not what can be used. I have created the destroy steps. Let's close this PR.

PR #321, opened 2023-10-01
Enhance all of the test frame and clean the codes to make it stronger. And…

xueli181114:
What I did in the PR:
* Clean up the manifests of cluster creation
* Move operator-roles and oidc-provider out from cluster manifests to aws
* Enhance the code to make it run stable
* Make day1 creation work
* Make destroy step work
* Fill README.md about case contribution and how to make the cases run
* Build make step in MakeFile for CI building
* Fill output for required manifests in day1
* Remove garbage testing code
image


Makefile
```diff
@@ -125,3 +125,12 @@ destroy_folder: install
 .PHONY: binary
 binary:
 	podman run --pull=always --rm registry.ci.openshift.org/ci/rhcs-tf-bin:latest cat /root/terraform-provider-rhcs > ~/terraform-provider-rhcs && chmod +x ~/terraform-provider-rhcs
+
+.PHONY: e2e_testnew
+e2e_testnew: tools install
```
osherdp, line 130: I would address the naming of the new target right away in this PR. Having 'new' in the name will eventually not be true and isn't meaningful enough for people to understand what's the difference between this and ``e2e_test``
Maybe we should rename the ``e2e_test -> e2e_sanity_test`` or something alike? cc @bardielle

tests/utils/helper/helper.go
```diff
@@ -65,6 +65,19 @@ func DigArray(object interface{}, keys ...interface{}) []interface{} {
 	return result
 }
 
+func DigArrayToString(object interface{}, keys ...interface{}) []string {
+	value := Dig(object, keys)
+	//ExpectWithOffset(1, value).ToNot(BeNil())
```
osherdp, line 70: lots of commented-out code. can we get rid of those throughout this PR?

Makefile
xueli181114: @osherdp How about name the new one with to `e2e_test_full_cycle` ? I'd like not to change existing one with current CI still running.

tests/utils/helper/helper.go
xueli181114: Good point, will remove the commented lines

Makefile
osherdp: The CI doesn't use ``e2e_test``, so there's no concern about this target being used somewhere else

xueli181114:
@amalykhi  @shellymiron @nirarg Can you please review?

nirarg (commented):
Your PR includes many unrelated changes
Please rebase and fix it

xueli181114:
@nirarg Rebase done.
@amalykhi @shellymiron it needs your reviewing again :)

tests/README.md
```diff
@@ -44,6 +44,29 @@ tests
...
+* Create the case in terraform-provider-rhcs/tests/e2e/<feature name>_test.go
+* Label the case with ***CI.Day1Post***
+* Label the case with importance ***CI.Critical***
+* Don't need to run creation step, just in BeforeEach step call function  ***PrepareRHCSClusterByProfileENV()*** it will load the clusterID prepared
+* Code for checking steps  only in the case
+### Contribute to day2
+* Create the case in terraform-provider-rhcs/tests/e2e/<feature name>_test.go
+* Label the case with ***CI.Day2***
+* Label the case with importance ***CI.Critical*** or ***CI.High***
+* Don't need to run creation step, just in BeforeEach step call function  ***PrepareRHCSClusterByProfileENV()*** it will load the clusterID prepared
```
amalykhi, line 62: Should I add PrepareRHCSClusterByProfileENV() before each TC? I thought the cluster should be prepared as previous step, and Day 2 cases rely that the cluster exist.

tests/README.md
```diff
@@ -44,6 +44,29 @@ tests
 
 ## Contibute to terraform rhcs provider tests
 Please read the structure and contribte code to the correct place
+### Contribute to day1 case
+* Define manifests files under terraform-provider-rhcs/tests/tf-manifests/rhcs/clusters
+* Define profile in terraform-provider-rhcs/tests/ci/profiles/tf_cluster_profile.yml
+* If any other data can be created by terrafom provder, define steps under terraform-provider-rhcs/tests/tf-manifests/<provider>/<resource name>
+* Define prepare steps according to the profile in terraform-provider-rhcs/tests/ci/profile_handler.go
+#### Day1 cases only do creation step, we have to define a day1 case to verify the day1 works well
```
amalykhi, line 52: What does it mean "define a day1 case to verify the day1 works well"?
If it is done after Day1 (cluster creation) - how is it different from Day2 cases?

tests/README.md
xueli181114: @amalykhi As the doc tell `PrepareRHCSClusterByProfileENV` will load the cluster prepared in pre step. It won't create cluster again. You can put it to beforeeach, or we can make it a global variable.

tests/README.md
xueli181114: @amalykhi The day1 creation only does creation, it won't check if the configuration succeed. For example, when you create a cluster with admin enabled in day1, the cluster will create with admin enabled. But actually, we need to verify that the cluster is indeed created with admin enabled which means we have to login to the cluster with the admin username/password to make sure the configuration/creation worked. 
Why we mark it as day1-post rather than day2, because OCM-backend meets the problem now that some day2 cases may remove/change the day1 configuration no matter caused by not good coding or unrecoverable step and cases run randomly. I don't want TF provider automation meets same issue. So we label them differently then in future we can run them by order day1-post first, and then day2 and then maybe day3(TBD, those cases will cause destructive results)

tests/README.md
shellymiron: @amalykhi in that case we will need to change our automated test cases later on.
but let's keep them like that for now (as it is done [here](https://github.com/terraform-redhat/terraform-provider-rhcs/blob/2fb06bc27ee85dcde09425bf519cb15b1dde2d44/tests/e2e/init_test.go#L19))
and once this PR merged we will change ours as well on a new PR.

tests/ci/profile_handler.go
```diff
@@ -198,26 +258,35 @@ func GenerateClusterCreationArgsByProfile(profile *Profile) (clusterArgs *EXE.Cl
...
 	Expect(err).ToNot(HaveOccurred())
 	return &profile
 }
 
-func CreateRHCSClusterByProfile(profile *Profile, creationArgs *EXE.ClusterCreationArgs, manifests_dir string) (string, error) {
+func LoadProfileYamlFileByENV() *Profile {
+	profileEnv := os.Getenv("CLUSTER_PROFILE")
+	if profileEnv == "" {
+		panic(fmt.Errorf("ENV Variable RHCS_PROFILE_ENV is empty, please make sure you set the env value"))
+	}
```
shellymiron, line 280: actually - why are we using here `CLUSTER_PROFILE` instead of `RHCS_PROFILE_ENV`?

nirarg:
Please rebase on main branch

tests/ci/profile_handler.go
xueli181114: Removed RHCS_PROFILE_ENV
instead using CLUSTER_PROFILE to keep consistent between testing components.
Will updating in future if required. :)

shellymiron: ack :)

PR #328, opened 2023-10-04
add htpasswd test and expanding current IDP infrastructure- automating ocp63151

shellymiron:
Test case automated: [OCP-63151](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-63151)

passing result:
image

### DIFF
--- a/tests/e2e/idps_test.go
+++ b/tests/e2e/idps_test.go
@@ -1,0 +1,86 @@
+package e2e
+
+import (
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	conn "github.com/terraform-redhat/terraform-provider-rhcs/tests/ci"
+	"github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/cms"
+	con "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
+	exe "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/exec"
+)
+
+var _ = Describe("TF Test", func() {
+	Describe("Identity Providers test cases", func() {
+
+		// all identity providers - declared for future cases
+		type IDPServices struct {
+			htpasswd,
+			github,
+			gitlab,
+			google,
+			ldap,
+			openid exe.IDPService
+		}
+
+		var idpService IDPServices
+
+		Describe("Htpasswd IDP test cases", func() {
+			var htpasswdMap = []interface{}{map[string]string{}}
+			var htpasswdUsername, htpasswdPassword string
+
+			BeforeEach(func() {
+
+				htpasswdUsername = "jacko"
+				htpasswdPassword = "1q2wFe4rpoe2318"
+				htpasswdMap = []interface{}{map[string]string{"username": htpasswdUsername, "password": htpasswdPassword}}
+				idpService.htpasswd = *exe.NewIDPService(con.HtpasswdDir) // init new htpasswd service
+			})
+
+			AfterEach(func() {
+				err := idpService.htpasswd.Destroy()
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			Context("Author:smiron-High-OCP-63151 @OCP-63151 @smiron", func() {
+				It("OCP-63151 - Provision HTPASSWD IDP against cluster using TF", func() {
+					By("Create htpasswd idp for an existing cluster")
+
+					idpParam := &exe.IDPArgs{
+						Token:         token,
+						ClusterID:     clusterID,
+						Name:          "htpasswd-idp-test",
+						HtpasswdUsers: htpasswdMap,
+					}
+					err := idpService.htpasswd.Create(idpParam, "-auto-approve", "-no-color")
+					Expect(err).ToNot(HaveOccurred())
+					idpID, _ := idpService.htpasswd.Output()
+
+					By("List existing HtpasswdUsers and compare to the created one")
+					htpasswdUsersList, _ := cms.ListHtpasswdUsers(conn.RHCSConnection, clusterID, idpID.ID)
+					Expect(htpasswdUsersList.Status()).To(Equal(http.StatusOK))
+					respUserName, _ := htpasswdUsersList.Items().Slice()[0].GetUsername()
+					Expect(respUserName).To(Equal(htpasswdUsername))
+
+					By("Login with created htpasswd idp")
+					getResp, err := cms.RetrieveClusterDetail(conn.RHCSConnection, clusterID)
+					Expect(err).ToNot(HaveOccurred())
+					server := getResp.Body().API().URL()
+
+					ocAtter := &exe.OcAttributes{
+						Server:          server,
+						Username:        htpasswdUsername,
+						Password:        htpasswdPassword,
+						ClusterID:       clusterID,
+						AdditioanlFlags: nil,
+						Timeout:         5,
+					}
+					err = exe.OcLogin(*ocAtter)
+					Expect(err).ToNot(HaveOccurred())
+
+				})
+			})
+		})
+	})
+})
--- a/tests/utils/exec/idps.go
+++ b/tests/utils/exec/idps.go
@@ -1,0 +1,97 @@
+package exec
+
+import (
+	"context"
+	"fmt"
+
+	CON "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
+	h "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/helper"
+)
+
+type IDPArgs struct {
+	ClusterID     string        `json:"cluster_id,omitempty"`
+	Name          string        `json:"name,omitempty"`
+	ID            string        `json:"id,omitempty"`
+	Token         string        `json:"token,omitempty"`
+	OCMENV        string        `json:"ocm_environment,omitempty"`
+	URL           string        `json:"url,omitempty"`
+	MappingMethod string        `json:"mapping_method,omitempty"`
+	HtpasswdUsers []interface{} `json:"htpasswd_users,omitempty"`
+}
+
+type IDPService struct {
+	CreationArgs *IDPArgs
+	ManifestDir  string
+	Context      context.Context
+}
+
+// for now holds only ID, additional vars might be needed in the future
+type IDPOutput struct {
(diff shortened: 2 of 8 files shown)


shellymiron:
@amalykhi @tshefi @nirarg @bardielle @xueli181114 please review :)

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,77 @@
...
+		var (
+			htpasswdUsername = "jacko"
+			htpasswdPassword = "1q2wFe4rpoe2318"
+		)
+		htpasswdMap := []interface{}{map[string]string{"username": htpasswdUsername, "password": htpasswdPassword}}
+
+		BeforeEach(func() {
+			clusterID = os.Getenv("CLUSTER_ID")
+			// for future use
+			idpServices = map[string]*exe.IDPService{
```
sagidayan, line 28: nit, Why the use of a map? I would suggest creating a **struct** to hold these services. This way, I won't have to address a service with a string that might "change". Instead of 
`idpSevices["service-x"]` it would simply be `idpServices.ServiceX`.

It would be easier to maintain. Plus, we won't add to the map entries during runtime. I think...

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,77 @@
...
+				"openid":   exe.NewIDPService(con.OpenidDir),
+			}
+		})
+		AfterEach(func() {
+			err := idpServices["htpasswd"].Destroy()
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Context("Author:smiron-High-OCP-63151 @OCP-63151 @smiron", func() {
+			It("OCP-63151 - Provision HTPASSWD IDP against cluster using TF", func() {
```
sagidayan, line 43: Perhaps a context for each IDP would be simpler to grasp. I assume this file will grow, so containing only needed variables in a context can make lives easier down the road.

For example, you can have a BeforeEach in a context of HTPasswd that only creates an HTTPasswd IDP service, instead of creating all of them all the time.

sagidayan: Oh i see you are using `By` 👍🏽 . Also a good way to do it. Though the variable initializations point is still valid IMO

sagidayan (commented):
Had a few comments. But looks good 👏🏽

tests/e2e/idps_test.go
shellymiron: tnx! fixed :D

tests/e2e/idps_test.go
shellymiron: So I changed the structure a little, used `Describe` for each IDP inside the general Describe and 
initialized the vars in the `BeforeEach`. let me know what you think

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,82 @@
...
+			AfterEach(func() {
+				err := idpService.htpasswd.Destroy()
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			Context("Author:smiron-High-OCP-63151 @OCP-63151 @smiron", func() {
+				It("OCP-63151 - Provision HTPASSWD IDP against cluster using TF", func() {
+					By("Create htpasswd idp for an existing cluster")
+
+					idpService.htpasswd = *exe.NewIDPService(con.HtpasswdDir)
```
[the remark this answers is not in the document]
sagidayan: For this to work, you will need a wrapper for each test. Up to you

sagidayan (commented):
Much better! Great job!

tests/utils/exec/idps.go
```diff
@@ -0,0 +1,108 @@
...
+
+type IDPService struct {
+	CreationArgs *IDPArgs
+	ManifestDir  string
+	Context      context.Context
+}
+
+type IDPOutput struct {
+	ID      string `json:"idp_id,omitempty"`
+	Name    string `json:"name,omitempty"`
```
amalykhi, line 37: I see only id in ..htpasswd/output.tf

tests/e2e/idps_test.go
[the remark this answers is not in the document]
shellymiron: true, needs to be inside `BeforeEach`. fixed :D

tests/utils/exec/idps.go
```diff
@@ -0,0 +1,104 @@
...
+	ID            string            `json:"id,omitempty"`
+	Token         string            `json:"token,omitempty"`
+	OCMENV        string            `json:"ocm_environment,omitempty"`
+	URL           string            `json:"url,omitempty"`
+	MappingMethod string            `json:"mapping_method,omitempty"`
+	GitHub        map[string]string `json:"github,omitempty"`
+	GitLab        map[string]string `json:"gitlab,omitempty"`
+	Google        map[string]string `json:"google,omitempty"`
+	Htpasswd      map[string]string `json:"htpasswd,omitempty"`
+	HtpasswdUsers []interface{}     `json:"htpasswd_users,omitempty"`
```
amalykhi, line 23: I see this parameter covers also "Users"
htpasswdMap = []interface{}{map[string]string{"username": htpasswdUsername, "password": htpasswdPassword}}
and not actually used. Do you think it can be removed?

tests/utils/exec/idps.go
```diff
@@ -0,0 +1,104 @@
...
+
+type IDPArgs struct {
+	ClusterID     string            `json:"cluster_id,omitempty"`
+	Name          string            `json:"name,omitempty"`
+	ID            string            `json:"id,omitempty"`
+	Token         string            `json:"token,omitempty"`
+	OCMENV        string            `json:"ocm_environment,omitempty"`
+	URL           string            `json:"url,omitempty"`
+	MappingMethod string            `json:"mapping_method,omitempty"`
+	GitHub        map[string]string `json:"github,omitempty"`
```
amalykhi, line 19: thinking now if we need Github/gitlab/htpasswd/OpenID Args here if they are hardcoded in the manifest and for defining the Directory - we specify it in initializating..

shellymiron: I agree, in any case we can later on bring them back if we will need
removed them for now

tests/utils/exec/idps.go
shellymiron: removed 👍

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,82 @@
...
+					respUserName, _ := htpasswdUsersList.Items().Slice()[0].GetUsername()
+					Expect(respUserName).To(Equal(htpasswdUsername))
+
+					By("Login with created htpasswd idp")
+					getresp, err := cms.RetrieveClusterDetail(conn.RHCSConnection, clusterID)
+					Expect(err).ToNot(HaveOccurred())
+					server := getresp.Body().API().URL()
+
+					userLoginCMD := exe.GenerateOCLoginCMD(server, htpasswdUsername, htpasswdPassword, clusterID, "--insecure-skip-tls-verify")
+					output, err := exe.RetryCMDRun(userLoginCMD, 5)
```
nirarg, line 73: A suggestion:
Why not having one function "OcLogin" that wraps those two functions
It can be function in struct and the server provided in the constructor

Than in the future will add more actions to this struct, using the CMD commands

nirarg (commented):
minor comment/question
Other than that it looks good

tests/e2e/idps_test.go
shellymiron: hmm let me see if I understood you correctly - 
the suggestion is to create a structure for future CMD operations, and one of the attributes
of this structure will be the function OcLogin that will combine those 2 functions?

nirarg: Yes, something like this.
The idea is to start creating the infrastructure for oc actions
If you prefer to do this in different PR I'm also ok with this

shellymiron: ack, so Iv'e changed part of it based on your comment, 
and the struct itself will be implemented in a different PR :D

shellymiron:
@sagidayan @amalykhi @nirarg can you /lgtm & /approve if ok?

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/cms"
+	con "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
+	exe "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/exec"
+)
+
+var _ = Describe("TF Test", func() {
+	Describe("Identity Providers test cases", func() {
+
+		// all identity providers - declared for future cases
+		type IDPServices struct {
```
sagidayan, line 18: nit. Why do we need this? if we have a closure for each IDP type, we can have a variable in each one for the IDP under testing. But again, not a dealbreaker.

shellymiron: tnx! will be handled in the next PR for idps

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,88 @@
...
+					Expect(err).ToNot(HaveOccurred())
+					idpID, _ := idpService.htpasswd.Output()
+
+					By("List existing HtpasswdUsers and compare to the created one")
+					htpasswdUsersList, _ := cms.ListHtpasswdUsers(conn.RHCSConnection, clusterID, idpID.ID)
+					Expect(htpasswdUsersList.Status()).To(Equal(http.StatusOK))
+					respUserName, _ := htpasswdUsersList.Items().Slice()[0].GetUsername()
+					Expect(respUserName).To(Equal(htpasswdUsername))
+
+					By("Login with created htpasswd idp")
```
bardielle, line 66: If I'm not wrong we need to wait before we can log into the cluster @sagidayan ?

tests/e2e/idps_test.go
```diff
@@ -0,0 +1,88 @@
...
+					getresp, err := cms.RetrieveClusterDetail(conn.RHCSConnection, clusterID)
+					Expect(err).ToNot(HaveOccurred())
+					server := getresp.Body().API().URL()
+
+					// userLoginCMD := exe.GenerateOCLoginCMD(server, htpasswdUsername, htpasswdPassword, clusterID, "--insecure-skip-tls-verify")
```
bardielle, line 71: Why do you comment it out?

sagidayan: in general, don't include commented code. If you need it have it, if not - remove it

PR #330, opened 2023-10-04
OCM-4021: Provider Attributes changes

nirarg:
* Remove unused attributes
* Remove all attributes but Token from the docs - internal attributes
* Add option for environment variables for all string attributes
* Add "Authentication and configuration" section in the main index and remove attributes section

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -58,8 +58,6 @@ var _ tfprovider.Provider = &Provider{}
 type Config struct {
 	URL          types.String `tfsdk:"url"`
 	TokenURL     types.String `tfsdk:"token_url"`
-	User         types.String `tfsdk:"user"`
-	Password     types.String `tfsdk:"password"`
 	Token        types.String `tfsdk:"token"`
 	ClientID     types.String `tfsdk:"client_id"`
 	ClientSecret types.String `tfsdk:"client_secret"`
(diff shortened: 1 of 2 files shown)


EricPonvelle (commented):
/lgtm

provider/provider.go
```diff
@@ -130,6 +119,16 @@ func (p *Provider) Schema(ctx context.Context, req tfprovider.SchemaRequest, res
 	}
 }
 
+func (p *Provider) getValue(attr types.String, envSuffix string) (string, bool) {
```
sagidayan, line 122: Maybe change the name? it's a bit ambiguous.

provider/provider.go
```diff
@@ -81,30 +79,21 @@ func (p *Provider) Schema(ctx context.Context, req tfprovider.SchemaRequest, res
 	resp.Schema = tfpschema.Schema{
 		Attributes: map[string]tfpschema.Attribute{
 			"url": tfpschema.StringAttribute{
-				Description: "URL of the API server.",
+				Description: fmt.Sprintf("URL sets the base URL of the API gateway. The default is `%s`", sdk.DefaultURL),
 				Optional:    true,
```
bardielle, line 83: maybe you can set here a default value?

nirarg: The default value is determined in the SDK level, no need to add it here as well

provider/provider.go
nirarg: As discussed, I changed the function name to `getAttrValueOrConfig`

PR #336, opened 2023-10-08
Downgrade framework version because of an issue caused by the new version of the TF CLI 1.6.0

bardielle:
The new TF CLI 1.6.0 caused errors of resource/data source types Not Found.
According to [that open issue](https://github.com/hashicorp/terraform-plugin-framework/issues/853):
`This is a Framework bug affecting users of Terraform 1.6 using providers built using Plugin Framework 1.4.0. The root cause is the recent caching functionality being enabled in https://github.com/hashicorp/terraform-plugin-framework/issues/828.`
A workaround for now is downgrading the `terraform-plugin-framework` version (from 1.4.0 to 1.3.5) and also the `terraform-plugin-go` (from 0.19.0 to 0.18.0)

### DIFF
--- a/.github/workflows/check-pull-request.yaml
+++ b/.github/workflows/check-pull-request.yaml
@@ -50,7 +50,6 @@ jobs:
       uses: hashicorp/setup-terraform@v2
       with:
         terraform_wrapper: false
-        terraform_version: 1.5.7
 
     - name: Run the tests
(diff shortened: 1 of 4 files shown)


nirarg:
Great finding @bardielle 
/approve
@sagidayan can you please take a look?

Issue #341, opened 2023-10-11
LDAP Validation Issue


jland-redhat:
The LDAP schema validation for email seems to be incorrect.

So if I attempt to do something like:

```
  ldap = {
    url        = var.ldap_url
    attributes = {
      email                  = ["mail"]
      id                     = var.attribute_id
      name                   = var.attribute_name
      preferred_username     = var.attribute_preferred_username
    }
```

It will fail with "Invalid Email"

The backing code looks like this:
```
var ldapAttrSchema = map[string]schema.Attribute{
	"email": schema.ListAttribute{
		Description: "The list of attributes whose values should be used as the email address.",
		ElementType: types.StringType,
		Optional:    true,
		Validators: []validator.List{
			listvalidator.ValueStringsAre(
				stringvalidator.RegexMatches(common.EmailRegexp, "Invalid email"),
			),
		},
	},
```

The EmailRegexp basically says the value needs to look like an email (i.e. example@gmail.com). 

But that is generally not going to be the case because this is validating the LDAP Key, **not** the value of the email itself.

Should be a pretty simple fix I will try to create a quick PR

nirarg:
Hi @jland-redhat ,
Thank you for the contribution

I'm not so familiar with the ldap identity provider, but do you say that email attribute not contains actual email address?

@sagidayan can you please review this and the PR?

jland-redhat:
Yes I actually already created a PR with what I believe is the fix, hopefully explains it a little more

sagidayan:
Thanks for opening the issue, and much more for opening the PR. Let's move the discussion there.

nirarg:
@jland-redhat thank you for contributing this
I'm closing the issue
Feel free to re-open in case you think need further actions

PR #342, opened 2023-10-11
Updating email key validation

jland-redhat:
PR in relationship to this issue:
https://github.com/terraform-redhat/terraform-provider-rhcs/issues/341

Just removed the validation since I don't believe it is correct.

Removed the test that was "testing" the validation of the email key. 

And modified some of the other test values so they are closer to expected values.

### DIFF
--- a/provider/identityprovider/ldap.go
+++ b/provider/identityprovider/ldap.go
@@ -74,11 +74,7 @@ var ldapAttrSchema = map[string]schema.Attribute{
 		Description: "The list of attributes whose values should be used as the email address.",
 		ElementType: types.StringType,
 		Optional:    true,
-		Validators: []validator.List{
-			listvalidator.ValueStringsAre(
-				stringvalidator.RegexMatches(common.EmailRegexp, "Invalid email"),
-			),
-		},
+		Computed:    true,
 	},
 	"id": schema.ListAttribute{
 		Description: "The list of attributes whose values should be used as the user ID. (default ['dn'])",
--- a/provider/identityprovider/ldap.go
+++ b/provider/identityprovider/ldap.go
@@ -3,7 +3,6 @@ package identityprovider
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
--- a/provider/identityprovider/ldap.go
+++ b/provider/identityprovider/ldap.go
@@ -13,6 +13,7 @@ import (
 )
 
 var LDAPAttrDefaultID []string = []string{"dn"}
+var LDAPAttrDefaultEmail []string = []string{"mail"}
 var LDAPAttrDefaultName []string = []string{"cn"}
 var LDAPAttrDefaultPrefferedUsername []string = []string{"uid"}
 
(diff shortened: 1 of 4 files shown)


bardielle:
I saw that in rosa cli there is no validation on email, however, I think it's a important to keep it in the terraform provider, @ciaranRoche what do you think?

jland-redhat:
But the validation is not correct. You are validating the LDAP **key** not the actual value.

jland-redhat:
Also pushed a fix to the unused import

provider/identityprovider/ldap.go
```diff
@@ -74,11 +73,7 @@ var ldapAttrSchema = map[string]schema.Attribute{
 		Description: "The list of attributes whose values should be used as the email address.",
 		ElementType: types.StringType,
 		Optional:    true,
-		Validators: []validator.List{
-			listvalidator.ValueStringsAre(
-				stringvalidator.RegexMatches(common.EmailRegexp, "Invalid email"),
-			),
-		},
+		Computed:    true,
```
bardielle, line 76: I'm not sure I understood - why did you change it to be computed?

ciaranRoche:
> I saw that in rosa cli there is no validation on email, however, I think it's an important to keep it in the terraform provider, @ciaranRoche what do you think?

Feel like this would be better handled on the service side WDYT? 

I would like to remove any split brain issues here, if you folks want to keep it I would suggest adding the validation to https://github.com/openshift-online/ocm-common/tree/main/pkg/idp/validations so we can add it to CS and not maintain the logic in two places.

provider/identityprovider/ldap.go
jland-redhat: Honestly not 100% sure what computed does here, but I figured it would make sense to be consistent with each of the other LDAP fields which have computed set to true


For example:
```
	"id": schema.ListAttribute{
		Description: "The list of attributes whose values should be used as the user ID. (default ['dn'])",
		ElementType: types.StringType,
		Optional:    true,
		Computed:    true,
	},
	"name": schema.ListAttribute{
		Description: "The list of attributes whose values should be used as the display name. (default ['cn'])",
		ElementType: types.StringType,
		Optional:    true,
		Computed:    true,
	},
```

sagidayan: now I am wondering if they all should not be computed... @jland-redhat Did you test this manually? With an actual LDAP service?

sagidayan:
Overall, looks good. I want to keep it open to get to the bottom of the LDAP `computed` attrs.

/ok-to-test

sagidayan:
@JohnStrunk can we have your input on this PR and the `computed` attrs?

JohnStrunk:
> @JohnStrunk can we have your input on this PR and the `computed` attrs?

Looking through the idp code, it looks like e-mail should have `computed: true`. Having `computed` set means we can assign the state value based on what we calculate/receive from the OCM API. For example, in Read(), which is triggered from Import(), we set the e-mail based on what is returned. This is important so that we can pick up on changes that are made via the UI (and other non-TF methods).

bardielle:
@jland-redhat please also change the population to be like @JohnStrunk mentioned [here](https://github.com/terraform-redhat/terraform-provider-rhcs/blob/ced64a8c53b28e23ee7ad817970d3a61e31abd02/provider/identityprovider/identity_provider_resource.go#L488)
```
				email := attributes.Email()
				state.LDAP.Attributes.EMail, err = common.StringArrayToList(email)
				if err != nil {
					response.Diagnostics.AddError("failed to convert LDAP attribute EMail to tf list", err.Error())
				}

```

bardielle:
I meant that you can change it not to use the "get" function but set it with the response from the backend by that:
`email := attributes.Email()` instead of `email, ok := attributes.GetEmail()`

@jland-redhat I think it's fine to leave just like that since all the computed attributes were set the same with the `get` function. 

Please fix the subsystem tests 

/lgtm

PR #345, opened 2023-10-14
Enhance the frame for CI

xueli181114:
What I did in the PR:
* Move logger to a package to make it global. Please in your future coding, use logger rather than ```fmt.Println()``` to control the output
* Make Gateway, Region, Version, ChannelGroup to something can be overwritten in profile CI testing to meet future on-demand testing requirement
* Move clusterID initialize to each Describe. Not all cases have a cluster ID because some of them are day1 negative
* Fill clusterID = CI.PrepareRHCSClusterByProfileENV() to make it can ready ENV CLUSTER_ID or get output from output of terraform output
* Move oc command to a new package `openshift` , EXE should only contains the manifest files execution
* Remove garbage code


xueli181114:
image

xueli181114:
@amalykhi @shellymiron Please review the new updating to the frame. Recommend to create cluster by command 
```
export CLUSTER_PROFILE=rosa-sts-ad; ginkgo run --label-filter day1-prepare tests/e2e
```
So all your day2 cases will use the created cluster which can fit CI

tests/utils/constants/config.go
```diff
@@ -33,10 +31,10 @@ type RHCSconfig struct {
...
 	project := "terraform-provider-rhcs"
-	rootDir := GetEnvWithDefault(CON.WorkSpace, strings.SplitAfter(currentDir, project)[0])
+	rootDir := GetEnvWithDefault(WorkSpace, strings.SplitAfter(currentDir, project)[0])
 
 	// defaulted to staging
-	RHCS.RHCSEnv = GetEnvWithDefault(CON.OCMEnv, RHCS.RHCSEnv)
+	RHCS.RHCSEnv = GetEnvWithDefault(RHCSENV, RHCS.RHCSEnv)
 
 	if os.Getenv("CLUSTER_PROFILE") != "" {
 		RHCS.ClusterProfile = os.Getenv("CLUSTER_PROFILE")
```
shellymiron, line 40: why not using the function `GetEnvWithDefault`?
it set default value in case of null vars

tests/ci/profile_handler.go
```diff
@@ -97,11 +97,10 @@ func PrepareAccountRoles(token string, accountRolePrefix string, awsRegion strin
 	}
 	args := &EXE.AccountRolesArgs{
 		AccountRolePrefix: accountRolePrefix,
-		// OCMENV            string `json:"ocm_environment,omitempty"`
```
amalykhi, line 100: I think we need ocm_environment.
IIUC it is used for applying correct trust policies. Right, @nirarg ?

tests/e2e/idps_test.go
```diff
@@ -5,13 +5,19 @@ import (
...
 	conn "github.com/terraform-redhat/terraform-provider-rhcs/tests/ci"
 	"github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/cms"
 	con "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
 	exe "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/exec"
+	"github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/openshift"
 )
 
 var _ = Describe("TF Test", func() {
+	var clusterID string
+	BeforeEach(func() {
```
amalykhi, line 18: Why do we need to get cluster ID before every test case? why not to get in in the "Before suit" - in init_test.go?

xueli181114: Because not all cases need a cluster ID. Especially some day1 negative cases, there is no cluster created at all.

tests/ci/profile_handler.go
xueli181114: Really? I remember it was removed? Let me double check.

tests/utils/constants/config.go
xueli181114: os.Getenv("CLUSTER_PROFILE") will give an empty string if no ENV set. So no need to set a default value.

xueli181114: CLUSTER_PROFILE is a must set ENV, we should raise error rather that set a default one.

tests/e2e/idps_test.go
amalykhi: But most of the cases will be Day2.  
If I understand the flow - the cluster will be created anyway by profile in the flow, but not every test case will use it. Looks for me overwhelming to set it for every Before Each, instead of - once per test suit

shellymiron: @xueli181114 I believe there won't be any error if there won't be any use for it for some cases,
so it can still be in the `init_test.go` IMO

tests/ci/profile_handler.go
nirarg: The `ocm_environment` wasn't removed, there was just added default value `production` in the module, so user doesn't have to indicate this.
We added new data source `rhcs_info` that will replace this requirement, its not planed to be changed in this module, the new WIP module is going to include this

xueli181114: Yes,  I checked the code. It is indeed required. Will fix it.

tests/e2e/idps_test.go
xueli181114: OK. I checked terraform output won't return error when there is no resources creation. Moving it back to init_test.go

tests/ci/profile_handler.go
xueli181114: Fixed this part. I can create a cluster on production now with env set.
To make all resources using same env, I set it in exec package, rather than profile handler.

tests/e2e/idps_test.go
xueli181114: Updated. It had been moved back to init_test.go

tests/README.md
```diff
@@ -121,9 +121,34 @@ This allows run by case filter to simulate CI. Anybody can customize the case la
   * Export the token
     * `export RHCS_TOKEN=<rhcs token>` #If it had been run, then you can skip
   * Export the label filter to locate the preparation case
-    * `export LabelFilter="destroy"` #This filter only choose all critical and high priority cases with day1-post and day2
+    * `export LabelFilter="destroy"` #This filter only choose the destroy case for resources destroy
   * Run make command
     * `make e2e_test`
+### To run on local simulate CI without make file
+This allows run by case filter to simulate CI. Anybody can customize the case label filter to filter out the cases want to run
```
bardielle, line 128: I didn't understand it. 
@EricPonvelle can you rephrase it or help @xueli181114 with it?

bardielle: Maybe something like that: 
```suggestion
### Running a local CI simulation
This feature allows for running tests through a case filter to simulate CI. Anyone can customize the case label filter to select the specific cases that would be run. 
```

tests/README.md
```diff
@@ -121,9 +121,34 @@ This allows run by case filter to simulate CI. Anybody can customize the case la
   * Export the token
     * `export RHCS_TOKEN=<rhcs token>` #If it had been run, then you can skip
   * Export the label filter to locate the preparation case
-    * `export LabelFilter="destroy"` #This filter only choose all critical and high priority cases with day1-post and day2
+    * `export LabelFilter="destroy"` #This filter only choose the destroy case for resources destroy
   * Run make command
     * `make e2e_test`
+### To run on local simulate CI without make file
```
amalykhi, line 127: Are the flows with  `make e2e_test` command still relevant?

Also it fails on • [PANICKED] [50.411 seconds]
TF Test Create cluster test [It] CreateClusterByProfile [day1-prepare]
/Users/amalykhi/repos/terraform-provider-rhcs/tests/e2e/cluster_creation.go:13

  [PANICKED] Test Panicked
  In [It] at: /opt/homebrew/Cellar/go/1.20.4/libexec/src/runtime/iface.go:262 @ 10/18/23 15:36:56.799

  interface conversion: interface {} is nil, not []interface {}

xueli181114: Good to know you fixed it. Resolve it.

tests/utils/log/logger.go
```diff
@@ -0,0 +1,15 @@
+package log
+
+import (
+	logging "github.com/sirupsen/logrus"
+)
+
+func GetLogger() *logging.Logger {
```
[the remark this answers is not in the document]
xueli181114: It's future's work, not high priority.

tests/ci/profile_handler.go
```diff
@@ -215,25 +213,35 @@ func GenerateClusterCreationArgsByProfile(token string, profile *Profile) (clust
...
-		if vpcOutput.ClusterPrivateSubnets == nil {
-			err = fmt.Errorf("error when creating the vpc, check the previous log. The created resources had been destroyed")
-			return
-		}
-		if profile.PrivateLink {
-			clusterArgs.AWSSubnetIDs = vpcOutput.ClusterPrivateSubnets
+		// Supports ENV set passed to make cluster provision more flexy in prow
+		// Export the subnetIDs via env variable if you have existing ones export SubnetIDs=<subnet1>,<subnet2>,<subnet3>
+		// Export the availability zones via env variable export AvailabilitiZones=<az1>,<az2>,<az3>
+		if os.Getenv("SubnetIDs") != "" && os.Getenv("AvailabilitiZones") != "" {
```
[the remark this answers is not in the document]
xueli181114: @yasun1 No, there is one already to overwrite the profile by ENV those are highlevel. This part is only exposed for shared vpc in case anybody wants to pass the value from other steps.

tests/ci/profile_handler.go
```diff
@@ -13,6 +13,7 @@ import (
 	CON "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
 	EXE "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/exec"
 	HELPER "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/helper"
+	. "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/log"
```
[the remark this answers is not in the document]
xueli181114: I prefer not to have so many log.Logger.xxx(xxx). There won't be too many functions defined in the package.

PR #349, opened 2023-10-18
Add post day 1 verification tests

shellymiron:
**PR changes:**

- 6 new verification tests added for testing post cluster creation 
- `FIPS`, `Etcd`, `Autoscale`, `ComputeMachineType` attributes were added 
- Getter/Setter for `clusterID` and `token` inside the test suite
- Test case coverage:` OCP-63140, OCP-63133, OCP-63143, OCP-64023, OCP-63141, OCP-68423`

### DIFF
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -17,8 +17,7 @@ var clusterID string
 
 func TestRHCSProvider(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "RHCS Provider Test")
-
+	RunSpecs(t, "e2e tests suite")
 }
 
 var _ = BeforeSuite(func() {
--- a/tests/utils/constants/config.go
+++ b/tests/utils/constants/config.go
@@ -26,14 +26,15 @@ type RHCSconfig struct {
 	RHCSEnv           string `env:"RHCS_ENV" default:"staging" yaml:"env"`
 	ClusterProfile    string `env:"CLUSTER_PROFILE" yaml:"clusterProfile,omitempty"`
 	ClusterProfileDir string `env:"CLUSTER_PROFILE_DIR" yaml:"clusterProfileDir,omitempty"`
-
-	YAMLProfilesDir string
+	RhcsOutputDir     string
+	YAMLProfilesDir   string
+	RootDir           string
 }
 
 func init() {
 	currentDir, _ := os.Getwd()
 	project := "terraform-provider-rhcs"
-	rootDir := GetEnvWithDefault(WorkSpace, strings.SplitAfter(currentDir, project)[0])
+	RHCS.RootDir = GetEnvWithDefault(WorkSpace, strings.SplitAfter(currentDir, project)[0])
 
 	// defaulted to staging
 	RHCS.RHCSEnv = GetEnvWithDefault(RHCSENV, RHCS.RHCSEnv)
@@ -45,7 +46,8 @@ func init() {
 		RHCS.ClusterProfileDir = os.Getenv("CLUSTER_PROFILE_DIR")
 	}
 
-	RHCS.YAMLProfilesDir = path.Join(rootDir, "tests", "ci", "profiles")
+	RHCS.RhcsOutputDir = GetRhcsOutputDir()
+	RHCS.YAMLProfilesDir = path.Join(RHCS.RootDir, "tests", "ci", "profiles")
 }
 
 // gatewayURL is used to get the global env of default gateway for testing
@@ -91,4 +93,16 @@ func GetEnvWithDefault(key string, defaultValue string) string {
 		}
 	}
 	return defaultValue
+}
+
+func GetRhcsOutputDir() string {
+	var rhcsNewOutPath string
+
+	if GetEnvWithDefault("RHCS_OUTPUT", "") != "" {
+		rhcsNewOutPath = os.Getenv("RHCS_OUTPUT")
+		return rhcsNewOutPath
+	}
+	rhcsNewOutPath = path.Join(RHCS.RootDir, "tests", "rhcs_output")
+	os.MkdirAll(rhcsNewOutPath, 0777)
+	return rhcsNewOutPath
 }
--- a/tests/utils/constants/constants.go
+++ b/tests/utils/constants/constants.go
@@ -34,6 +34,9 @@ var (
 	RHCSPrefix          = "rhcs"
 	TFYAMLProfile       = "tf_cluster_profile.yml"
 	LdapURL             = "ldap://ldap.forumsys.com/dc=example,dc=com?uid"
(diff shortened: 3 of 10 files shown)


tests/e2e/machine_pool_test.go
```diff
@@ -25,8 +25,8 @@ var _ = Describe("TF Test", func() {
 		It("MachinePoolExampleNegative", func() {
 
 			MachinePoolArgs := &exe.MachinePoolArgs{
-				Token:       token,
-				Cluster:     clusterID,
+				Token:       GetClusterID(),
```
sagidayan, line 28: I think this is wrong... token <-> cluster_id

shellymiron:
@xueli181114 @yasun1 Hey! can you please help to review? :)

tests/e2e/e2e_suite_test.go
```diff
@@ -15,11 +15,18 @@ var ctx context.Context
 var token string
 var clusterID string
 
+func GetToken() string {
+	return token
+}
+
+func GetClusterID() string {
```
amalykhi, line 22: Could you advise what is better - to use getter here? Or just call the function once (as in line 22/28) and just use it inside the tests?
@nirarg @bardielle @sagidayan

tests/e2e/cluster_destroy_test.go
```diff
@@ -13,7 +13,7 @@ var _ = Describe("TF Test", func() {
 
 				// Generate/build cluster by profile selected
 				profile := CI.LoadProfileYamlFileByENV()
-				err := CI.DestroyRHCSClusterByProfile(token, profile)
+				err := CI.DestroyRHCSClusterByProfile(GetToken(), profile)
```
xueli181114, line 16: may I know why we rename it to _test?

tests/e2e/e2e_suite_test.go
```diff
@@ -15,11 +15,18 @@ var ctx context.Context
 var token string
 var clusterID string
 
+func GetToken() string {
```
xueli181114, line 18: - Why it is renamed to _test?

tests/e2e/e2e_suite_test.go
xueli181114: What's the purpose for the function?

tests/e2e/e2e_suite_test.go
```diff
@@ -15,11 +15,18 @@ var ctx context.Context
...
+}
+
 func TestRHCSProvider(t *testing.T) {
-	RegisterFailHandler(Fail)
 	token = os.Getenv(CON.TokenENVName)
 	clusterID = CI.PrepareRHCSClusterByProfileENV()
 	ctx = context.Background()
-	RunSpecs(t, "RHCS Provider Test")
-
+	RegisterFailHandler(Fail)
```
xueli181114, line 30: Please RegisterFailHandler at the beginning , otherwise unexpected error cannot be caught

tests/utils/constants/config.go
```diff
@@ -92,3 +94,16 @@ func GetEnvWithDefault(key string, defaultValue string) string {
 	}
 	return defaultValue
 }
+
+func GetRhcsOutputDir() string {
+	var rhcsNewOutPath string
+
+	if GetEnvWithDefault("RHCS_OUTPUT", "") != "" {
+		rhcsNewOutPath = os.Getenv("RHCS_OUTPUT")
+		os.MkdirAll(rhcsNewOutPath, 0777)
```
xueli181114, line 103: If we have this env RHCS_OUTPUT set means there is a dir, do not need to create again.

tests/utils/constants/constants.go
```diff
@@ -33,6 +33,9 @@ var (
 	WorkSpace           = "WORKSPACE"
 	RHCSPrefix          = "rhcs"
 	TFYAMLProfile       = "tf_cluster_profile.yml"
+	DefaultMPLabels     = map[string]string{
```
xueli181114, line 36: Let's make it into a config file like backend.

tests/utils/helper/helper.go
```diff
@@ -25,6 +26,13 @@ func Parse(data []byte) map[string]interface{} {
 	return object
 }
 
+func GetJsonFromPath(path string, filename string) map[string]interface{} {
+	combinedFilePath := path + "/" + filename
```
xueli181114, line 30: how about use `path.Join` in case somebody running code on Windows

tests/ci/rhcs_output/data.json
```diff
@@ -0,0 +1,4 @@
+{
+    "HtpasswdPass" : "1q2wFe4rpoe2318",
```
xueli181114, line 2: This file is not supposed to be uploaded.

tests/e2e/e2e_suite_test.go
shellymiron: just getting the clusterID
same is dont for the token.
I think it is more nit that way, but I don't mind changing

tests/e2e/cluster_destroy_test.go
shellymiron: because golang does not understand the file without the `test` postfix, it is still working without it,
but it gives warnings.

tests/e2e/e2e_suite_test.go
shellymiron: because it is a test suite - it replaces the init_test file

tests/utils/helper/helper.go
shellymiron: path.join giving an error on the `helper.go` file, it does not want to import the path package

tests/ci/rhcs_output/data.json
shellymiron: ok, that's what I wanted to know before :) in that case I'll implement @amalykhi solution to generate random
chars for the password. we encountering github massage when pushing to the PR that this password is exposed 
and there is a leakage.

tests/utils/constants/constants.go
shellymiron: can you send a reference to that in the backend repo? could not find how it was implemented

tests/utils/helper/helper.go
xueli181114: It's not related to the import, you named the parameter as `path`, thus once you imported another path, you will get an error

tests/utils/constants/constants.go
xueli181114: Let's keep it for current PR. I will send you via slack about configuration in backend.

shellymiron: ack, tnx!

tests/utils/helper/helper.go
shellymiron: ahh I see
ok, will change that

tests/e2e/e2e_suite_test.go
shellymiron: changed back to be without the getter

PR #356, opened 2023-10-22
Removing an example of htpasswd_idp password due to an alert from the…

bardielle:
… scanner


docs/resources/identity_provider.md
```diff
@@ -20,7 +20,7 @@ resource "rhcs_identity_provider" "htpasswd_idp" {
   htpasswd = {
     users = [{
       username = "user1"
```
nirarg, line 22:
```suggestion
      username = "<username>"
```

sagidayan:
did not understand the issue... plus no ticket attached

That being said, don't see the issue with this PR. 
/approve

PR #362, opened 2023-10-26
Add a PlanModifier to indicate an attribute is immutable.

JohnStrunk:
This replaces the old ValueCannotBeChanged modifier that existed before the framework upgrade. This modifier will only allow an attribute's value to be set during the initial creation of the containing resource. After that, any attempt to change the attribute will result in an error.

This modifier should be viewed as an alternative to the RequiresReplace modifier where we would rather return an error to the user than cause the resource to be recreated.

The modifier cannot be used on a "computed" attribute because computed attributes may get assigned values that are not present in the configuration.

It also does not permit "unknown" values (after the initial create) since they cannot be validated.

### DIFF
--- a/provider/common/immutable_modifier.go
+++ b/provider/common/immutable_modifier.go
@@ -1,0 +1,203 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


JohnStrunk:
/cc @bardielle
If this looks good, we can add it to the attributes that are only used during resource create.

Thinking about this, we could also (or instead of this) create a planmodifier that issues a warning when an attribute is changed. We could say something like "Warning: Attribute 'foo' has changed. This change will be ignored because it is only used during the initial creation of the resource." This way, we would send a message so that the user knows their change won't take effect, but we won't actually generate an error and halt the "apply." It seems a bit silly to refuse to apply (and require them to revert the config) if it's not going to have an effect.

sagidayan:
Thanks @JohnStrunk 
/approve

bardielle:
/lgtm 
great work!

PR #364, opened 2023-10-28
OCM-4535,OCM-4537,OCM-4653 | feat: additional SGs day1/2

gdbranco:
Related issues: 
* [OCM-4535](https://issues.redhat.com//browse/OCM-4535)
* [OCM-4537](https://issues.redhat.com//browse/OCM-4537)
* [OCM-4653](https://issues.redhat.com//browse/OCM-4653)

# What
Provides option to supply additional compute, infra and control plane security groups for the default worker, infra and control planes machine pool (day1) as well as for future worker machine pools (day2)


provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -269,6 +269,21 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 					listplanmodifier.RequiresReplace(),
 				},
 			},
+			"aws_additional_compute_security_group_ids": schema.ListAttribute{
+				Description: "AWS additional compute security group ids.",
+				ElementType: types.StringType,
+				Optional:    true,
```
bardielle, line 275: What about updating those attributes? we would like to support it? if no please add a plenmodifier 
If we would like to support updating it please add the logic in update function

gdbranco: There's no current functionality to update the attributes. It will be supported at some point in the future

bardielle: So you should block it with plan modifier 
```
PlanModifiers: []planmodifier.String{
					common.Immutable(),
```

bardielle:
/lgtm 
@nirarg please review it as well

provider/machinepool/machine_pool_resource.go
```diff
@@ -846,7 +883,7 @@ func (r *MachinePoolResource) ImportState(ctx context.Context, req resource.Impo
 }
 
 // populateState copies the data from the API object to the Terraform state.
-func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *MachinePoolState) {
+func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *MachinePoolState) error {
```
nirarg, line 886: Because you return here error, please check `err != nil` for all callers

provider/machinepool/machine_pool_resource.go
```diff
@@ -856,7 +890,21 @@ func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *Mac
...
+			additionalSecurityGroupsList, err := common.StringArrayToList(additionalSecurityGroups)
+			if err != nil {
+				return err
+			}
+			state.AdditionalSecurityGroupIds = additionalSecurityGroupsList
+		} else {
+			state.AdditionalSecurityGroupIds = types.ListNull(types.StringType)
+		}
+	} else {
+		state.AdditionalSecurityGroupIds = types.ListNull(types.StringType)
```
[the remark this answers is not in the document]
nirarg: Did you have further discussion about this comment?
If I'm not wrong, setting Null value should be done only for `computed` attribute
Also in the cluster resource you didn't use the else part

provider/cluster/cluster_resource.go
```diff
@@ -646,6 +692,33 @@ func populateClusterState(object *cmv1.Cluster, state *ClusterState) error {
...
 	}
 
+	additionalComputeSecurityGroupIds, ok := object.AWS().GetAdditionalComputeSecurityGroupIds()
+	if ok {
+		awsAdditionalSecurityGroupIds, err := common.StringArrayToList(additionalComputeSecurityGroupIds)
+		if err != nil {
+			return err
+		}
+		state.AWSAdditionalComputeSecurityGroupIds = awsAdditionalSecurityGroupIds
+	}
```
nirarg, line 702: nit: I think you can use one liner for the if in all places like this

```suggestion
	if additionalComputeSecurityGroupIds, ok := object.AWS().GetAdditionalComputeSecurityGroupIds(); ok {
		awsAdditionalSecurityGroupIds, err := common.StringArrayToList(additionalComputeSecurityGroupIds)
		if err != nil {
			return err
		}
		state.AWSAdditionalComputeSecurityGroupIds = awsAdditionalSecurityGroupIds
	}
```

gdbranco:
@nirarg 
> Also in the cluster resource you didn't use the else part

I didn't had any issues on the cluster side, but for the machine pool I kept getting errors if I didn't add this. I noticed other attributes in the cluster do not have the else either so I expect it would be somewhat standard. But I can check this to make sure

provider/cluster/cluster_resource.go
gdbranco: So, I do agree with this, but all code around the changes are also not using the one liner version. So I'm thinking of keeping this for consistency and we can update all at once as a refactor. What do you think?

provider/machinepool/machine_pool_resource.go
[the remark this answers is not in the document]
gdbranco: If I don't set it I keep getting a conversion of value error. Any tips to adjusting this?

nirarg:
> I didn't had any issues on the cluster side, but for the machine pool I kept getting errors if I didn't add this. I noticed other attributes in the cluster do not have the else either so I expect it would be somewhat standard. But I can check this to make sure

@gdbranco 
I tested it locally:
I removed this else block and didn't assign any value to this variable in my manifest. I didn't get the error.
Can you please elaborate how you got the error?
We can have a call for this, if needed

PR #365, opened 2023-10-29
adding gitlab/hub idp tests + fixes

shellymiron:
Adding gitlab/hub idp tests and set prod as default Gateway.

### DIFF
--- a/tests/e2e/idps_test.go
+++ b/tests/e2e/idps_test.go
@@ -144,5 +144,85 @@ var _ = Describe("TF Test", func() {
 					})
 			})
 		})
+		Describe("GitLab IDP test cases", func() {
+
+			var gitlabIDPClientSecret, gitlabIDPClientId string
+
+			BeforeEach(func() {
+				gitlabIDPClientId = h.RandStringWithUpper(20)
+				gitlabIDPClientSecret = h.RandStringWithUpper(30)
+				idpService.gitlab = *exe.NewIDPService(con.GitlabDir) // init new gitlab service
+			})
+
+			AfterEach(func() {
+				err := idpService.gitlab.Destroy()
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			Context("Author:smiron-High-OCP-64028 @OCP-64028 @smiron", func() {
+				It("OCP-64028 - Provision GitLab IDP against cluster using TF", ci.Day2, ci.High, ci.FeatureIDP, func() {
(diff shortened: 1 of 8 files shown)


tests/tf-manifests/rhcs/idps/openid/variables.tf
```diff
@@ -4,7 +4,7 @@ variable "token" {
 }
 variable "gateway" {
   type    = string
-  default = "https://api.stage.openshift.com"
+  default = "https://api.openshift.com"
```
xueli181114, line 7: Please don't set the default gateway to production. Keep consistent with cluster's manifest. There is a global env controlling the gateway urls.

shellymiron: the thing is, the `RHCS_ENV` or the `GATEWAY_URL` environment variable, does not influence the url of:
```
provider "rhcs" {
  token = var.token
  url   = var.url <-----------
}
```
it is always keep staying on "staging" because we have "url" attribute also for the idp themselves, and it overrides it in the command line vars we sent to `terraform apply`, for example:

```
resource "rhcs_identity_provider" "gitlab_idp" {
  cluster        = var.cluster_id
  mapping_method = var.mapping_method
  name           = "Gitlab"
  gitlab = {
    client_id     = var.client_id
    client_secret = var.client_secret
    url           = var.url
  }
}
```

shellymiron: I will change it back, but I just hope it will not fail in CI and that this is just something that happens when running locally

shellymiron:
image

tests/e2e/idps_test.go
```diff
@@ -144,5 +144,85 @@ var _ = Describe("TF Test", func() {
...
+			})
+
+			Context("Author:smiron-High-OCP-64027 @OCP-64027 @smiron", func() {
+				It("OCP-64027 - Provision GitHub IDP against cluster using TF", ci.Day2, ci.High, ci.FeatureIDP, func() {
+					By("Create GitHub idp for an existing cluster")
+
+					idpParam := &exe.IDPArgs{
+						Token:         token,
+						ClusterID:     clusterID,
+						Name:          "github-idp-test",
```
amalykhi, line 209: I would add TC id to the name for easier debug

shellymiron: ack, i'll open a PR for that

PR #370, opened 2023-11-01
Enhance the oc login command to kill no kubeconfig warning in prow

xueli181114:
When there is an ENV variable `KUBECONFIG`set and then try to login with `oc` command at the same time file path of `KUBECONFIG` not existing, it will cause warning of config not found. 
This PR enhanced the oc login command to make it redirect the kubeconfig path to an indicated file rather than using global `KUBECONFIG` env.
There will be a dir `$RHCS_OUTPUT/kubeconfig` created when run the tests, and it will be destroyed in destroy step(This is for clean local env. In prow, folders will be skipped when update $SHARE_DIR secret dir.)

### DIFF
--- a/tests/e2e/cluster_destroy_test.go
+++ b/tests/e2e/cluster_destroy_test.go
@@ -1,19 +1,27 @@
 package e2e
 
 import (
+	"os"
+
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


xueli181114:
@shellymiron @amalykhi please review

xueli181114:
image

PR #372, opened 2023-11-02
 OCM-214: Feature/cluster autoscaler resource

sagidayan:
Ported the great work of @osherdp (💪🏽) to the new TF framework.

Cleaned the code a bit, mainly in the schema. 
Validations are now separate and a single Range attribute is reused

this PR should replace #240


provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+
+func (r *ClusterAutoscalerResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Cluster-wide autoscaling configuration.",
+		Attributes: map[string]schema.Attribute{
+			"cluster": schema.StringAttribute{
+				Description: "Identifier of the cluster.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					common.Immutable(),
```
bardielle, line 64: :+1:

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+		response.Diagnostics.AddError(
+			"Failed waiting for cluster to get ready",
+			fmt.Sprintf(
+				"Cannot poll state of cluster with identifier '%s': %v",
+				plan.Cluster.ValueString(), err,
+			),
+		)
+		return
+	}
+
```
bardielle, line 233:
```suggestion
// Wait till the cluster is ready:
	err := common.WaitTillClusterReady(ctx, r.collection, state.Cluster.ValueString())
	if err != nil {
		resp.Diagnostics.AddError(
			"Cannot poll cluster state",
			fmt.Sprintf(
				"Cannot poll state of cluster with identifier '%s': %v",
				state.Cluster.ValueString(), err,
			),
		)
		return
	}
	
	resource := r.collection.Cluster(plan.Cluster.ValueString())
```

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+	object := update.Body()
+	state := &ClusterAutoscalerState{}
+	populateAutoscalerState(object, plan.Cluster.ValueString(), state)
+
+	diags = response.State.Set(ctx, state)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
```
bardielle, line 355: why do we need that if statement?

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+	}
+
+	response.State.RemoveResource(ctx)
+}
+
+func (r *ClusterAutoscalerResource) ImportState(ctx context.Context, request resource.ImportStateRequest,
+	response *resource.ImportStateResponse) {
+	clusterId := request.ID
+
+	getResponse, err := r.collection.Cluster(clusterId).Autoscaler().Get().SendContext(ctx)
```
bardielle, line 389: why do we have a special code in import function - the import is with clusterId (only) so we can use the pasthrough

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+
+func (r *ClusterAutoscalerResource) Create(ctx context.Context, request resource.CreateRequest, response *resource.CreateResponse) {
+	plan := &ClusterAutoscalerState{}
+	diags := request.Plan.Get(ctx, plan)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	resource := r.collection.Cluster(plan.Cluster.ValueString())
```
bardielle, line 211: we should also validate that there isn't an existing autoscaler for that cluster

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,640 @@
...
+		return
+	}
+
+	state := &ClusterAutoscalerState{}
+	populateAutoscalerState(getResponse.Body(), clusterId, state)
+	diags := response.State.Set(ctx, state)
+	response.Diagnostics.Append(diags...)
+}
+
+func (r *ClusterAutoscalerResource) waitForClusterToGetReady(ctx context.Context, clusterId string) error {
```
bardielle, line 404: no need it anymore...

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: if we already have errors we should exit.

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: Im not sure about this... We should check before assuming.

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: did not understand...

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: This is all @JohnStrunk 's work :)
Thanks

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: Thanks for the tip. Will update.

sagidayan:
@bardielle This PR needs one more label to be merged. FYI

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,605 @@
...
+var _ resource.Resource = &ClusterAutoscalerResource{}
+var _ resource.ResourceWithImportState = &ClusterAutoscalerResource{}
+var _ resource.ResourceWithConfigure = &ClusterAutoscalerResource{}
+
+func (r *ClusterAutoscalerResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_cluster_autoscaler"
+}
+
+func (r *ClusterAutoscalerResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
```
nirarg, line 58: All attributes but "cluster" are optional, not computed and without default value. What happens when user creates it just with the cluster ID? shouldn't we save the default values in the state?

provider/clusterautoscaler/cluster_autoscaler_resource.go
```diff
@@ -0,0 +1,605 @@
...
+			fmt.Sprintf(
+				"Cannot poll state of cluster with identifier '%s': %v",
+				plan.Cluster.ValueString(), err,
+			),
+		)
+		return
+	}
+
+	resource := r.collection.Cluster(plan.Cluster.ValueString())
+
```
bardielle, line 226: I see in rosa cli there is a validation if the autoscaler is already exit: 
```
		r.Reporter.Errorf("Autoscaler for cluster '%s' already exists. "+
			"You should edit it via 'rosa edit autoscaler'", clusterKey)
		os.Exit(1)
```

Please add such validation here

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: If they are not computed, we should not assign values. Once a user changes a value - it will update the backend - and store the value in the state.

provider/clusterautoscaler/cluster_autoscaler_resource.go
sagidayan: Done

provider/clusterautoscaler/cluster_autoscaler_resource.go
nirarg: Yes, I understand that, but I guess there are default values, and now I understand that the API doesn't return the default values.
So I guess its fine

nirarg:
@osherdp and @sagidayan
Thanks a lot for this effort, looks great

PR #376, opened 2023-11-07
Add new data source for cluster rosa

bardielle:
This PR adds a new data source for rosa clusters
It's based on the rosa cluster state and uses the same state struct - that's why I had to mark some of the attributes as deprecated. 
I added a new attribute in the STS nested attribute to display the role-arns 
Also, I split the populate function in rosa cluster resource into 2 functions - one for non-optional attributes and the other for optional(that set null for attributes with no values)


nirarg:
@bardielle can this PR be closed?

bardielle:
no need that PR anymore

PR #388, opened 2023-11-15
Fix OCP version for account roles and cluster creation

amalykhi:
from log
time="2023-11-15T10:28:29+02:00" level=info msg="Cluster OCP latest version is set to 4.14.2"


xueli181114:
@amalykhi Can you please paste the log of the cluster provision success?

PR #392, opened 2023-11-19
add idp negative tests - validate mandatory fields

shellymiron:
This PR will validate all the mandatory fields each idp resource has, as described in the[ terraform rhcs provider doc](https://registry.terraform.io/providers/terraform-redhat/rhcs/latest/docs/resources/identity_provider)

image

### DIFF
--- a/tests/e2e/idps_test.go
+++ b/tests/e2e/idps_test.go
@@ -7,7 +7,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	CI "github.com/terraform-redhat/terraform-provider-rhcs/tests/ci"
 	ci "github.com/terraform-redhat/terraform-provider-rhcs/tests/ci"
 	"github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/cms"
 	con "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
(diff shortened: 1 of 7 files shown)


cmyster:
LGTM

tests/ci/labels.go
```diff
@@ -7,6 +7,7 @@ import (
 // Features
 var FeatureMachinepool = Label("feature-machinepool")
 var FeatureIDP = Label("feature-idp")
+var FeatureIDPNegative = Label("feature-idp-negative")
```
amalykhi, line 10: I don't think we need additional label for this

tests/e2e/idps_test.go
```diff
@@ -29,240 +29,473 @@ var _ = Describe("TF Test", func() {
...
+				})
 			})
+			Context("LDAP IDP test cases", func() {
+				BeforeEach(func() {
 
-			AfterEach(func() {
-				err := idpService.htpasswd.Destroy()
-				Expect(err).ToNot(HaveOccurred())
+					userName = "newton"
+					password = "password"
```
amalykhi, line 102: I wouldn't use a hardcoded password even if it's a negative test case.

shellymiron: this specific password must be hardcoded, this is the password for ldap test cases when we need to use 
`"ldap://ldap.forumsys.com/dc=example,dc=com?uid"`

tests/ci/labels.go
shellymiron: ok, so maybe later on we will add a general "negative-tc" label? so we could run all the negative tests at once?
or should each feature run all it's cases without negative/positive consideration?

tests/e2e/idps_test.go
cmyster: (just a suggestion) On the other hand, we can set it as a setting and then use the same code for positive ("newton") and negative test ("notnewton") or something trickier such as a username injection attack.

shellymiron: sure, I plan to do several negative tests in the future
I'll insert those aswell

tests/utils/exec/idps.go
```diff
@@ -51,7 +52,7 @@ func (idp *IDPService) Init(manifestDirs ...string) error {
 }
 
 func (idp *IDPService) Create(createArgs *IDPArgs, extraArgs ...string) error {
-	createArgs.URL = CON.GateWayURL
+	// createArgs.URL = CON.GateWayURL
```
amalykhi, line 55: Why commented?

shellymiron: `createArgs.URL` is not necessary in order to run the args as Iv'e seen.
but moreover - when we do use it, and there is an idp which require its own URL, the test fail as duplication.

amalykhi: Let's delete it is it not necessary and duplication?

shellymiron: ack, deleting

PR #395, opened 2023-11-21
OCM-4931: Remove a validation in machinepool so it can be created with the subnet from different AZ than the cluster was created - for singleAZ cluster

bardielle:
@amalykhi


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.4.1 (Nov 20, 2023)
```
gdbranco, line 1: Extra change I believe

bardielle: This is the change log for official release version, it's a cherry-pick from v1.4.1 
we will release that bug fix int he next release I guess

PR #396, opened 2023-11-22
Add tf import IDP test cases and methods - part 1

shellymiron:
Based on this test case: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-65981

We have 3 tf import cases to automate: `cluster`, `idp` and `machine-pool`. 
This is the first part of this effort, that will also declare how it is gonna be used/implemented for the rest of the resources. 

There is also negative tests in all of the 3 cases, we will consider/discussed if to separate to a different case or implement in the same one/s.

image


tests/ci/labels.go
```diff
@@ -17,6 +17,9 @@ var Day2 = Label("day2")
 // destroy
 var Destroy = Label("destroy")
 
+// import
+var ImportLabel = Label("tf-import")
```
amalykhi, line 21: may be use feature-import name - similar to our idp and mp cases.

tests/utils/exec/tf-exec.go
```diff
@@ -4,17 +4,53 @@ import (
...
+type ImportArgs struct {
+	URL          string `json:"url,omitempty"`
+	Token        string `json:"token,omitempty"`
+	ResourceKind string `json:"resource_kind,omitempty"`
+	ResourceName string `json:"resource_name,omitempty"`
+	ClusterID    string `json:"cluster_id,omitempty"`
+	ObjectName   string `json:"obj_name,omitempty"`
+}
+
+type ImportService struct {
```
amalykhi, line 29: as it is separate service - I would put it to separate file

shellymiron: fixed

tests/ci/labels.go
shellymiron: changed

PR #398, opened 2023-11-23
OCM-5013: Check `make generate` has being call per PR 

bardielle:


nirarg:
I don't think we need to add this dependency
Better to add another task validating all generated

scripts/check-gen.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
sagidayan, line 1: This script name is confusing. Its not really checking if we did run generate. It checks if there is any changes in the repo. (when running this after make generate - then the context is correct). I would rename this script and pass a parameter that will be the message. this way we can re-use if for other checks. not just the `generate`. but other.

PR #404, opened 2023-11-27
OCM-4927: Prevent trying to replace the default machinepool

JohnStrunk:
We have a number of attributes w/in the machinepool resource that cannot be changed once the pool has been created. These attributes are marked w/ "requiresReplace" so that if they are changed, TF will destroy and recreate the MP with the desired settings.
Unfortunately, it is not possible to recreate the default "worker" machinepool, so this process fails if a user edits one of these attributes for the default pool.

This change implements a new "plan modifier" that is similar to the existing "requiresReplace" modifier, except that it checks the name of the machinepool that is being affected:
- If it is the "worker" pool, it returns an error (at plan-time) before any changes are made to the user's cluster.
- For any other pool, it acts identically to the normal "requiresReplace" modifier

### DIFF
--- a/provider/machinepool/machine_pool_resource.go
+++ b/provider/machinepool/machine_pool_resource.go
@@ -32,7 +32,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
-	"github.com/hashicorp/terraform-plugin-framework/resource/schema/float64planmodifier"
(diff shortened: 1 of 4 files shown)


nirarg:
@JohnStrunk , thank you for this PR
We consider to remove all "RequireReplace" plan modifiers to avoid recreating the resources
In most of the places, we need to have the update blocked but it shouldn't cause to recreation
Any thoughts about this?
This is the Jira ticket: [OCM-4928](https://issues.redhat.com/browse/OCM-4928)

nirarg:
If we decide to keep using "RequiresReplace", than we need to merge this PR
We must handle "default machinepool" special case and avoid "RequiresReplace" in this case
@bardielle FYI

nirarg:
This PR can be closed, as it was decided to avoid using the "requiredReplace" modifier

PR #408, opened 2023-11-28
OCM-4983: Update machinepool error message re replicas/autoscaling

JohnStrunk:
Updates the error message when neither or both replicas and autoscaling are set.


JohnStrunk:
/cc @bardielle

@yuwang-RH

PR #412, opened 2023-11-29
OCM-5079 | fix: allow cluster compute SG IDs to be computed and block updates

gdbranco:
Related issue: OCM-5079

Allow cluster attribute `AWSAdditionalComputeSecurityGroupIds` to be computed by backend. So that on a state refresh it may be set to null if the API returns null. This scenario happens when removing the standard machine pool (named worker).

Also blocks further updates to any of the security groups attributes


provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -1113,6 +1116,33 @@ func (r *ClusterRosaClassicResource) Update(ctx context.Context, request resourc
...
+	_, shouldPatchAdditionalComputeSecurityGroupIds := common.ShouldPatchList(
+		state.AWSAdditionalComputeSecurityGroupIds, plan.AWSAdditionalComputeSecurityGroupIds)
+	if shouldPatchAdditionalComputeSecurityGroupIds {
+		response.Diagnostics.AddError(
+			cantUpdateClusterMessage,
+			"Cannot update cluster '%s', 'AWS Additional Compute Security Group IDs' cannot be updated")
+		return
+	}
+
+	_, shouldPatchAdditionalControlPlaneSecurityGroupIds := common.ShouldPatchList(
```
robpblake, line 1128: @gdbranco Is this even possible? The schema declares the `compute` and `infra` additional security groups as immutable. Doesn't that block the user from updating them?

gdbranco: I thought so too, but I'm not sure how that actually applies. Would need @nirarg or @bardielle help on this

Maybe because it's immutable it forces cluster destruction to create a new one

JohnStrunk: Going back to @robpblake 's question... The user cannot update the compute and infra SGs due to `Immutable()`. So, these tests *should* always be false, unless something is able to change these values out-of-band from TF. In that case, the code may try to restore it to the user's config, causing plan and state to differ here.

Some additional notes:
- `Immutable()` will never cause a resource replacement. It will just abort the run w/ an error
- For items that are actually immutable (whether we choose `Immutable()` or `RequiresReplace()`), I don't think they should be referenced during the `Update()` cycle. We can't change them, so there's no need to even look at them.

robpblake: @gdbranco @JohnStrunk I wonder if the problem here is actually with at least the `availability_zones` property?

IIUC, OCM-5079 relates to the cluster being recreated when the default MachinePool is removed by the user outside of the bounds of the Terraform plan in which the cluster is created. I believe when the user does this, the `nodes` element of the `cluster` spec is updated (or removed), which could impact the `availability_zones` property, setting it to a null list in the terraform state on the next read.

The schema for the Cluster lists the `availability_zones` as one of the properties using the `listplanmodifier.RequiresReplace()` which would cause the cluster to be recreated when TF compares the requested state against the actual state.

@gdbranco Do I have the behaviour of the CS API correct when the default MachinePool is removed?

cheers,
Rob

JohnStrunk: Before we guess what's going on, do we have a reproducer for this?
In the output of the plan, TF will tell us exactly why it's replacing a given resource.

PR #413, opened 2023-11-30
OCM-5027 | feat: Added support for KubeletConfig resource

robpblake:
This PR adds support for the `KubeletConfig` resource introduced as part of XCMSTRAT-110. Currently this allows customers to modify the `podPidsLimit` field of the `KubeletConfig` for all worker/infra nodes on their ROSA / OSD Classic clusters.

By default, we allow a range of `4,096 -> 16,384` for the value of the `podPidsLimit` field on `KubeletConfig`. However, for some customers they are allowed to exceed this default range if they have a pre-agreed capability with Red Hat. For those customers, the supported range is `4,096 -> 3694303` (this is referred to as the "unsafe" range).

As validation happens in offline mode in Terraform, we can only provide a warning to a customer if they request a value in the "unsafe" range e.g > 16,384. The OCM backend enforces the validation, and the `terraform apply` will fail if a customer is requesting a `podPidsLimit` that is not supported for their organisation. 

KubeletConfig is not currently supported on HCP clusters.


.gitignore
```diff
@@ -6,3 +6,4 @@ terraform.tfstate*
 /vendor
 /tests/rhcs_output
 terraform-provider-rhcs
+.idea
```
nirarg, line 9: Don't add IDE specific files to the repository git ignore
Each contributer/developer has the flexability to use his own IDE
You can add this to your local machine git ignore file
See for more info:
https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

provider/kubeletconfig/kubeletconfig_client.go
nirarg, line 1: This client wrapper doesn't have anything specific for terraform provider, therefore I suggest to consider moving it to https://github.com/openshift-online/ocm-common
This way it can be used also from the CLI

provider/kubeletconfig/kubeletconfig_clusterwait.go
nirarg, line 1: I suggest to use the `common/cluster_waiter.go directly`
You can improve the existing common file with the waiter structure you suggested here
And use the same way from all resources

provider/kubeletconfig/kubeletconfig_matcher.go
nirarg, line 1: I see this file is created only for tests, is it right?
Do you think this also should be moved to https://github.com/openshift-online/ocm-common?

.gitignore
robpblake: Done.

provider/kubeletconfig/kubeletconfig_clusterwait.go
robpblake: Good point. I've updated to move my structure into the `common/cluster_waiter.go` file and updated any code relying on it.

cheers,
Rob

provider/kubeletconfig/kubeletconfig_matcher.go
robpblake: @nirarg Updated this PR to use a common client from `ocm-common` 👍

provider/kubeletconfig/kubeletconfig_client.go
robpblake: @nirarg I've updated to use the common client from `ocm-common`.

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,261 @@
...
+	return ks
+}
+
+func (k *KubeletConfigResource) getKubeletConfigStateFromPlan(
+	ctx context.Context, plan tfsdk.Plan, diags *diag.Diagnostics) *KubeletConfigState {
+	ks := &KubeletConfigState{}
+	ds := plan.Get(ctx, ks)
+	diags.Append(ds...)
+	return ks
+}
```
nirarg, line 157: Those two functions are duplication, lets keep one with name `getKubeletConfigState`

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,261 @@
...
+func (k *KubeletConfigResource) writeStateToResponse(
+	ctx context.Context, ks *KubeletConfigState, tfState *tfsdk.State, diagnostics *diag.Diagnostics) {
+
+	diags := tfState.Set(ctx, ks)
+	diagnostics.Append(diags...)
+}
+
+func (k *KubeletConfigResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+
+	state := k.getKubeletConfigStateFromPlan(ctx, req.Plan, &resp.Diagnostics)
```
nirarg, line 194: I suggest to use here "plan" and not "state" to make it clear that the plan is used here (you can get the state as well)
```suggestion
	plan := k.getKubeletConfigState(ctx, req.Plan, &resp.Diagnostics)
```

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,261 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	clusterId := state.Cluster.ValueString()
+
+	_, err := k.configClient.Get(ctx, clusterId)
+	if err != nil {
```
nirarg, line 202: Use oneliner
```suggestion
	if_, err := k.configClient.Get(ctx, clusterId); err != nil {
```

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (k *KubeletConfigResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
+	response.TypeName = request.ProviderTypeName + resourceTypeName
+}
+
+func (k *KubeletConfigResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+
+	resp.Schema = schema.Schema{
+		Description: "KubeletConfig Configuration",
```
nirarg, line 69:
```suggestion
		Description: "KubeletConfig allows setting a customized Kubelet configuration",
```

docs/resources/kubeletconfig.md
```diff
@@ -0,0 +1,21 @@
+---
```
nirarg, line 1: Please add an example how to use this resource
This example should be added under [generate_example_usages](https://github.com/terraform-redhat/terraform-provider-rhcs/tree/main/generate_example_usages/resources), than the auto generator copy the example here

nirarg (commented):
Great work
Just few comments and we can push this

provider/kubeletconfig/kubeletconfig_resource.go
robpblake: @nirarg One function returns the state from `tfsdk.State` and the other from `tfsdk.Plan`. The functions that return the state using the `tfsdk.Plan` are called by the `delete` and `read` methods of the resource, which AFAICS, don't have access to the `tfsdk.State` struct.

Am I missing your point?

nirarg: Oh, I see it now, my mistake ... resolved :-)

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,257 @@
...
+func (k *KubeletConfigResource) getKubeletConfigStateFromPlan(
+	ctx context.Context, plan tfsdk.Plan, diags *diag.Diagnostics) *KubeletConfigState {
+	ks := &KubeletConfigState{}
+	ds := plan.Get(ctx, ks)
+	diags.Append(ds...)
+	return ks
+}
+
+func (k *KubeletConfigResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	plan := k.getKubeletConfigStateFromState(ctx, req.State, &resp.Diagnostics)
```
nirarg, line 157: Why did you change `state` to `plan` here
The comment wasn't relevant to here
Here you are using the `state` (and not the `plan`)

provider/kubeletconfig/kubeletconfig_resource.go
```diff
@@ -0,0 +1,257 @@
...
+			fmt.Sprintf("Failed to update existing KubeletConfig for cluster '%s': %v", clusterId, err))
+		return
+	}
+
+	k.convertApiResourceToState(updateResponse, plan)
+	k.writeStateToResponse(ctx, plan, &resp.State, &resp.Diagnostics)
+}
+
+func (k *KubeletConfigResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	plan := k.getKubeletConfigStateFromState(ctx, req.State, &resp.Diagnostics)
```
nirarg, line 223: Why did you change `state` to `plan` here
The comment wasn't relevant to here
Here you are using the `state` (and not the `plan`)

provider/kubeletconfig/kubeletconfig_resource.go
nirarg: I fixed this and going to push additional commit to this PR, to unblock the merge

provider/kubeletconfig/kubeletconfig_resource.go
nirarg: I fixed this and going to push additional commit to this PR, to unblock the merge

PR #415, opened 2023-11-30
OCM-4621: Change proxy validator to support only `additional_trust_bu…

bardielle:
…ndle` assignment

### DIFF
--- a/provider/clusterrosaclassic/cluster_rosa_classic_resource.go
+++ b/provider/clusterrosaclassic/cluster_rosa_classic_resource.go
@@ -1646,6 +1646,8 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster, 
 		state.AWSSubnetIDs = awsSubnetIds
 	}
 
+	hasProxy := true
+	hasAdditionalTrustBundle := true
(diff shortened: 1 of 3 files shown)


provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -1666,15 +1668,29 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
...
 	}
 
 	trustBundle, ok := object.GetAdditionalTrustBundle()
-	if ok && common.IsStringAttributeUnknownOrEmpty(state.Proxy.AdditionalTrustBundle) {
-		if state.Proxy == nil {
-			state.Proxy = &proxy.Proxy{}
+	if ok {
+		// update the state only if the current state is Null or Empty
+		// it can happen after `import` command or when it was updated from a different cli tool
+		if state.Proxy == nil || common.IsStringAttributeKnownAndEmpty(state.Proxy.AdditionalTrustBundle) {
```
nirarg, line 1681: I don't think this if statement is required
Is there any reason why not to set AdditionalTrustBundle when it already exist in the state?

PR #430, opened 2023-12-10
add rhcs_rosa_classic cluster import test case

shellymiron:
Test case: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-65684 
I noticed that there are parts of this specific test case that cannot be automated, are complicated and also interfere with other automated test cases that will run after (such as destroying the resource), hence, this test case will test the process of importing the cluster and checking the operation was successful. 
the negative part of this test will be automated on a separate PR.
we can open a discussion on that if there are other opinions ofcourse :)

### DIFF
--- a/tests/e2e/idps_test.go
+++ b/tests/e2e/idps_test.go
@@ -631,7 +631,7 @@ var _ = Describe("TF Test", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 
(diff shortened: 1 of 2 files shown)


shellymiron:
@amalykhi @xueli181114 @tshefi please review :) I will also appreciate your opinion on the way we should automate the test

PR #433, opened 2023-12-11
OCM-5233 | fix: remove compute SGs from state

gdbranco:
Related issue: [OCM-5233](https://issues.redhat.com//browse/OCM-5233)


gdbranco:
@JohnStrunk @nirarg PTAL

provider/cluster/cluster_resource.go
```diff
@@ -692,15 +695,6 @@ func populateClusterState(object *cmv1.Cluster, state *ClusterState) error {
 		state.AWSSubnetIDs = awsSubnetIds
 	}
 
-	additionalComputeSecurityGroupIds, ok := object.AWS().GetAdditionalComputeSecurityGroupIds()
```
bardielle, line 695: @gdbranco can you add a test case in the subsystem that tests the update flow?

PR #441, opened 2023-12-18
OCM-5233 | fix: remove immutable from compute SGs

gdbranco:
Related issue: [OCM-5233](https://issues.redhat.com//browse/OCM-5233)

Immutable doesn't allow updates to the state even though we are not really updating or tracking this in the state at all


provider/machinepool/machine_pool_resource.go
```diff
@@ -898,7 +898,6 @@ func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *Mac
 			}
 		} else {
 			state.UseSpotInstances = types.BoolNull()
-			state.MaxReplicas = types.Int64Null()
```
gdbranco, line 901: I believe this was incorrectly set at this section as it is also handled slightly below within the actual autoscaling resources

PR #445, opened 2023-12-20
OCM-5346 | fix: Do not wait for clusters in terminal states to be ready

robpblake:
This PR updates the cluster wait functions to ensure that we only wait for a cluster that is in a state that can become `ready`. A cluster that is in a state `error` or `uninstalling` will never become `ready` and we can shortcut the waiting function.

### DIFF
--- a/provider/common/cluster_waiter.go
+++ b/provider/common/cluster_waiter.go
@@ -15,7 +15,7 @@ const pollingIntervalInMinutes = 2
 //go:generate mockgen -source=cluster_waiter.go -package=common -destination=mock_clusterwait.go
 type ClusterWait interface {
 	WaitForClusterToBeReady(ctx context.Context, clusterId string) error
-	RetryClusterReadiness(ctx context.Context,clusterId string, attempts int, sleep time.Duration, timeout int64) (*cmv1.Cluster, error)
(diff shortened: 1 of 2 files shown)


gdbranco:
lgtm. Would advise to wait for Nir or Danielle review :)

provider/common/cluster_waiter.go
```diff
@@ -47,15 +47,23 @@ func (dw *DefaultClusterWait) WaitForClusterToBeReady(ctx context.Context, clust
 	resource := dw.collection.Cluster(clusterId)
 	// We expect the cluster to already exist
 	// Try to get it and if result with NotFound error, return error to user
-	if resp, err := resource.Get().SendContext(ctx); err != nil && resp.Status() == http.StatusNotFound {
-		message := fmt.Sprintf("Cluster %s not found, error: %v", clusterId, err)
+	resp, err := resource.Get().SendContext(ctx)
```
bardielle, line 50: since this function is a duplication (or a specific case ) of RetryClusterReadiness, can you refactor the code to NOT use `WaitForClusterToBeReady` at all and use `RetryClusterReadiness`

bardielle: @robpblake  it's fine for me to approve it and fix it in a new PR as a refactoring for several resources

PR #447, opened 2023-12-20
OCM-4928: Checking attributes' assertion in cluster_rosa_classic reso…

bardielle:
…urce for attributes that cannot be changed


provider/common/helpers.go
```diff
@@ -194,3 +199,9 @@ func Sha1Hash(data []byte) (string, error) {
 func HasValue(val attr.Value) bool {
 	return !val.IsUnknown() && !val.IsNull()
 }
+
+func AssertStateAndPlanValue(stateAttr attr.Value, planAttr attr.Value, attrName string, diags *diag.Diagnostics) {
```
nirarg, line 203: The Assert prefix says that this function should return an error in case the condition is false
Why don't you return error?
Consider  the name: `ValidateEquals`
And add description: `ValidateEquals checks if given attributes are equal, if not add error to diagnostic`

So also that should affect caller function, change it to validate instade of assert

nirarg:
That is looks great for me
Just added one minor "function naming" comment

@gdbranco can you please also take a look?
We decided to go in this direction and remove the "RequiresReplace" in order to avoid potential bugs (like unexpected clusters destruction)
See [OCM-4928](https://issues.redhat.com/browse/OCM-4928) and its linked bugs for more info

nirarg (commented):
@bardielle one more comment
Please remove the Immutable function, to make sure it's not used anymore

docs/resources/cluster_rosa_classic.md
```diff
@@ -47,47 +47,47 @@ resource "rhcs_cluster_rosa_classic" "rosa_sts_cluster" {
 
 ### Required
 
-- `aws_account_id` (String) Identifier of the AWS account.
+- `aws_account_id` (String) Identifier of the AWS account.Once the resource was created, the attribute cannot be changed.
```
gdbranco, line 50:
```suggestion
- `aws_account_id` (String) Identifier of the AWS account.Once the resource is created, the attribute cannot be changed.
```

Not really sure if this is a better terminology or not. Might be nice to talk about it with other folks

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -1053,6 +1018,42 @@ func (r *ClusterRosaClassicResource) Read(ctx context.Context, request resource.
...
+	common.AssertStateAndPlanValue(state.Private, plan.Private, "private", &diags)
+	common.AssertStateAndPlanValue(state.AvailabilityZones, plan.AvailabilityZones, "availability_zones", &diags)
+	common.AssertStateAndPlanValue(state.MachineCIDR, plan.MachineCIDR, "machine_cidr", &diags)
+	common.AssertStateAndPlanValue(state.ServiceCIDR, plan.ServiceCIDR, "service_cidr", &diags)
+	common.AssertStateAndPlanValue(state.PodCIDR, plan.PodCIDR, "pod_cidr", &diags)
+	common.AssertStateAndPlanValue(state.HostPrefix, plan.HostPrefix, "host_prefix", &diags)
+	common.AssertStateAndPlanValue(state.ChannelGroup, plan.ChannelGroup, "channel_group", &diags)
+	common.AssertStateAndPlanValue(state.Ec2MetadataHttpTokens, plan.Ec2MetadataHttpTokens, "ec2_metadata_http_tokens", &diags)
+	common.AssertStateAndPlanValue(state.AWSAdditionalControlPlaneSecurityGroupIds, plan.AWSAdditionalControlPlaneSecurityGroupIds, "aws_additional_control_plane_security_group_ids", &diags)
+	common.AssertStateAndPlanValue(state.AWSAdditionalInfraSecurityGroupIds, plan.AWSAdditionalInfraSecurityGroupIds, "aws_additional_infra_security_group_ids", &diags)
```
gdbranco, line 1044: Missing `aws_additional_control_plane_security_group_ids`

docs/resources/cluster_rosa_classic.md
nirarg: In addition, maybe better to write this once in const and than use it in all places

provider/common/helpers.go
bardielle: I changed to `ValidateStateAndPlanEquals`

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
bardielle: thank you! nice catch

docs/resources/cluster_rosa_classic.md
bardielle: I changed it to be `After the creation of the resource, it is not possible to update the attribute value.`
And it's a const string in common package

PR #448, opened 2023-12-20
OCM-5345 | fix: Allow pod_pids_limit to be set via a variable

robpblake:
This PR updates the validator for the `pod_pids_limit` field of `KubeletConfig` to defer validation when the value is currently unknown or null. 

`null` values are handled by TF itself with a much clearer error message. `unknown` values are possible during the `configure` method of the provider when the value is set via a variable, see: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/accessing-values#when-can-a-value-be-unknown-or-null

### DIFF
--- a/provider/kubeletconfig/kubeletconfig_validators.go
+++ b/provider/kubeletconfig/kubeletconfig_validators.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


provider/kubeletconfig/kubeletconfig_validators.go
```diff
@@ -50,6 +50,10 @@ func (p PidsLimitValidator) MarkdownDescription(_ context.Context) string {
 
 func (p PidsLimitValidator) ValidateInt64(_ context.Context, req validator.Int64Request, resp *validator.Int64Response) {
 
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
```
nirarg, line 53: Please use the common function here
something like:
```suggestion
	if !common.HasValue(req.ConfigValue) {
```

robpblake:
@nirarg Updated to use the common function. Could you re-review please.

cheers,
Rob

PR #457, opened 2023-12-27
OCM-5407: Validate Immutable attributes not changed after magic import

nirarg:
This PR changes "validate No Immutable Attribute Change" validation to be done also from "magicImport" function (for worker/default machine pool).
In order to allow this, need to consider also unknow plan attributes, because "magicImport" is done as part of the "Create" function, in which it possible.

This change also fixes: [OCM-5408](https://issues.redhat.com/browse/OCM-5408)

### DIFF
--- a/provider/machinepool/machine_pool_resource.go
+++ b/provider/machinepool/machine_pool_resource.go
@@ -27,6 +27,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-validators/float64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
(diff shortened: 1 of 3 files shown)


provider/machinepool/machine_pool_resource.go
```diff
@@ -522,7 +522,7 @@ func (r *MachinePoolResource) readState(ctx context.Context, state *MachinePoolS
 	return
 }
 
-func assertNoChanges(state, plan *MachinePoolState) diag.Diagnostics {
+func validateNoImmutableAttChange(state, plan *MachinePoolState) diag.Diagnostics {
```
bardielle, line 525: I think it's bettek to have a name conversion - so if you change here the name, please update in all other resources...

nirarg: Done - change the function in cluster rosa resource as well

PR #460, opened 2023-12-28
OCM-5285: MP labels map must have a size of at least 1

sagidayan:
If users wish to delete all labels, they should simply remove the `labels` attribute from the `tf` statement.

### DIFF
--- a/subsystem/machine_pool_resource_test.go
+++ b/subsystem/machine_pool_resource_test.go
@@ -662,11 +662,23 @@ var _ = Describe("Machine pool creation", func() {
 					}
 				  },
 				  "replicas": 12,
-                  "labels": {}
+                  "labels": null
 				}`),
 			),
 		)
 
+		// Invalid deletion - labels map can't be empty
+		terraform.Source(`
+		  resource "rhcs_machine_pool" "my_pool" {
+		    cluster      = "123"
+		    name         = "my-pool"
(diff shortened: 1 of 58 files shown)


subsystem/machine_pool_resource_test.go
```diff
@@ -662,11 +662,23 @@ var _ = Describe("Machine pool creation", func() {
...
+		  resource "rhcs_machine_pool" "my_pool" {
+		    cluster      = "123"
+		    name         = "my-pool"
+		    machine_type = "r5.xlarge"
+		    replicas     = 12
+            labels       = {}
+		  }
+		`)
+		Expect(terraform.Apply()).ToNot(BeZero())
+		// Valid deletion
```
bardielle, line 681: Please also add an update flow:
1. Create MP with labels 
2. remove them by setting it to empty map -> should fail 
3. Remove them by setting it to null -> should succeed

bardielle:
@sagidayan please change the base branch to `v1.5.0-branch` and cherry pic to main

subsystem/machine_pool_resource_test.go
sagidayan: That is what I am doing in the test. 🤔  

in tf `=null` is like not passing the attribute. I can add that case but seems pointless. wdyt?

sagidayan:
PR was messed up. Will create a new one.

PR #462, opened 2023-12-28
OCM-5406 | fix: set att to empty if removed from plan

gdbranco:
Related issue: [OCM-5406](https://issues.redhat.com//browse/OCM-5406)

### DIFF
--- a/provider/defaultingress/ingress.go
+++ b/provider/defaultingress/ingress.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -102,6 +103,7 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
 			"cluster_routes_hostname": schema.StringAttribute{
 				Description: "Components route hostname for oauth, console, download.",
 				Optional:    true,
+				Validators:  []validator.String{stringvalidator.LengthAtLeast(1)},
 			},
 			"load_balancer_type": schema.StringAttribute{
 				Description: fmt.Sprintf("Type of Load Balancer. Options are %s.", strings.Join(validLbTypes, ",")),
@@ -112,6 +114,7 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
 			"cluster_routes_tls_secret_ref": schema.StringAttribute{
 				Description: "Components route TLS secret reference for oauth, console, download.",
 				Optional:    true,
+				Validators:  []validator.String{stringvalidator.LengthAtLeast(1)},
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


bardielle:
@gdbranco we had the same issue with the attribute `labels` in mp.
We decided not to set it to computed as a solution but to deal with that in the provider itself 
@sagidayan can you help here?

sagidayan (changes requested):
Can you please try this method instead?
1. remove computed on both attributes
2. import `"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"`
3. add a validator to both as such:  `Validators: []validator.String{stringvalidator.LengthAtLeast(1)}`

Then retest. Changing values in your `.tf` to an empty string will now throw an error saying you cannot do that. To remove the values, you should be able to simply remove them from your `.tf` configuration or set them to null.

Should resolve the inconsistent result error.

This change will be inline with the idea that these attrs are not computed, and they don't need to be.

Please let me know if this is a suitable solution in your opinion.

tsorya:
Can't we just set nil in case of empty string?

sagidayan:
@tsorya then we will need to set the attrs to be computed. And I am not convinced that they need to.

tsorya:
The problem is that we can't sent nil to ocm as it does nothing so we have nil till those fields were not set but in order to empty them we need to send empty string. That will always create case that we will not have match between ocm and tf like deleting those fields from main.tf that will actually always have empty string value in ocm.
I agree with Sagi, we should just block setting empty string

sagidayan:
@gdbranco Can you please add the proper tests to the subsystem (i.e passing an empty string should result in an error)

provider/defaultingress/ingress.go
```diff
@@ -389,10 +392,18 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan
 		ingressBuilder := getDefaultIngressBuilder(ctx, plan)
 
 		if !reflect.DeepEqual(state.ClusterRoutesHostname, plan.ClusterRoutesHostname) {
-			ingressBuilder.ClusterRoutesHostname(plan.ClusterRoutesHostname.ValueString())
+			value := ""
+			if !plan.ClusterRoutesHostname.IsNull() {
```
bardielle, line 396: What will happen if `state.ClusterRoutesHostname="stringA" && plan.ClusterRoutesHostname = nil` ?
I think we will not update the backend in that case

gdbranco: Not sure if I follow @bardielle . If they are different we do want to update to what the plan value is, as it's the new value. If the plan is nil means we want to remove the setting, which is equivalent to supplying '' (empty) to the api

subsystem/default_ingress_resource_test.go
```diff
@@ -338,6 +338,50 @@ var _ = Describe("default ingress", func() {
 		Expect(terraform.Apply()).To(BeZero())
 	})
 
+	It("Create cluster with default ingress - cluster_routes_hostname and cluster_routes_tls_secret_ref set to empty value", func() {
+		// Prepare the server:
+		server.AppendHandlers(
```
sagidayan, line 343: If im not mistaken, You don't need all these server handlers. once you "apply" with an empty string, terraform should fail before hitting any API (due to the validation)

subsystem/default_ingress_resource_test.go
```diff
@@ -338,6 +338,50 @@ var _ = Describe("default ingress", func() {
...
+							"dns_name": "redhat.com",
+							"load_balancer_type": "classic",
+							"route_wildcard_policy": "WildcardsDisallowed",
+							"route_namespace_ownership_policy": "Strict"
+						}
+					`),
+			),
+		)
+		// Run the apply command:
+		terraform.Source(`resource "rhcs_default_ingress" "default_ingress" {
```
sagidayan, line 377: can you add 3 configs? one with both empty strings and 2 more for each one empty while the other is not. all 3 should fail, right?

provider/defaultingress/ingress.go
nirarg: The purpose of not allowing empty string in the state is to avoid from having two kinds of valid empty values ("" and null), but we can continue use the null.
When user sets null, its means he wants to delete the existing value, which is legal action.
Having that, null is the only empty value which can be passed to the API and return from it
I agree with @bardielle that we shouldn't ignore this null value and this should resulted by removing the existing value.

sagidayan:
Oh, and please change the PR title and commit message... 😉  They are not reflecting the change now.

subsystem/default_ingress_resource_test.go
gdbranco: If we are adding tests were they are set individually I really don't think that an extra for both means much as the configuration is individual anyways :thinking:

provider/defaultingress/ingress.go
gdbranco: If I'm not mistaken null is not accepted by the API though. The way to remove the existing value would be to remove the attribute from the TF plan :thinking:

bardielle: Can you test it manually? 
I mean update the attribute `ClusterRoutesHostname` to be null and use the ocm API query to validate it was removed

gdbranco: Yes, just finished. Setting to either null or removing the attributes works as expected. Setting the attribute to empty fails the plan step

PR #464, opened 2023-12-28
OCM-5285: MP labels can't be an empty map

sagidayan:
If users wish to delete all labels, they should simply remove the `labels` attribute from the tf statement.


bardielle:
Add more scenarios for subsystem please

sagidayan:
> Add more scenarios for subsystem please

All the update tests are there. My change is a part of a full create/update/delete flow.

osherdp:
@sagidayan @bardielle  just something that I'm curious of 
Usually backports/hotfixes are first happening at upstream and them propagated to older releases/branches
Any reason we're doing the opposite here?

PR #476, opened 2023-12-31
automate ocp-64907 :: validate custom property operations on cluster 

shellymiron:
Automation test case : [OCP-64907](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-64907)

The test purpose is to validate custom property operations on cluster - such as adding, deleting, editing custom property to an existing cluster.

Depends on this [PR](https://github.com/terraform-redhat/terraform-provider-rhcs/pull/477).

image

### DIFF
--- a/tests/e2e/general_day_two_test.go
+++ b/tests/e2e/general_day_two_test.go
@@ -1,6 +1,8 @@
 package e2e
 
 import (
+	"path/filepath"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	ci "github.com/terraform-redhat/terraform-provider-rhcs/tests/ci"
@@ -11,6 +13,7 @@ import (
 )
 
 var _ = Describe("TF day2 scenrios", func() {
+
 	Context("Author:smiron-Medium-OCP-67570 @OCP-67570 @smiron", func() {
 		var dnsService *exe.DnsService
 		BeforeEach(func() {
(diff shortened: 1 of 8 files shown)


tests/e2e/general_day_two_test.go
```diff
@@ -68,4 +69,74 @@ var _ = Describe("TF day2 scenrios", func() {
...
+		BeforeEach(func() {
+			// Load profile from YAML file based on environment
+			profile = ci.LoadProfileYamlFileByENV()
+
+			// Initialize the cluster service
+			clusterService, err = exe.NewClusterService(profile.ManifestsDIR)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Capture the original custom properties
+			originalCustomProperties = map[string]string{
```
xueli181114, line 89: We should record the original properties by getting cluster detail rather than read from constants. In future we will have other properties added.

shellymiron: ack, to fix it, I created a new function that called `ReadTerraformTFVars`,
which will read `terraform.tfvars` file recorded when the test began, and return its content as a map in the BeforeEach. late on, in AfterEach we will restore to the original property by applying the var.

shellymiron:
image

Re-run after adding a new function -` ReadTerraformTFVars()`

tests/utils/exec/tf-exec.go
```diff
@@ -10,10 +10,12 @@ import (
 	"os/exec"
 	"path"
 	"strconv"
+	"strings"
 
 	CON "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/constants"
 	h "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/helper"
 	. "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/log"
+	l "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/log"
```
xueli181114, line 18: log had been imported, why import again?

PR #485, opened 2024-01-03
OCM-5425 | fix: do not set default values if not supplied

gdbranco:
Related issue: [OCM-5425](https://issues.redhat.com//browse/OCM-5425)

# What
Only sends the value if the plan is actually different from the state which reveals an actual intention to adjust those attributes

# Why
TF provider always sent default values even when not required. Because the API responds by validating each attribute individually it would always respond that the `Route selector` attribute is invalid as it's the first one to be validated.

### DIFF
--- a/provider/defaultingress/ingress.go
+++ b/provider/defaultingress/ingress.go
@@ -389,7 +389,7 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan 
 			plan = &DefaultIngress{}
 		}
 
-		ingressBuilder := getDefaultIngressBuilder(ctx, plan)
+		ingressBuilder := getDefaultIngressBuilder(ctx, state, plan)
 
 		if !reflect.DeepEqual(state.ClusterRoutesHostname, plan.ClusterRoutesHostname) {
 			value := ""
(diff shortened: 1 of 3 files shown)


bardielle:
Please add a subsystem test

provider/defaultingress/ingress.go
```diff
@@ -425,31 +425,34 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan
...
 	ingressBuilder := cmv1.NewIngress()
-	routeSelectors, err := common.OptionalMap(ctx, state.RouteSelectors)
-	if err != nil {
-		return nil
+	if !reflect.DeepEqual(state.RouteSelectors, plan.RouteSelectors) {
+		routeSelectors, err := common.OptionalMap(ctx, plan.RouteSelectors)
+		if err != nil {
+			return nil
+		}
+		if routeSelectors == nil {
```
bardielle, line 435: why can't we sent the nil as is (in case of nil list we sending it as a nil object)

bardielle:
Not sure I understood your solution. Is sounds that the issue is in the backend if it can't get an empty object (list or map)....

provider/defaultingress/ingress.go
gdbranco: If we send nil the SDK does not set the value in the request. As it omits nil values, so then the API will not pick this up

gdbranco: This was already being sent as empty, but there's was no check that the plan differed from the current state

nirarg: Why is it needed to be validated?
In case there is no change that updating the same value should do nothing, right?

gdbranco: Depending on the version and also some conditions there are values that cannot be supplied. By setting this to some value without comparison it ends up trying to supply when it should be nil and API responds with error

PR #488, opened 2024-01-04
add qe_usage var env to track leakages

shellymiron:
The new custom property - QE_USAGE, will help us to track after leakages by indicating the name of
the profile from which the cluster was created and did not teardown.

Depends on this [PR](https://github.com/openshift/release/pull/47237)

### DIFF
--- a/tests/ci/profile_handler.go
+++ b/tests/ci/profile_handler.go
@@ -198,7 +198,6 @@ func GenerateClusterCreationArgsByProfile(token string, profile *Profile) (clust
 	profile.Version = PrepareVersion(RHCSConnection, profile.Version, profile.ChannelGroup, profile)
 
 	clusterArgs = &EXE.ClusterCreationArgs{
-		// Token:            token,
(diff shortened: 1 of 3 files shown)


tests/utils/exec/cluster.go
```diff
@@ -31,6 +31,8 @@ type ClusterCreationArgs struct {
 	DefaultMPLabels                      map[string]string `json:"default_mp_labels,omitempty"`
 	DisableSCPChecks                     bool              `json:"disable_scp_checks,omitempty"`
 	MultiAZ                              bool              `json:"multi_az,omitempty"`
+	Properties                           map[string]string `json:"properties,omitempty"`
+	CustomProperties                     map[string]string `json:"custom_properties,omitempty"`
```
xueli181114, line 35: Why Properties and another CustomProperties? Let's keep one only. Is properties used anywhere?

shellymiron: you are right, fixed :)

PR #490, opened 2024-01-07
add the option to destroy import idp resources

shellymiron:
small addition functionality to the idp import test

image

### DIFF
--- a/tests/e2e/idps_test.go
+++ b/tests/e2e/idps_test.go
@@ -780,11 +780,13 @@ var _ = Describe("TF Test", func() {
 			})
 
 			AfterEach(func() {
-				err := idpService.google.Destroy()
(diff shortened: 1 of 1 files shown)


tests/e2e/idps_test.go
```diff
@@ -780,11 +780,12 @@ var _ = Describe("TF Test", func() {
...
 
 			AfterEach(func() {
-				err := idpService.google.Destroy()
-				Expect(err).ToNot(HaveOccurred())
 
+				By("Clean resources after test")
 				err = idpService.gitlab.Destroy()
 				Expect(err).ToNot(HaveOccurred())
+
+				importService.Destroy()
```
xueli181114, line 788: We need to check the destroy result with Expect.

shellymiron: tnx! fixed

PR #491, opened 2024-01-07
 OCM-5428 | OCM-5414 |  OCM-5409: machine pool data source

bardielle:
This PRs:
* [OCM-5414](https://issues.redhat.com//browse/OCM-5414): add availability_zonese as a computed attribute to machinepool resource
* [OCM-5409](https://issues.redhat.com//browse/OCM-5409): add aws_submet_ids as a computed attribute to machinepool resource
* [OCM-5428](https://issues.redhat.com//browse/OCM-5428): add machine-pool datasource
* Generate docs


bardielle:
output:
```

terraform state show data.rhcs_machine_pool.mp2
# data.rhcs_machine_pool.mp2:
data "rhcs_machine_pool" "mp2" {
    availability_zones      = [
        "us-east-2a",
        "us-east-2b",
        "us-east-2c",
    ]
    aws_subnet_ids          = [
        "subnet-0afdd90afd6008ff0",
        "subnet-04c6fcc2ce591fb4b",
        "subnet-0a437951b8b316357",
    ]
    cluster                 = "28k1v2l8pofsbmor1injvng0d8i7ie3q"
    disk_size               = 300
    id                      = "terr-cluster-machine-pool"
    machine_type            = "m5.xlarge"
    multi_availability_zone = true
    name                    = "terr-cluster-machine-pool"
    replicas                = 6
}
```

provider/machinepool/machine_pool_datasource.go
```diff
@@ -0,0 +1,199 @@
...
+	if notFound {
+		// If we can't find the machine pool, it was deleted. Remove if from the
+		// state and don't return an error so the TF apply() will automatically
+		// recreate it.
+		tflog.Warn(ctx, fmt.Sprintf("machine pool (%s) of cluster (%s) not found, removing from state",
+			state.ID.ValueString(), state.Cluster.ValueString(),
+		))
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
nirarg, line 192: This shouldn't have the same behavior like in the resource
You are removing here resource while you actually handle data source
I think it just should return error the the data source doesn't exist, but not sure
What do you think?

provider/machinepool/machine_pool_resource.go
```diff
@@ -184,6 +189,11 @@ func (r *MachinePoolResource) Schema(ctx context.Context, req resource.SchemaReq
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"aws_subnet_ids": schema.ListAttribute{
+				Description: "AWS subnet IDs. ",
```
nirarg, line 193:
```suggestion
				Description: "A list of IDs of subnets in which the machines of this machine pool are created. Relevant only for a machine pool with multiple subnets. For machine pool with single subnet check \"subnet_id\" attribute"
				"AWS subnet IDs.",
```

provider/machinepool/machine_pool_resource.go
```diff
@@ -184,6 +189,11 @@ func (r *MachinePoolResource) Schema(ctx context.Context, req resource.SchemaReq
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"aws_subnet_ids": schema.ListAttribute{
```
nirarg, line 192: Consider to remove the "aws_" prefix.
I understand that it better to have this, but maybe in this case it preferred to have it similar to "sunbnet_id" attribute

provider/machinepool/machine_pool_resource.go
```diff
@@ -989,6 +999,11 @@ func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *Mac
 	} else {
 		state.AvailabilityZone = types.StringValue("")
 	}
+	azListValue, err := common.StringArrayToList(azs)
+	if err != nil {
+		return err
+	}
+	state.AvailabilityZones = azListValue
```
nirarg, line 1006: I think that this should be done only in case `len(azs) != 1`, to avoid having two confusing similar attributes

provider/machinepool/machine_pool_resource.go
```diff
@@ -997,6 +1012,11 @@ func (r *MachinePoolResource) populateState(object *cmv1.MachinePool, state *Mac
 	} else {
 		state.SubnetID = types.StringValue("")
 	}
+	subnetListValue, err := common.StringArrayToList(subnets)
+	if err != nil {
+		return err
+	}
+	state.AWSSubnetIDs = subnetListValue
```
nirarg, line 1019: I think that this should be done only in case `len(subnets) != 1`, to avoid having two confusing similar attributes

provider/machinepool/machine_pool_resource.go
bardielle: I tried to be aligned with cluster rosa resource

bardielle: I changed it to subnet_ids

PR #494, opened 2024-01-08
OCM-5495: Remove depricated Dockerfile

amalykhi:


amalykhi:
@bardielle Please, review

PR #501, opened 2024-01-11
Fix OCP-65684 :: remove state file to prevent destructive import

shellymiron:
`defer h.DeleteFile(con.GrantTFstateFile(con.ImportResourceDir))`

this will delete the state file after importing the rosa sts cluster, and after that we can run other machine-pool/idp import cases without issues.

### DIFF
--- a/tests/e2e/verfication_post_day1_test.go
+++ b/tests/e2e/verfication_post_day1_test.go
@@ -146,7 +146,7 @@ var _ = Describe("TF Test", func() {
 
 		Context("Author:smiron-Medium-OCP-65684 @OCP-65684 @smiron", func() {
 			It("OCP-65684 - cluster_rosa_classic resource can be import by the terraform import command",
-				ci.Day2, ci.Medium, ci.FeatureImport, ci.Exclude, func() {
+				ci.Day2, ci.Medium, ci.FeatureImport, func() {
 
 					By("Run the command to import rosa_classic resource")
 					importParam := &exe.ImportArgs{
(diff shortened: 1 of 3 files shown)


tests/e2e/verfication_post_day1_test.go
```diff
@@ -176,8 +176,9 @@ var _ = Describe("TF Test", func() {
 					err = importService.Import(importParam)
 					Expect(err.Error()).To(ContainSubstring("Cannot import non-existent remote object"))
 
-					// skip due to bug :: OCM-5246
-					// Expect(output).To(ContainSubstring(profile.ComputeMachineType))
+					By("remove .tfstate file to prevent destructive outcomes")
+					defer h.DeleteFile(con.GrantTFstateFile(con.ImportResourceDir))
```
amalykhi, line 180: May be add function to the Service with the clean up. @xueli181114  WDYT?

tests/e2e/verfication_post_day1_test.go
```diff
@@ -176,8 +176,9 @@ var _ = Describe("TF Test", func() {
 					err = importService.Import(importParam)
 					Expect(err.Error()).To(ContainSubstring("Cannot import non-existent remote object"))
 
-					// skip due to bug :: OCM-5246
-					// Expect(output).To(ContainSubstring(profile.ComputeMachineType))
+					By("remove .tfstate file to prevent destructive outcomes")
```
amalykhi, line 179: I don't  think we should use words like destructive - regular clean up(revert) after the test execution)

shellymiron: fixed :)

tests/e2e/verfication_post_day1_test.go
amalykhi: You are right. But may be add function to utils somewhere, where we will give manifest folder to clean up the state?

shellymiron: ack, I added a function :D

shellymiron:
image

PR #505, opened 2024-01-15
fix typo in ingress error message

lihongan:
Currently the actual output looks like 
```
│ Cannont upate default ingress forcluster '28gjcif4b7fluce0uksce104492bjk0h': status is 400 ...
``` 

Expected:
```
Cannot update default ingress for cluster ...
```


lihongan:
cc @gdbranco

PR #509, opened 2024-01-18
OCM-5632 | feat: support password generate/hash in terraform provider

marcolan018:

This PR is to add the feature for secure cluster credential to terraform provider. 

Fixes #
https://issues.redhat.com/browse/OCM-5632

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -397,12 +399,13 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 				Attributes: map[string]schema.Attribute{
 					"username": schema.StringAttribute{
```
nirarg, line 400: So is it expected that the user never provides the username?
Why not allow this? I see that the code below allows the case in which the user add their own username.

Other thought here, it doesn't look so intuitive that user need to add empty list in order to trigger the admin credentials, maybe better also to add an boolean as well?

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -76,6 +77,7 @@ const (
 	propertyRosaTfCommit      = tagsPrefix + "tf_commit"
 	waitTimeoutInMinutes      = 60
 	DefaultMachinePoolMessage = "This attribute is specifically applies for the default Machine Pool and becomes irrelevant once the resource is created. Any modifications to the default Machine Pool should be made through the Terraform imported Machine Pool resource. For more details, refer to [Default Machine Pool in ROSA Cluster](../guides/worker-machine-pool.md)"
+	clusterAdminUserName      = "cluster-admin"
```
nirarg, line 80: Shouldn't this constant be in ocm-common?

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -681,11 +684,35 @@ func createClassicClusterObject(ctx context.Context,
 
 	if state.AdminCredentials != nil {
 		htpasswdUsers := []*cmv1.HTPasswdUserBuilder{}
+		var username, password string
+		if common.HasValue(state.AdminCredentials.Username) {
+			username = state.AdminCredentials.Username.ValueString()
+		} else {
+			username = clusterAdminUserName
+		}
```
nirarg, line 692:
```suggestion
                username := clusterAdminUserName
		var  password string
		if common.HasValue(state.AdminCredentials.Username) {
			username = state.AdminCredentials.Username.ValueString()
		}
```

provider/identityprovider/htpasswd.go
```diff
@@ -75,19 +76,23 @@ var htpasswdUserList = map[string]schema.Attribute{
...
 	userList := []*cmv1.HTPasswdUserBuilder{}
 	for _, user := range state.Users {
+		hashedPwd, err := idputils.GenerateHTPasswdCompatibleHash(user.Password.ValueString())
+		if err != nil {
+			return nil, err
+		}
 		userBuilder := &cmv1.HTPasswdUserBuilder{}
 		userBuilder.Username(user.Username.ValueString())
-		userBuilder.Password(user.Password.ValueString())
+		userBuilder.HashedPassword(hashedPwd)
```
nirarg, line 90: We should keep the option for users to add their own password, no?

nirarg (commented):
Started review, few comments

Please change  remove the feat from the PR subject to have `OCM-5632: ...`, this will be resulted by adding this PR as a link in the jira ticket

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: @nirarg 
thanks for the review.
Yes, we should support user-provided username, and if not provided, "cluster-admin" is used.
For that logic, what's the correct configuration for this attribute? Is it ok as below?

```
					"username": schema.StringAttribute{
						Description: "Admin username that will be created with the cluster.",
						Default:    "cluster-admin",
						Validators:  identityprovider.HTPasswdUsernameValidators,
					},
```

provider/identityprovider/htpasswd.go
marcolan018: yes, the hashedpassword is calculated from the user provided one: 
hashedPwd, err := idputils.GenerateHTPasswdCompatibleHash(user.Password.ValueString())

for idp creation, we don't provide the option to generate the password. Just wants to keep the same behavior here as rosa cli

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: fixed

nirarg:
Please complete the following:

1. Generate documentation by running `make generate`
2. Add subsystem tests and fix existing tests following the changes

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -393,22 +395,29 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 					"upgrade to OpenShift 4.12.z from 4.11 or before).",
 				Optional: true,
 			},
+			"create_admin_user": schema.BoolAttribute{
+				Description: "Indicates if create cluster admin user",
```
bardielle, line 399: Please elaborate on all the optional usages of that attribute with (and without) the `admin_credentials` attribute

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -680,13 +689,37 @@ func createClassicClusterObject(ctx context.Context,
...
-	if state.AdminCredentials != nil {
-		htpasswdUsers := []*cmv1.HTPasswdUserBuilder{}
-		htpasswdUsers = append(htpasswdUsers, cmv1.NewHTPasswdUser().
-			Username(state.AdminCredentials.Username.ValueString()).Password(state.AdminCredentials.Password.ValueString()))
+	if common.HasValue(state.CreateAdminUser) && state.CreateAdminUser.ValueBool() ||
+		common.HasValue(state.AdminCredentials) {
+		if !common.HasValue(state.AdminCredentials) {
+			state.AdminCredentials = flattenAdminCredentials("", "")
+		}
+		username, password := expandAdminCredentials(ctx, state.AdminCredentials, diags)
```
nirarg, line 697:
```suggestion
	username, password := expandAdminCredentials(ctx, state.AdminCredentials, diags)
	if common.BoolWithFalseDefault(state.CreateAdminUser) || common.HasValue(state.AdminCredentials) {
```
This is refactoring `refactoring-1`
See comment with refactoring `refactoring-2` for the whole suggestion

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -680,13 +689,37 @@ func createClassicClusterObject(ctx context.Context,
...
+		htpasswdUsers := []*cmv1.HTPasswdUserBuilder{
+			cmv1.NewHTPasswdUser().Username(username).HashedPassword(hashedPwd),
+		}
 		htpassUserList := cmv1.NewHTPasswdUserList().Items(htpasswdUsers...)
 		htPasswdIDP := cmv1.NewHTPasswdIdentityProvider().Users(htpassUserList)
 		builder.Htpasswd(htPasswdIDP)
+		state.AdminCredentials = flattenAdminCredentials(username, password)
+	} else {
+		state.AdminCredentials = flattenAdminCredentials("", "")
 	}
```
nirarg, line 723:
```suggestion
	}
	state.AdminCredentials = flattenAdminCredentials(username, password)
```
his is refactoring `refactoring-2`
See comment with refactoring `refactoring-1` for the whole suggestion

provider/clusterrosaclassic/cluster_rosa_classic_state.go
```diff
@@ -99,3 +105,34 @@ type PrivateHostedZone struct {
 	ID      types.String `tfsdk:"id"`
 	RoleARN types.String `tfsdk:"role_arn"`
 }
+
+func flattenAdminCredentials(username string, password string) types.Object {
```
bardielle, line 109: Please add another function for ObjectNull, maybe something like that 
```
func flattenAdminCredentials() types.Object {
 return types.ObjectNull(map[string]attr.Type{
		"username": types.StringType,
		"password": types.StringType,
	})
}

```

nirarg: Or alternatively it can be like:
```
func adminCredentialsNull() types.Object {
  return flattenAdminCredentials("", "")
}
```
Up to you :-)

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: fixed

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: added more details in the Description

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: fixed

provider/clusterrosaclassic/cluster_rosa_classic_state.go
marcolan018: I select the 2nd option for clean code
thanks

marcolan018:
@nirarg 
in the latest commit, I added the subsystem test for the changes, can you take a look at it? Thanks

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -995,6 +1027,17 @@ func (r *ClusterRosaClassicResource) Read(ctx context.Context, request resource.
 	response.Diagnostics.Append(diags...)
 }
 
+func validateAdminCredentialsChange(ctx context.Context, state,
+	plan *ClusterRosaClassicState) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	stateUsername, statePassword := expandAdminCredentials(ctx, state.AdminCredentials, diags)
+	planUsername, planPassword := expandAdminCredentials(ctx, plan.AdminCredentials, diags)
+	if planUsername != "" && stateUsername != planUsername || planPassword != "" && statePassword != planPassword {
```
nirarg, line 1035: Is it possible to create the AdminCredentials after the cluster creation?
Following this if statement I understand it possible, but is it possible by the API?
If it possible, so should the update also check if the boolean `create_admin_user` was changed from false to true?

marcolan018: added the check for `create_admin_user`
also added the subsystem test
@nirarg please help to take a look, thanks

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -995,6 +1027,27 @@ func (r *ClusterRosaClassicResource) Read(ctx context.Context, request resource.
 	response.Diagnostics.Append(diags...)
 }
 
+func validateAdminCredentialsChange(ctx context.Context, state,
+	plan *ClusterRosaClassicState) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	if state.AdminCredentials.IsNull() {
+		if common.BoolWithFalseDefault(plan.CreateAdminUser) {
+			diags.AddError(common.AssertionErrorSummaryMessage, fmt.Sprintf(common.AssertionErrorDetailsMessage, "create_admin_user", false, true))
+		}
```
nirarg, line 1036: What happen if AdminCredentials object exists and user change CreateAdminUser flag from true to false?
I think this case is not covered here

So maybe its better just to compare the bool value between the plan and state?
Something like:
```
common.ValidateStateAndPlanEquals(state.CreateAdminUser, plan.CreateAdminUser, "create_admin_user", &diags)
```
It can even be outside from 'validateAdminCredentialsChange' function

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -995,6 +1027,27 @@ func (r *ClusterRosaClassicResource) Read(ctx context.Context, request resource.
 	response.Diagnostics.Append(diags...)
 }
 
+func validateAdminCredentialsChange(ctx context.Context, state,
+	plan *ClusterRosaClassicState) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	if state.AdminCredentials.IsNull() {
+		if common.BoolWithFalseDefault(plan.CreateAdminUser) {
```
nirarg, line 1034: Nit:
I suggest to create equals function, something like:
```
func adminCredentialsEqual(ctx context.Context, object1 types.Object, object2 types.Object, diags diag.Diagnostics) boolean
```
Then use it for the validation:
```
if ! adminCredentialsEqual(ctx, state.AdminCredentials , plan.AdminCredentials , diags) {
    diags.AddError(common.AssertionErrorSummaryMessage, fmt.Sprintf(common.AssertionErrorDetailsMessage, "admin_credentials", state.AdminCredentials, plan.AdminCredentials))
}
```

nirarg: Or in second thought, why not keep using the deepEqual, like it done before?
```
if !reflect.DeepEqual(state.AdminCredentials, plan.AdminCredentials) {
	diags.AddError(common.AssertionErrorSummaryMessage, fmt.Sprintf(common.AssertionErrorDetailsMessage, "admin_credentials", *state.AdminCredentials, *plan.AdminCredentials))
}
```

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -1704,6 +1759,9 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
 	state.State = types.StringValue(string(object.State()))
 	state.Name = types.StringValue(object.Name())
 	state.CloudRegion = types.StringValue(object.Region().ID())
+	if state.AdminCredentials.IsUnknown() {
+		state.AdminCredentials = adminCredentialsNull()
+	}
```
nirarg, line 1764: This is not needed here because in the createCluster Null is assigned as a default value

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: @nirarg 
in current logic, if AdminCredentials exists, the value of CreateAdminUser will be ignored.
So for the case `AdminCredentials object exists and user change CreateAdminUser flag from true to false`, the flow will enter the `else` section to compare the `AdminCredentials` value in state&plan
is this ok?

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: I prefer to add equals function, because the `AdminCredentials` object has computed value. Even the user tries to apply the same plan as before, the `DeepEqual` will return false when compare `AdminCredentials` in state and plan.
e.g. user provides an empty `AdminCredentials` obj, and after apply, in the state the obj will have the default username and generated password. If user applies the original plan again, the `DeepEqual` func will find the empty `AdminCredentials` obj in plan doesn't equal to the one in state which already have the computed value.
I am not 100% sure about this is the correct behavior in a terraform provider for computed attribute.  What's your thoughts on this? @nirarg

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: @nirarg 
if w/o this, in update scenario, I will get the error `Attribute admin_credentials, cannot be changed from <null> to <unknown>`.

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
nirarg: Ok, I'm ok with it

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
nirarg: But I think that if users change it we should tell them explicit that it shouldn't be changed, otherwise they might expect to different result.
Changing only this attribute can cause to "update" in the next apply

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: completed the code refactoring, add the equal func for adminCredential

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
marcolan018: fixed

nirarg:
/approve
/retest
@bardielle can you have another review here?

provider/clusterrosaclassic/cluster_rosa_classic_resource.go
```diff
@@ -1039,6 +1078,12 @@ func validateNoImmutableAttChange(state, plan *ClusterRosaClassicState) diag.Dia
 	common.ValidateStateAndPlanEquals(state.MultiAZ, plan.MultiAZ, "multi_az", &diags)
 	common.ValidateStateAndPlanEquals(state.WorkerDiskSize, plan.WorkerDiskSize, "worker_disk_size", &diags)
 
+	// cluster admin attributes
+	common.ValidateStateAndPlanEquals(state.CreateAdminUser, plan.CreateAdminUser, "create_admin_user", &diags)
+	if !adminCredentialsEqual(state.AdminCredentials, plan.AdminCredentials) {
```
bardielle, line 1083: I think you can use the ValidateStateAndPlanEquals for that since it verified `!stateAttr.Equal(planAttr) ` please check it and remove it if it already covered....

provider/clusterrosaclassic/cluster_rosa_classic_state.go
```diff
@@ -99,3 +107,48 @@ type PrivateHostedZone struct {
...
+	}
+
+	return conf.Username.ValueString(), conf.Password.ValueString()
+}
+
+func adminCredentialsNull() types.Object {
+	return flattenAdminCredentials("", "")
+}
+
+func adminCredentialsEqual(state, plan types.Object) bool {
```
bardielle, line 146: The same comment here - 
I think you can use the ValidateStateAndPlanEquals for that since it verified `!stateAttr.Equal(planAttr) ` please check it and remove it if it already covered....

marcolan018: @bardielle thanks for the comments
per my understanding, `ValidateStateAndPlanEquals` is not useful for the complex object so it's not used in existing code, but use reflect.DeepEqual() like https://github.com/terraform-redhat/terraform-provider-rhcs/blob/8b3e60f7fee4608e925b3c87a4c631d92d6461dd/provider/clusterrosaclassic/cluster_rosa_classic_resource.go#L1024-L1029

Also, when comparing AdminCredential, we need to handle the case that state is nil&plan is unknown, which should not be recognized as invalid. That's why I add the `adminCredentialsEqual` func
let me know if you have any concern on this. Thanks

nirarg: > Also, when comparing AdminCredential, we need to handle the case that state is nil&plan is unknown

This is not true any more, as you use the UseStateIfUnknown plan modifier, no?
So if the state is null and the plan is unknown, this function should already be called with plan=null

bardielle: Did you test the `ValidateStateAndPlanEquals` function with adminCredentials? I think it's worth to check.
This function has been called from the update function so it means the attribute will have a value

marcolan018: @bardielle @nirarg 
Thanks for the comments. I tried to use `ValidateStateAndPlanEquals` to compare admincredentials in state&plan, it works, I updated my commit, please help to re-review.

marcolan018: @bardielle @nirarg 
I tried `ValidateStateAndPlanEquals`, and it could not work.
my test case:
created a cluster w/o cluster admin user created, then changed the main.tf with updated rhcs_cluster_rosa_classic.properties, then I will run into the error when issue `terraform apply`
```
rhcs_cluster_rosa_classic.rosa_sts_cluster: Modifying... [id=29c3pc9j7258jj5i6roc0atbol1i7qmj]
╷
│ Error: Attribute value cannot be changed
│ 
│   with rhcs_cluster_rosa_classic.rosa_sts_cluster,
│   on main.tf line 50, in resource "rhcs_cluster_rosa_classic" "rosa_sts_cluster":
│   50: resource "rhcs_cluster_rosa_classic" "rosa_sts_cluster" {
│ 
│ Attribute admin_credentials, cannot be changed from <null> to <unknown>
```

Also, all test cases in subsystem test related with update flow failed.

bardielle: Thank you for validating it. I'm approving the PR

PR #518, opened 2024-01-30
OCM-4953 | fix: remove paths from SHA256SUM releases

enriquebelarte:
SHA256SUM file needs a column with each filename of the release.
We were adding the full path and it breaks the Hashicorp release sync.


nirarg:
/override pull-ci-terraform-redhat-terraform-provider-rhcs-main-rosa-sts-advanced-critical-high-presubmit

PR #439, opened 2023-12-15
Add ROSA HCP support

tphee:
### Added:
cluster_rosa_hcp_resource.go          # Create ROSA HCP cluster
cluster_rosa_hcp_state.go                # Create ROSA HCP cluster
cluster_data_source.go                      # Retrive cluster api_url and console_url
cluster_data_state.go                         # Retrive cluster api_url and console_url

### Modified:
cluster.go
cluster_waiter.go
ocm_policies_data_source.go
ocm_policies_state.go
rosa_operator_roles_data_source.go
rosa_operator_roles_state.go
provider.go

PR #261, opened 2023-09-06
Example to create ROSA HCP Cluster

tphee:
### Added:
cluster_rosa_hcp_resource.go          # Create ROSA HCP cluster
cluster_rosa_hcp_state.go                 # Create ROSA HCP cluster
cluster_hcp_waiter_resource.go       # Wait for cluster current_compute == desired_compute
cluster_hcp_waiter_state.go              # Wait for cluster current_compute == desired_compute
cluster_data_source.go                      # Retrive cluster api_url and console_url
cluster_data_state.go                         # Retrive cluster api_url and console_url
examples/create_rosa_hcp_cluster/ # Example to create ROSA HCP cluster

### Modified:
ocm_policies_data_source.go             # Support HCP policies
ocm_policies_state.go                          # Support HCP policies
rosa_operator_roles_data_source.go # Support HCP operator roles
rosa_operator_roles_state.go              # Support HCP operator roles
provider.go                                              # Support HCP

### Example to create ROSA HCP Cluster:
Prerequisites: You have created an AWS VPC with public and private subnets.
```
make install
cd examples/create_rosa_hcp_cluster/hcp_cluster_with_managed_oidc_config
export TF_VAR_account_role_prefix=<account_role_prefix>
export TF_VAR_operator_role_prefix=<operator_role_prefix>
export TF_VAR_availability_zones='["<aws_az_region>"]'
export TF_VAR_aws_subnet_ids='["<aws_public_subnetid>","<aws_private_subnetid>"]'
export TF_VAR_cloud_region=<aws_region_name>
export TF_VAR_token=<ocm_offline_token>
terraform init
terraform apply
```

### Example terraform apply output:
```
Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

module.create_account_roles.module.rosa_account_roles[0].random_string.default_random: Creating...
random_password.password: Creating...
module.create_account_roles.module.rosa_account_roles[0].random_string.default_random: Creation complete after 0s [id=k7xv]
random_password.password: Creation complete after 0s [id=none]
module.create_account_roles.module.rosa_account_roles[0].module.rosa_operator_role_policies[6].aws_iam_policy.operator-policy: Creating...
module.create_account_roles.module.rosa_account_roles[0].module.rosa_operator_role_policies[0].aws_iam_policy.operator-policy: Creating...
[...]

PR #522, opened 2024-02-01
OCM-5758 | feat: HCP resources

gdbranco:
Related issue: [OCM-5758](https://issues.redhat.com//browse/OCM-5758)

Includes resources to managed HCP clusters
- [x] HCP Cluster (rhcs_cluster_rosa_hcp)
  - [x] Upgrade (user interacts with version field)
- [x] Autoscaler (rhcs_cluster_autoscaler)
- [x] Ingress (rhcs_hcp_default_ingress)
- [x] Tuning Configs (rhcs_tuning_config)
- [x] Machine Pool (rhcs_hcp_machine_pool)
  - [x] Upgrade (user interacts with version field)
- [x] Operator policies (rhcs_hcp_policies) 
  - Data source only

Modules will be available under it's [own repo](https://github.com/terraform-redhat/terraform-rhcs-rosa-hcp) . PR for it requires this to be merged and a new release for the provider

Included tests
- HCP Cluster
  - Subsystem
    - [x] Create
    - [x] Update
    - [x] Delete
    - [x] Upgrade
    - [x] Importing
- Autoscaler
  - Subsystem
    - [x] Create
    - [x] Update
    - [x] Delete
    - [x] Import
- Default Ingress
  - Subsytem
    - [x] Import
    - [x] Update
- Machine Pool
  - Subsystem
    - [x] Static validation
    - [x] Create
    - [x] Standard worker machine pool
    - [x] Import
    - [x] Delete
    - [x] Update
    - [x] Upgrade
- Tuning Configs
  - Subsystem
    - [x] Create
    - [x] Update
    - [x] Delete
    - [x] Import

**Change type**
- [x] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [x] Documentation
- [ ] Performance
- [x] Refactor
- [ ] Style
- [x] Unit tests
- [x] Subsystem tests

**Checklist**


nathanweatherly:
Thank you to [tphee](https://github.com/tphee) for the initial PRs (#261 & #439) that first explored ROSA HCP support! Great work!

internal/ocm/resource/cluster.go
```diff
@@ -71,16 +75,22 @@ func (c *Cluster) CreateNodes(autoScalingEnabled bool, replicas *int64, minRepli
...
 			minReplicasVal = int(*minReplicas)
 		}
-		if err := validations.MinReplicasValidator(minReplicasVal, multiAZ, false, 0); err != nil {
-			return err
+		// TODO need to identify the private subnet or remove this validation from TF
+		if clusterTopology == rosa.Classic {
+			if err := validations.MinReplicasValidator(minReplicasVal, multiAZ, clusterTopology == rosa.Hcp, 0); err != nil {
+				return err
+			}
 		}
```
ziccardi, line 83: The condition `clusterTopology == rosa.Hcp` will always be false because we enter the `if` statement only if `clusterTopology == rosa.Classic`. Is this intended?

internal/ocm/resource/cluster.go
```diff
@@ -95,8 +105,11 @@ func (c *Cluster) CreateNodes(autoScalingEnabled bool, replicas *int64, minRepli
...
 			replicasVal = int(*replicas)
 		}
-		if err := validations.MinReplicasValidator(replicasVal, multiAZ, false, 0); err != nil {
-			return err
+		// TODO need to identify the private subnet or remove this validation from TF
+		if clusterTopology == rosa.Classic {
+			if err := validations.MinReplicasValidator(replicasVal, multiAZ, clusterTopology == rosa.Hcp, 1); err != nil {
+				return err
+			}
 		}
```
ziccardi, line 113: see previous comment

internal/ocm/resource/cluster.go
```diff
@@ -125,11 +144,13 @@ func (c *Cluster) CreateAWSBuilder(awsTags map[string]string, ec2MetadataHttpTok
...
-	ec2MetadataHttpTokensVal := cmv1.Ec2MetadataHttpTokensOptional
-	if ec2MetadataHttpTokens != nil {
-		ec2MetadataHttpTokensVal = cmv1.Ec2MetadataHttpTokens(*ec2MetadataHttpTokens)
+	if clusterTopology == rosa.Classic {
+		ec2MetadataHttpTokensVal := cmv1.Ec2MetadataHttpTokensOptional
+		if ec2MetadataHttpTokens != nil {
+			ec2MetadataHttpTokensVal = cmv1.Ec2MetadataHttpTokens(*ec2MetadataHttpTokens)
+		}
+		awsBuilder.Ec2MetadataHttpTokens(ec2MetadataHttpTokensVal)
 	}
```
ziccardi, line 153:
```suggestion
if clusterTopology == rosa.Classic {
		if ec2MetadataHttpTokens != nil {
			awsBuilder.Ec2MetadataHttpTokens(cmv1.Ec2MetadataHttpTokens(*ec2MetadataHttpTokens))
		} else {
			awsBuilder.Ec2MetadataHttpTokens(cmv1.Ec2MetadataHttpTokensOptional)
		}
	}
```

provider/clusterrosa/hcp/datasource.go
```diff
@@ -0,0 +1,284 @@
...
+	state := &ClusterRosaHcpState{}
+	diags := request.Config.Get(ctx, state)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	// Find the cluster:
+	get, err := r.clusterCollection.Cluster(state.ID.ValueString()).Get().SendContext(ctx)
+	if err != nil && get.Status() == http.StatusNotFound {
```
ziccardi, line 240: I think this can be simplified to:
```suggestion
	if get.Status() == http.StatusNotFound {
```

provider/clusterrosa/hcp/datasource.go
```diff
@@ -0,0 +1,284 @@
...
+		tflog.Warn(ctx, fmt.Sprintf("cluster (%s) not found, removing from state",
+			state.ID.ValueString(),
+		))
+		response.State.RemoveResource(ctx)
+		return
+	} else if err != nil {
```
ziccardi, line 246: No need for an `else if`
```suggestion
	}
	
	if err != nil {
```

provider/clusterrosa/hcp/resource_test.go
```diff
@@ -0,0 +1,307 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Cluster Rosa Resource Suite")
+}
+
+var _ = Describe("Rosa HCP Sts cluster", func() {
+	Context("createHcpClusterObject", func() {
+		It("Creates a cluster with correct field values", func() {
+			clusterState := generateBasicRosaHcpClusterState()
+			rosaClusterObject, err := createHcpClusterObject(context.Background(), clusterState, diag.Diagnostics{})
+			Expect(err).To(BeNil())
```
ziccardi, line 177:
```suggestion
			Expect(err).NotTo(HaveOccurred())
```

provider/clusterrosa/hcp/resource_test.go
```diff
@@ -0,0 +1,307 @@
...
+			channel, ok := rosaClusterObject.Version().GetChannelGroup()
+			Expect(ok).To(BeTrue())
+			Expect(channel).To(Equal("stable"))
+		})
+	})
+	It("Throws an error when version format is invalid", func() {
+		clusterState := generateBasicRosaHcpClusterState()
+		clusterState.Version = types.StringValue("a.4.1")
+		_, err := createHcpClusterObject(context.Background(), clusterState, diag.Diagnostics{})
+		Expect(err).ToNot(BeNil())
```
ziccardi, line 210:
```suggestion
		Expect(err).To(HaveOccurred())
```

provider/clusterrosa/hcp/resource_test.go
```diff
@@ -0,0 +1,307 @@
...
+	})
+
+	It("Throws an error when version is unsupported", func() {
+		clusterState := generateBasicRosaHcpClusterState()
+		clusterState.Version = types.StringValue("4.1.0")
+		_, err := createHcpClusterObject(context.Background(), clusterState, diag.Diagnostics{})
+		Expect(err).ToNot(BeNil())
```
ziccardi, line 217:
```suggestion
		Expect(err).To(HaveOccurred())
```

provider/clusterrosa/hcp/resource_test.go
```diff
@@ -0,0 +1,307 @@
...
+	})
+
+	It("appends the non-default channel name to the requested version", func() {
+		clusterState := generateBasicRosaHcpClusterState()
+		clusterState.ChannelGroup = types.StringValue("somechannel")
+		rosaClusterObject, err := createHcpClusterObject(context.Background(), clusterState, diag.Diagnostics{})
+		Expect(err).To(BeNil())
```
ziccardi, line 224:
```suggestion
		Expect(err).ToNot(HaveOccurred())
```

provider/clusterrosa/hcp/resource_test.go
```diff
@@ -0,0 +1,307 @@
...
+		Expect(ok).To(BeTrue())
+		Expect(channel).To(Equal("somechannel"))
+	})
+
+	Context("populateRosaHcpClusterState", func() {
+		It("Converts correctly a Cluster object into a ClusterRosaHcpState", func() {
+			clusterState := &ClusterRosaHcpState{}
+			clusterJson := generateBasicRosaHcpClusterJson()
+			clusterJsonString, err := json.Marshal(clusterJson)
+			Expect(err).To(BeNil())
```
ziccardi, line 239:
```suggestion
			Expect(err).ToNot(HaveOccurred())
```

provider/clusterrosa/hcp/upgrade/upgrade.go
```diff
@@ -0,0 +1,236 @@
...
+			return false, fmt.Errorf("failed to parse upgrade version: %v", err)
+		}
+		switch upgrade.PolicyState.Value() {
+		case cmv1.UpgradePolicyStateValueDelayed, cmv1.UpgradePolicyStateValueStarted:
+			tflog.Debug(ctx, fmt.Sprintf("---- %v", desiredVersion.Equal(toVersion)))
+			if desiredVersion.Equal(toVersion) {
+				correctUpgradePending = true
+			} else {
+				return false, fmt.Errorf("a cluster upgrade is already in progress")
+			}
```
ziccardi, line 133: Early exit is more readable
```suggestion
			if !desiredVersion.Equal(toVersion) {
				return false, fmt.Errorf("a cluster upgrade is already in progress")
			} 	
			correctUpgradePending = true
```

provider/clusterrosa/hcp/upgrade/upgrade.go
```diff
@@ -0,0 +1,236 @@
...
+	return gates, str, nil
+}
+
+func getMissingGateAgreements(
+	upgradePolicy *cmv1.ControlPlaneUpgradePolicy,
+	upgradePoliciesClient *cmv1.ControlPlaneUpgradePoliciesClient) ([]*cmv1.VersionGate, error) {
+	response, err := upgradePoliciesClient.Add().Parameter("dryRun", true).Body(upgradePolicy).Send()
+
+	if err != nil {
+		if response.Error() != nil {
```
ziccardi, line 213:
```suggestion
	if err != nil && response.Error() != nil {
```

provider/defaultingress/hcp/resource.go
```diff
@@ -0,0 +1,330 @@
...
+func getDefaultIngressBuilder(ctx context.Context, state, plan *DefaultIngress) *cmv1.IngressBuilder {
+	ingressBuilder := cmv1.NewIngress()
+	if !common.IsStringAttributeUnknownOrEmpty(plan.ListeningMethod) && state.ListeningMethod != plan.ListeningMethod {
+		ingressBuilder.Listening(cmv1.ListeningMethod(plan.ListeningMethod.ValueString()))
+	}
+	return ingressBuilder
+}
+
+func validateDefaultIngress(ctx context.Context, state *DefaultIngress) error {
+	return nil
```
ziccardi, line 329: Is this intended?

provider/machinepool/hcp/upgrade/upgrade.go
```diff
@@ -0,0 +1,248 @@
...
+		if err != nil {
+			return false, fmt.Errorf("failed to parse upgrade version: %v", err)
+		}
+		switch upgrade.PolicyState.Value() {
+		case cmv1.UpgradePolicyStateValueDelayed, cmv1.UpgradePolicyStateValueStarted:
+			if desiredVersion.Equal(toVersion) {
+				correctUpgradePending = true
+			} else {
+				return false, fmt.Errorf("a cluster upgrade is already in progress")
+			}
```
ziccardi, line 146:
```suggestion
			if !desiredVersion.Equal(toVersion) {
				return false, fmt.Errorf("a cluster upgrade is already in progress")
			}
			correctUpgradePending = true
```

provider/machinepool/hcp/upgrade/upgrade.go
```diff
@@ -0,0 +1,248 @@
...
+	return gates, str, nil
+}
+
+func getMissingGateAgreements(
+	upgradePolicy *cmv1.NodePoolUpgradePolicy,
+	upgradePoliciesClient *cmv1.NodePoolUpgradePoliciesClient) ([]*cmv1.VersionGate, error) {
+	response, err := upgradePoliciesClient.Add().Parameter("dryRun", true).Body(upgradePolicy).Send()
+
+	if err != nil {
+		if response.Error() != nil {
```
ziccardi, line 225:
```suggestion
	if err != nil && response.Error() != nil {
```

internal/ocm/resource/cluster.go
gdbranco: Yes, this is intended as the validation for HCP is different. I'm currently leaving the validation for the API which I find more guaranteed anyways

internal/ocm/resource/cluster.go
gdbranco: I guess this is more of a preference, I find picking the default first and not having the else clearer :laughing:

gdbranco: I can also just go for the builder directly with the Optional default and remove the variable anyways, so that would be better and we still don't have the else. Thanks for shedding some light :+1:

provider/defaultingress/hcp/resource.go
gdbranco: Yes, for now. The validation for classic is around attributes that HCP doesn't have. I can just remove it as it's a no op, but it's mostly for someone in the future to implement...

provider/clusterrosa/hcp/datasource.go
gdbranco: hmm, I agree it could be like that. I don't know if there's a reason for it as it's the pattern followed for all not found errors at the moment

internal/ocm/resource/cluster.go
ziccardi: ok, but then I believe it would be more readable to pass directly `false`:
```suggestion
		if clusterTopology == rosa.Classic {
			if err := validations.MinReplicasValidator(minReplicasVal, multiAZ, false, 0); err != nil {
				return err
			}
		}
```
WDYT?

provider/defaultingress/hcp/resource.go
ziccardi: Ok. I suggest adding a comment explaining what you just said here then

provider/clusterrosa/hcp/datasource.go
ziccardi: Yeah. Leaving it is harmless although useless (unless I'm missing something).

nirarg (commented):
This is the beginning of the review
Totally this is looks great

Lets try to having common functions as much as possible instead of duplicating the code in both classic and hcp. Any function that not use the state can be used as common.

I think `provider/clusterrosa/rosa` should be renamed to `provider/clusterrosa/common`

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+			},
+			"etcd_encryption": schema.BoolAttribute{
+				Description: "Encrypt etcd data. Note that all AWS storage is already encrypted. " + common.ValueCannotBeChangedStringDescription,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.Bool{
+					boolplanmodifier.UseStateForUnknown(),
+				},
+			},
+			// "autoscaling_enabled": schema.BoolAttribute{
```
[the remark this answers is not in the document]
gdbranco: I removed it because the autoscaling is within the node pool resource this time around. Let me remove it entirely

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+			},
+			"console_url": schema.StringAttribute{
+				Description: "URL of the console.",
+				Computed:    true,
+			},
+			"domain": schema.StringAttribute{
+				Description: "DNS domain of cluster.",
+				Computed:    true,
+			},
+			"replicas": schema.Int64Attribute{
```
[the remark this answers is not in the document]
gdbranco: I had thought about this before, just started out following classic as it wasn't nested there indeed. But yeah, we can do that

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+			},
+			"host_prefix": schema.Int64Attribute{
+				Description: "Length of the prefix of the subnet assigned to each node. " + common.ValueCannotBeChangedStringDescription,
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+			"channel_group": schema.StringAttribute{
```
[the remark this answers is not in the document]
gdbranco: Some customers do use channel groups other than stable if I'm not mistaken, there aren't many but it is possible

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := ocmClusterResource.CreateAWSBuilder(rosa.Hcp, awsTags, nil, kmsKeyARN,
+		isPrivateLink, awsAccountID, awsBillingAccountId, stsBuilder, awsSubnetIDs, nil, nil,
+		nil, nil, nil); err != nil {
+		return nil, err
+	}
+
+	if err := ocmClusterResource.SetAPIPrivacy(isPrivateLink, isPrivateLink, stsBuilder != nil); err != nil {
```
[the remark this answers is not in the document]
gdbranco: It's not that it's not relevant, but STS clusters in general always need to be using AWS private link service for a private cluster. So I'd say it is easier to only ask for it :thinking:

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+
+const (
+	defaultTimeoutInMinutes   = int64(60)
+	nonPositiveTimeoutSummary = "Can't poll cluster state with a non-positive timeout"
+	nonPositiveTimeoutFormat  = "Can't poll state of cluster with identifier '%s', the timeout that was set is not a positive number"
+	pollingIntervalInMinutes  = 2
+
+	awsCloudProvider      = "aws"
+	rosaProduct           = "rosa"
+	MinVersion            = "4.10.0"
```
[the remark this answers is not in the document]
gdbranco: Definitely agree, let's add a comment about it

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+			avail = append(avail, v.RawID())
+		}
+		return fmt.Errorf("desired version (%s) is not in the list of available upgrades (%v)", desiredVersion, avail)
+	}
+
+	return nil
+}
+
+// Ensure user has acked upgrade gates and schedule the upgrade
+func scheduleUpgrade(ctx context.Context, client *cmv1.ClustersClient, clusterID string, desiredVersion *semver.Version, userAckString string) error {
```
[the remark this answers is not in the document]
gdbranco: Not quite, the internals are different for classic and hcp/node pool due to the endpoint. We could setup a template to be a bit more unified though

provider/clusterrosa/hcp/resource.go
```diff
@@ -0,0 +1,1455 @@
...
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: "Name of the cluster. Cannot exceed 15 characters in length. " + common.ValueCannotBeChangedStringDescription,
+				Required:    true,
+				Validators: []validator.String{
+					stringvalidator.LengthAtMost(15),
+				},
+			},
+			"cloud_region": schema.StringAttribute{
```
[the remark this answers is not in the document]
gdbranco: Because [this](https://github.com/terraform-redhat/terraform-provider-rhcs/blob/02cc52087a080963975c18dc82e7041aa6d5cfc2/provider/clusterrosa/rosa/validators.go#L20) validator is shared, I'd say we should stick to `cloud_region` or we would need to duplicate the functionality

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
gdbranco: I'd rather have the control for this directly on the node pool resource instead of the cluster level. Trying to move away from managing it through the cluster

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
gdbranco: Decided to expose it as private only, but will be keeping supplying both with the same value. The reason is that for STS in general, which includes HCP, you cannot have a private cluster without private link. This is an option only for classic non sts

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
gdbranco: I'd need to think a bit more on this. The flow overall is the same, but the builder and the client are different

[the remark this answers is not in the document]
gdbranco: I'll open a ticket specifically for this one

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
gdbranco: Open a ticket for this one

gdbranco:
Closing to open against main branch

PR #537, opened 2024-02-11
OCM-5399 | fix: Remove duplicated wait function

nirarg:
This commit also fixes [OCM-5753](https://issues.redhat.com//browse/OCM-5753):
* In case of timeout and state not "READY" return error

**Which issue(s) this PR fixes**:
Fixes [OCM-5753](https://issues.redhat.com//browse/OCM-5753), [OCM-5399](https://issues.redhat.com//browse/OCM-5399)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [x] Refactor
- [ ] Style
- [ ] Unit tests
- [x] Subsystem tests

**Checklist**

### DIFF
--- a/subsystem/cluster_waiter_test.go
+++ b/subsystem/cluster_waiter_test.go
@@ -141,10 +141,6 @@ var _ = Describe("Cluster creation", func() {
 					VerifyRequest(http.MethodGet, "/api/clusters_mgmt/v1/clusters/123"),
 					RespondWithJSON(http.StatusOK, templateReadyState),
 				),
-				CombineHandlers(
-					VerifyRequest(http.MethodGet, "/api/clusters_mgmt/v1/clusters/123"),
-					RespondWithJSON(http.StatusOK, templateReadyState),
-				),
 			)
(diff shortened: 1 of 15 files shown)


subsystem/cluster_waiter_test.go
```diff
@@ -175,7 +175,7 @@ var _ = Describe("Cluster creation", func() {
 				}
 			`)
 
-		Expect(terraform.Apply()).To(BeZero())
+		Expect(terraform.Apply()).NotTo(BeZero())
 		resource := terraform.Resource("rhcs_cluster_wait", "rosa_cluster")
```
bardielle, line 179: Does it possible to add a test that cluster is still installing and the waiter will be failed?

nirarg: This is the test which reproduce the issue
The cluster wait is finished with timeout and state="waiting"
This is why added the change to expect result not to be zero

PR #544, opened 2024-02-14
OCM-6092 | fix: add "create_admin_user" in cluster datasource

marcolan018:

Fixes #[OCM-6092](https://issues.redhat.com/browse/OCM-6092)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**

### DIFF
--- a/docs/data-sources/cluster_rosa_classic.md
+++ b/docs/data-sources/cluster_rosa_classic.md
@@ -31,9 +31,9 @@ data "rhcs_cluster_rosa_classic" "cluster" {
 
 ### Read-Only
 
-- `admin_credentials` (Attributes) This attribute is not support for cluster data source. Therefore, it will not be displayed as an output of the datasource (see [below for nested schema](#nestedatt--admin_credentials))
+- `admin_credentials` (Attributes) This attribute is not supported for cluster data source. Therefore, it will not be displayed as an output of the datasource (see [below for nested schema](#nestedatt--admin_credentials))
 - `api_url` (String) URL of the API server.
-- `autoscaling_enabled` (Boolean) This attribute is not support for cluster data source. Therefore, it will not be displayed as an output of the datasource
+- `autoscaling_enabled` (Boolean) This attribute is not supported for cluster data source. Therefore, it will not be displayed as an output of the datasource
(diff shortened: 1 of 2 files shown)


docs/data-sources/cluster_rosa_classic.md
```diff
@@ -47,6 +47,7 @@ data "rhcs_cluster_rosa_classic" "cluster" {
 - `cloud_region` (String) Cloud region identifier, for example 'us-east-1'.
 - `compute_machine_type` (String) This attribute is not support for cluster data source. Therefore, it will not be displayed as an output of the datasource
 - `console_url` (String) URL of the console.
+- `create_admin_user` (Boolean) This attribute is not support for cluster data source. Therefore, it will not be displayed as an output of the datasource
```
gdbranco, line 50:
```suggestion
- `create_admin_user` (Boolean) This attribute is not supported for cluster data source. Therefore, it will not be displayed as an output of the datasource
```

gdbranco: Considering it's not supported do we even need to add it? Maybe we could have a state for the resource and one for the data source

marcolan018: fixed the message.
For the dedicate state of data source, if we decide to use it. I prefer to make that change in another pr. Current pr only focusing on fix the bug.
@nirarg comments?

nirarg: > Considering it's not supported do we even need to add it? Maybe we could have a state for the resource and one for the data source

The resource and data source are using the same state object, therefore all attributes must be added
There are two issues in terraform that enforce us to add these variables:
* The state object can't use Embedded struct (hasicorp issue: [242](https://github.com/hashicorp/terraform-plugin-framework/issues/242))
* We didn't find a way to skip specific attributes when generating the docs (if you want to have the attribute but not published in the registry documentations)

I agree that we need to find a way to fix this, but this shouldn't block this PR

/cc @bardielle

ciaranRoche (commented):
/lgtm

PR #553, opened 2024-02-26
OCM-6344 | feat: add a waiter for std compute nodes

gdbranco:
[OCM-6344](https://issues.redhat.com//browse/OCM-6344)

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [x] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [x] Subsystem tests

**Checklist**

### DIFF
--- a/provider/clusterrosa/hcp/resource.go
+++ b/provider/clusterrosa/hcp/resource.go
@@ -279,6 +279,10 @@ func (r *ClusterRosaHcpResource) Schema(ctx context.Context, req resource.Schema
 				Description: "Wait until the cluster is either in a ready state or in an error state. The waiter has a timeout of 60 minutes, with the default value set to false",
 				Optional:    true,
 			},
+			"wait_for_std_compute_nodes_complete": schema.BoolAttribute{
+				Description: "Wait until the cluster standard compute pools are created. The waiter has a timeout of 60 minutes, with the default value set to false",
+				Optional:    true,
+			},
 		},
 	}
 }
@@ -510,6 +514,16 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 	}
 	summary := "Can't build cluster"
 
+	shouldWaitCreationComplete := common.BoolWithFalseDefault(state.WaitForCreateComplete)
+	shouldWaitComputeNodesComplete := common.BoolWithFalseDefault(state.WaitForStdComputeNodesComplete)
+	if shouldWaitComputeNodesComplete && !shouldWaitCreationComplete {
+		response.Diagnostics.AddError(
+			summary,
+			"When waiting for standard compute nodes to complete it is also required to wait for creation of the cluster",
+		)
+		return
+	}
+
 	// In case version with "openshift-v" prefix was used here,
 	// Give a meaningful message to inform the user that it not supported any more
 	if common.HasValue(state.Version) && strings.HasPrefix(state.Version.ValueString(), rosa.VersionPrefix) {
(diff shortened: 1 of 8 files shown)


provider/clusterrosa/hcp/resource.go
```diff
@@ -578,6 +583,21 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		}
 	}
 
+	if common.HasValue(state.WaitForStdComputeNodesComplete) &&
+		state.WaitForStdComputeNodesComplete.ValueBool() {
```
nirarg, line 587:
```suggestion
	if common.BoolWithFalseDefault(state.WaitForStdComputeNodesComplete) {
```

provider/clusterrosa/hcp/resource.go
```diff
@@ -578,6 +583,21 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+		tflog.Info(ctx, "Waiting for standard compute nodes to get ready")
+		object, err = r.ClusterWait.WaitForStdComputeNodesToBeReady(ctx, object.ID(), rosa.DefaultWaitTimeoutInMinutes)
+		if err != nil {
+			response.Diagnostics.AddError(
+				"Waiting for cluster creation finished with error",
```
nirarg, line 592:
```suggestion
				"Waiting for std compute nodes completion finished with error",
```

provider/clusterrosa/hcp/resource.go
```diff
@@ -578,6 +583,21 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+				fmt.Sprintf("Waiting for cluster creation finished with the error %v", err),
```
nirarg, line 593:
```suggestion
				fmt.Sprintf("Waiting for std compute nodes completion finished with the error %v", err),
```

provider/clusterrosa/hcp/resource.go
```diff
@@ -578,6 +583,21 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+			)
+			if object == nil {
+				return
+			}
```
nirarg, line 597: Is it possible that err is returned and object is not nil?

provider/clusterrosa/hcp/resource.go
```diff
@@ -578,6 +583,21 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		}
 	}
 
+	if common.HasValue(state.WaitForStdComputeNodesComplete) &&
```
nirarg, line 586: What happen if user set `state.WaitForCreateComplete` to true? do we need to enforce waiting for cluster ready first?

gdbranco: It waits first because it comes first in the section above

provider/clusterrosa/hcp/resource.go
gdbranco: There are instances where even though it errors it expects to populate the object, similarly to error state in cluster waiter

provider/clusterrosa/hcp/resource.go
gdbranco: I could see an issue, let's say the user signals to wait only for the compute nodes and the cluster goes to error. It would timeout but it might no be clear why

I guess we could enforce that for this to be enabled the wait for cluster also needs to be

nirarg: Yes, I agree with this approach, thanks

ciaranRoche (commented):
/lgtm

PR #560, opened 2024-02-29
OCM-6438 | fix: add etcd kms key arn support to hcp clusters

gdbranco:
[OCM-6438](https://issues.redhat.com//browse/OCM-6438)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [x] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [x] Unit tests
- [x] Subsystem tests

**Checklist**


internal/ocm/resource/cluster_test.go
```diff
@@ -220,7 +220,7 @@ var _ = Describe("Cluster", func() {
 		})
 		It("PrivateLink false invalid Ec2MetadataHttpTokens - success", func() {
 			// TODO Need to add validation for Ec2MetadataHttpTokens
-			err := cluster.CreateAWSBuilder(rosaTypes.Classic, nil, pointer("test"), nil, false, nil, nil, nil, nil, nil, nil, nil, nil, nil)
+			err := cluster.CreateAWSBuilder(rosaTypes.Classic, nil, pointer("test"), nil, nil, false, nil, nil, nil, nil, nil, nil, nil, nil, nil)
```
nirarg, line 223: I think there should be added more tests for hcp rosa type here.
For the new "etcdKmsKeyArn" param please add:
1. success test with etcdKmsKeyArn and type hcp
2. Failure test with etcdKmsKeyArn and type classic

I think this can be completed in another PR so we can unblock this PR

nirarg (commented):
One small comment about unit tests
But lets merge this PR and take care of the comment in another PR
/lgtm
/approve

internal/ocm/resource/cluster_test.go
gdbranco: I think for 1 this is what you are looking for
https://github.com/terraform-redhat/terraform-provider-rhcs/pull/560/files#diff-a2cd4e1e3866ba9a8adf53d4a36f54f0fc1bfc927da62012579a05641cb1d34bR921

There can't be a subsystem test for classic though as it doesn't have the etcd kms key arn attribute, so there's no way to set it anyway. The validation is in place in case of miss use of the internal function that populates the fields of a cluster, since the api spec used is the same

PR #574, opened 2024-03-15
add validation for cluster admin user name to avoid empty string

marcolan018:

changes in https://issues.redhat.com/browse/OCM-5632 led to errors in e2e test. 
username in admin_credentials is no longer mandatory, we need to add validator to guarantee it's not empty string.

Fixes #OCM-6709(https://issues.redhat.com/browse/OCM-6709)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


provider/identityprovider/htpasswd.go
```diff
@@ -37,6 +37,7 @@ var (
 
 	HTPasswdUsernameValidators = []validator.String{
 		stringvalidator.RegexMatches(regexp.MustCompile(`^[^/:%]*$`), "username may not contain the characters: '/:%'"),
+		stringvalidator.RegexMatches(regexp.MustCompile(`.*\S.*`), "username may not be empty string"),
```
nirarg, line 40: I think that checking the length here may make it more clear (make it simple):
```suggestion
		stringvalidator.LengthAtLeast(<minimum length>),
```

marcolan018: @nirarg
there is no min length for user name. 
The regex .*\S.* is used to reject any string which only contains space, such as "", or "   ".  Maybe just needs to change the message to something like `username may not be empty/blank string`
comments?

nirarg: checking string is not empty is the same like checking length is bigger than 1, and I think it will be simpler for understanding.
Anyway, the solution suggested here looks good and its up to you if to change it.

/approve
@gdbranco please review this
Is this going to be included in v1.6.0?

tests/e2e/negative_day_one_test.go
```diff
@@ -59,7 +59,7 @@ var _ = Describe("RHCS Provider Negative Test", func() {
 					creationArgs.AdminCredentials["username"] = ""
 					err = clusterService.Apply(creationArgs, true, true)
 					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(ContainSubstring("Attribute 'username' is mandatory"))
+					Expect(err.Error()).To(ContainSubstring("Attribute admin_credentials.username username may not be empty/blank string"))
```
gdbranco, line 62:
```suggestion
					Expect(err.Error()).To(ContainSubstring("Attribute 'admin_credentials.username' username may not be empty/blank string"))
```

provider/identityprovider/htpasswd.go
```diff
@@ -37,6 +37,7 @@ var (
 
 	HTPasswdUsernameValidators = []validator.String{
 		stringvalidator.RegexMatches(regexp.MustCompile(`^[^/:%]*$`), "username may not contain the characters: '/:%'"),
+		stringvalidator.RegexMatches(regexp.MustCompile(`.*\S.*`), "username may not be empty/blank string"),
```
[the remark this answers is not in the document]
gdbranco: Oh but you also check for spaces, nvm

PR #597, opened 2024-03-26
OCM-7049 | fix: allow proxy to be reset

gdbranco:
[OCM-7049](https://issues.redhat.com//browse/OCM-7049)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [x] Subsystem tests

**Checklist**


provider/clusterrosa/classic/cluster_rosa_classic_resource.go
```diff
@@ -1542,29 +1541,32 @@ func populateRosaClassicClusterState(ctx context.Context, object *cmv1.Cluster,
 			state.Proxy = &proxy.Proxy{}
 		}
 		httpProxy, ok := proxyObj.GetHTTPProxy()
-		if ok {
+		if ok && httpProxy != "" {
```
nirarg, line 1544: Why using empty string for reset the value and not using null for this?
Before this change, what happen if user update the value from valid value to null?
`httpProxy = null`

gdbranco: Nothing happened, plus the API usually omits null. It currently expects empty strings to reset

provider/cluster/cluster_resource.go
```diff
@@ -203,7 +202,6 @@ func (r *ClusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 				Description: "proxy",
 				Attributes:  proxy.ProxyResource(),
 				Optional:    true,
-				Validators:  []validator.Object{proxy.ProxyValidator()},
```
nirarg, line 206: I suggest to add this to the description - empty string used to reset

PR #598, opened 2024-03-27
OCM-902 | fix:update the check for imdsv2 default value

sunzhaohua2:
@xueli181114


xueli181114:
@sunzhaohua2 The running log on local is expected.

PR #622, opened 2024-04-09
OCM-7347 | fix: allow to reset tuning configs

gdbranco:
[OCM-7347](https://issues.redhat.com//browse/OCM-7347)

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [x] Subsystem tests

**Checklist**

### DIFF
--- a/provider/machinepool/hcp/machine_pool_resource.go
+++ b/provider/machinepool/hcp/machine_pool_resource.go
@@ -779,6 +779,10 @@ func (r *HcpMachinePoolResource) doUpdate(ctx context.Context, state *HcpMachine
 		state.AWSNodePool.Tags = plan.AWSNodePool.Tags
 	}
 
+	if common.HasValue(plan.TuningConfigs) {
+		state.TuningConfigs = plan.TuningConfigs
+	}
+
 	// Save the state:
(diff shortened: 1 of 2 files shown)


provider/machinepool/hcp/machine_pool_resource.go
```diff
@@ -1186,7 +1190,9 @@ func populateState(ctx context.Context, object *cmv1.NodePool, state *HcpMachine
 		}
 		state.TuningConfigs = tuningConfigsList
 	} else {
-		state.TuningConfigs = types.ListNull(types.StringType)
+		if !common.HasValue(state.TuningConfigs) {
```
ciaranRoche, line 1193: Does this need to be wrapped in the else

gdbranco: Gotcha, I think we need similar refactors throughout then

ciaranRoche: not a blocker, just NP, I wouldnt roll with making more refactors, functionally its good as is

PR #626, opened 2024-04-18
OCM-6751 | fix: adjusts various issues

gdbranco:
OCM-7035 | fix: cluster should not be editable in autoscaler
OCM-7037 | fix: shouldn't edit STS fields in cluster
OCM-7348 | fix: name and subnet ID on machine pool shouldn't be empty
OCM-6805 | fix: set a regex to validate AWS acc ID
OCM-6593 | fix: allow to set etcd_encryption to false it no value is supplied to kms key arn
OCM-6805 | feat: forward billing account changes
OCM-6806 | fix: shouldn't allow to edit cloud region
OCM-7031 | fix: machine pool instance type shouldn't be editable
OCM-7029 | fix: machine pool subnet ID shouldn't be editable
OCM-6831 | fix: add validation to rosa_creator_arn property

**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [x] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [x] Unit tests
- [x] Subsystem tests

**Checklist**


provider/clusterrosa/hcp/resource.go
```diff
@@ -429,6 +431,24 @@ func createHcpClusterObject(ctx context.Context,
...
+		if !ok {
+			errDescription := fmt.Sprintf("Expected property '%s'. Please include it by supplying 'data.aws_caller_identity.current.arn'", rosa.PropertyRosaCreatorArn)
+			diags.AddError(
+				errHeadline,
+				errDescription,
+			)
+			return nil, errors.New(errHeadline + "\n" + errDescription)
+		}
+		if !rosa.UserArnRE.MatchString(creatorArn) {
+			errDescription := fmt.Sprintf("Property '%s' does not have a valid user arn. Please include it by supplying 'data.aws_caller_identity.current.arn'", rosa.PropertyRosaCreatorArn)
```
marcolan018, line 444: should we use another message here for the invalid user arn, not "Please include it by supplying 'data.aws_caller_identity.current.arn'"?

gdbranco: hmm, not sure what would be best. Given we are trying to have more actionable details I'd say this would be fitting. But if we want to keep ourselves a bit less opinionated we can also just say it's invalid

gdbranco: @marcolan018 Adjusted so that the aws data source is treated as an example instead. Let me know if this is better

nirarg:
Great improvements :-) 
/approve

PR #631, opened 2024-04-22
OCM-5748 | fix: Updated to use new RHCS module

jerichokeyne:
Replaces the use of the old `rosa-sts/aws` module with the new `rosa-classic/rhcs` module in the test code 

Fixes #[OCM-5748](https://issues.redhat.com//browse/OCM-5748)

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [x] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [x] Subsystem tests

**Checklist**

### DIFF
--- a/tests/tf-manifests/aws/add-account-roles/rosa-classic/main.tf
+++ b/tests/tf-manifests/aws/add-account-roles/rosa-classic/main.tf
@@ -21,11 +21,11 @@ data "aws_partition" "current" {
 }
 
 locals {
-  path = coalesce(var.path, "/")
-  major_version = "${split(".", var.openshift_version)[0]}.${split(".", var.openshift_version)[1]}"
-  versionfilter = var.openshift_version == null ? "" : " and raw_id like '%${local.major_version}%'"
+  path               = coalesce(var.path, "/")
+  major_version      = "${split(".", var.openshift_version)[0]}.${split(".", var.openshift_version)[1]}"
+  versionfilter      = var.openshift_version == null ? "" : " and raw_id like '%${local.major_version}%'"
   installer_role_arn = "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:role${local.path}${var.account_role_prefix}-Installer-Role"
-  aws_account_id = "${data.aws_caller_identity.current.account_id}"
+  aws_account_id     = data.aws_caller_identity.current.account_id
 }
 
 data "rhcs_policies" "all_policies" {}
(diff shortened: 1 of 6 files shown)


tests/tf-manifests/aws/add-account-roles/rosa-classic/main.tf
```diff
@@ -28,17 +28,10 @@ data "rhcs_versions" "all" {
 }
 
 module "create_account_roles" {
-  source  = "terraform-redhat/rosa-sts/aws"
-  version = ">= 0.0.14"
+  source  = "terraform-redhat/rosa-classic/rhcs//modules/account-iam-resources"
```
radtriste, line 31: do we need the `//` ?

jerichokeyne: yeah. that's telling terraform to use a sub-module of the rhcs module

radtriste:
Depends on https://github.com/openshift/release/pull/51447

radtriste (approved):
/lgtm

radtriste:
I think this method needs to be revisited if you want to add the Classic modules: https://github.com/terraform-redhat/terraform-provider-rhcs/blob/main/tests/utils/helper/manifests_handler.go#L69

jerichokeyne:
@radtriste Sorry for the delay, had to work on some other stuff and this took longer than expected to implement. Can you take a look at my latest changes? This seemed to be the simplest way to implement such functionality as updating the source and version info for the modules is a bit more complicated than for updating providers

tests/utils/constants/constants.go
```diff
@@ -54,6 +54,9 @@ const (
 	MajorVersion              = "MAJOR_VERSION_ENV"
 	RHCSVersion               = "RHCS_VERSION"
 	RHCSSource                = "RHCS_SOURCE"
+	ModuleVersion             = "MODULE_VERSION"
+	ModuleSource              = "MODULE_SOURCE"
+	ModuleSourceLocal         = "MODULE_SOURCE_LOCAL"
```
radtriste, line 59: Why not having `MODULE_SOURCE` only and let the user decide whether local or not ?
is there any difference ?

tests/utils/helper/manifests_handler.go
```diff
@@ -11,19 +11,42 @@ import (
 	. "github.com/terraform-redhat/terraform-provider-rhcs/tests/utils/log"
 )
 
+func MocduleSourceAlignment(manifestsFilePath string, source string, sourceLocal string, version string) error {
```
radtriste, line 14: typo on method name

tests/utils/constants/constants.go
jerichokeyne: Basically the way terraform determines if the `source` option for a module is local or remote is purely based on whether or not the `version` option is set. So I set it up so that if the `MODULE_SOURCE_LOCAL` environment variable is set at all, then it will remove the `version` option so that you can use a local path in `MODULE_SOURCE`

tests/utils/helper/manifests_handler.go
jerichokeyne: Woops. Fixed it

tests/utils/constants/constants.go
radtriste: Ok. Can you add a comment for its usage ?

jerichokeyne: Added comments for the three new environment variables

radtriste:
@jerichokeyne Please rebase and I will approve it tomorrow

radtriste (approved):

PR #639, opened 2024-04-26
OCM-7643 | ci: Modify the member type of ResourceRange struct

jameszwang:
[OCM-7643](https://issues.redhat.com//browse/OCM-7643) | ci: Modify the member type of ResourceRange struct


radtriste (approved):
/lgtm

PR #725, opened 2024-07-22
OCM-9161 | feat: support IMDSv2 for HCP cluster

marcolan018:

This PR updates cluster resource for HCP so that user can specify IMDSv2 related attributes when creating a HCP clusters like classic ones
Fixes #https://issues.redhat.com/browse/OCM-9161

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


provider/clusterrosa/hcp/resource.go
```diff
@@ -351,6 +353,17 @@ func (r *ClusterRosaHcpResource) Schema(ctx context.Context, req resource.Schema
 					objectplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"ec2_metadata_http_tokens": schema.StringAttribute{
+				Description: "This value determines which EC2 Instance Metadata Service mode to use for EC2 instances in the cluster." +
+					"This can be set as `optional` (IMDS v1 or v2) or `required` (IMDSv2 only)." + common.ValueCannotBeChangedStringDescription,
+				Optional: true,
```
gdbranco, line 359: Should we enforce it instead of having the optional and computed behavior?

subsystem/hcp/cluster_resource_test.go
```diff
@@ -825,12 +825,13 @@ var _ = Describe("HCP Cluster", func() {
...
 					VerifyJQ(`.region.id`, "us-west-1"),
 					VerifyJQ(`.product.id`, "rosa"),
-
+					VerifyJQ(`.aws.ec2_metadata_http_tokens`, "required"),
 					RespondWithPatchedJSON(http.StatusCreated, template, `[
 					{
 					  "op": "add",
 					  "path": "/aws",
 					  "value": {
+					  	  "ec2_metadata_http_tokens" : "required",
```
gdbranco, line 834: If we keep it as an optional value could you also include a test for when the value is not supplied that it is being set as optional?

provider/clusterrosa/hcp/resource.go
marcolan018: @gdbranco 
do you mean to change this to "Required: true"? We should not make this change IMO.
BTW, in classic cluster resource this attribute is also set as optional and computed.

gdbranco: Yeah, we have been trying to move away from optional and computed values. But it's alright. I suppose it's easier anyway. As I expect the API to start returning the value for this for clusters that were not creating setting it

subsystem/hcp/cluster_resource_test.go
marcolan018: @gdbranco 
added test case which no ec2_metadata_http_tokens set, please help to re-review the changes, thanks

PR #754, opened 2024-08-19
OCM-10200 | feat: Add OCM client to RHCS TF provider

hunterkepley:
[OCM-10200](https://issues.redhat.com//browse/OCM-10200) is a ticket about replacing the oidc config fetching from front-end to back-end to circumvent user proxies. Some customers are unable to fetch oidc config thumbprints with its current state. The changes have been made to ROSA already, this is the final step in the process/feature

Adds OCM Client for use with the thumbprint changes (moving oidc config thumbprint fetching to backend to circumvent proxies)

Fixes #

**Change type**
- [X] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**
- [X] Subject and description added to both, commit and PR.
- [X] Relevant issues have been referenced.


internal/ocm/client.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/rosa/pkg/config"
```
gdbranco, line 28: We should not re introduce rosa as a dependency to rhcs though, we need to move things into ocm-common or setup it up internally to rhcs itself if we believe it will eventually diverge in terms of functionality

internal/ocm/client.go
```diff
@@ -0,0 +1,175 @@
...
+		Build()
+	if err != nil {
+		return
+	}
+
+	// Prepare the builder for the connection adding only the properties that have explicit
+	// values in the configuration, so that default values won't be overridden:
+	builder := sdk.NewConnectionBuilder()
+	builder.Logger(logger)
+	builder.Agent(info.UserAgent + "/" + info.Version + " " + sdk.DefaultAgent)
```
gdbranco, line 122: The way this is setup from here it is picking up rosa information instead of rhcs

PR #817, opened 2024-10-11
OCM-11521 | fix: validate custom machinepool disk size

willkutler:
Adds validation for custom disk size for both day 1 & 2 classic and HCP machine pools created with TF. Previously these fields were not validated in the provider, so failures would only happen on the API side.

Fixes #[OCM-11521](https://issues.redhat.com//browse/OCM-11521)

**Change type**
- [ ] New feature
- [x ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [x ] Subsystem tests

**Checklist**
- [ x] Subject and description added to both, commit and PR.
- [ x] Relevant issues have been referenced.


provider/clusterrosa/hcp/resource.go
```diff
@@ -497,9 +497,10 @@ func createHcpClusterObject(ctx context.Context,
 	}
 
 	workerDiskSize := common.OptionalInt64(state.WorkerDiskSize)
+	openshiftVersion := common.OptionalString(state.Version)
 
 	if err := ocmClusterResource.CreateNodes(rosaTypes.Hcp, false, replicas, nil, nil,
-		computeMachineType, nil, availabilityZones, true, workerDiskSize); err != nil {
+		computeMachineType, nil, availabilityZones, true, workerDiskSize, openshiftVersion); err != nil {
```
marcolan018, line 503: can we pass a nil directly here since version is not required in hcp flow?

internal/ocm/resource/cluster.go
```diff
@@ -44,6 +45,20 @@ func (c *Cluster) CreateNodes(clusterTopology rosaTypes.ClusterTopology, autoSca
 	}
 
 	if workerDiskSize != nil {
+		if clusterTopology == rosaTypes.Hcp {
+			err := diskValidator.ValidateNodePoolRootDiskSize(int(*workerDiskSize))
+			if err != nil {
+				return err
+			}
+		} else if clusterTopology == rosaTypes.Classic {
```
marcolan018, line 53: is there the 3rd possible value for clusterTopology?
If not, don't need to check `if clusterTopology == rosaTypes.Classic`
if yes, needs to add an additional `else` block to handle it, maybe throw an error such as invalid topology

internal/ocm/resource/cluster_test.go
```diff
@@ -174,13 +174,14 @@ var _ = Describe("Cluster", func() {
...
 		It("Autoscaling disabled multiAZ true set one zone - failure", func() {
-			err := cluster.CreateNodes(rosaTypes.Classic, false, nil, nil, nil, nil, nil, []string{"us-east-1a", "us-east-1b", "us-east-1c"}, true, nil)
+			err := cluster.CreateNodes(rosaTypes.Classic, false, nil, nil, nil, nil, nil, []string{"us-east-1a", "us-east-1b", "us-east-1c"}, true, nil, nil)
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(Equal("Multi AZ cluster requires at least 3 compute nodes"))
 		})
 		It("Custom disk size", func() {
 			diskSize := int64(543)
-			err := cluster.CreateNodes(rosaTypes.Classic, false, pointer(int64(3)), nil, nil, nil, nil, nil, false, &diskSize)
+			version := "4.14.0"
```
gdbranco, line 183: Should it include a test with a negative flow version?

internal/ocm/resource/cluster.go
willkutler: no, it's just classic and HCP - I just thought it made sense to structure it like this in case we end up adding another one in the future. I can change it to be a regular if/else though

marcolan018:
lgtm

gdbranco:
Failure unrelated to the validation added, cluster provision is being slow. Merging

PR #837, opened 2024-11-04
OCM-12269 | ci: fix with reducing the number of sg in id:73068

yuwang-RH:
[OCM-12269](https://issues.redhat.com//browse/OCM-12269) | ci: fix with reducing the number of sg in id:73068

### DIFF
--- a/tests/e2e/hcp_machine_pool_test.go
+++ b/tests/e2e/hcp_machine_pool_test.go
@@ -263,7 +263,7 @@ var _ = Describe("HCP MachinePool", ci.Day2, ci.FeatureMachinepool, func() {
 				Replicas:                 helper.IntPointer(replicas),
 				MachineType:              helper.StringPointer(machineType),
 				Name:                     helper.StringPointer(name),
-				AdditionalSecurityGroups: helper.StringSlicePointer(output.SGIDs),
+				AdditionalSecurityGroups: helper.StringSlicePointer(sgIDs),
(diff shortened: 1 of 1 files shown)


yuwang-RH:
@xueli181114 @jerichokeyne Could you help review this PR?Thanks.

tests/utils/profilehandler/handler.go
```diff
@@ -1054,9 +1054,18 @@ func (ctx *profileContext) DestroyRHCSClusterResources(token string) error {
 		if err != nil {
 			errs = append(errs, err)
 		} else {
-			_, err = vpcService.Destroy()
+			_, err := vpcService.Destroy()
 			if err != nil {
-				errs = append(errs, err)
+				if strings.Contains(err.Error(), "DependencyViolation") {
+					Logger.Warnf("DependencyViolation, waiting for 15 minutes before retrying. The error is %v", err)
+					time.Sleep(15 * time.Minute)
```
xueli181114, line 1061: Maybe it's better to use for loop rather than sleeping 15 mins?

yuwang-RH: The time it takes to destroy vpc is uncertain; sometimes it's longer, sometimes shorter, it almost takes about 20min ,so here we only want to retry once with an interval based on experience.
Considering it is a sporadic issue, although the probability of it occurring is a bit high recently, I will add comment to declare the time.sleep is just a workaround, will be removed in future if the dependency can be deleted before the vpc destroy. What do you think about it?

PR #840, opened 2024-11-06
OCM-12333 | ci: fix id:65071 fix subnet leak

yuwang-RH:
OCM-12333 | ci: fix id:65071 fix subnet leak


tests/e2e/classic_machine_pool_test.go
```diff
@@ -444,7 +447,23 @@ var _ = Describe("Create MachinePool", ci.Day2, ci.FeatureMachinepool, func() {
 		Expect(err).ToNot(HaveOccurred())
 		defer func() {
 			By("Remove added subnet from VPC")
-			awsClient.DeleteSubnet(subnet.ID)
+			// subnet delettion should be all dependency of the subnet is removed
+			// In this cases, the machinepool is the subnet dependency, it needs some time to be deleted from aws
+			err = wait.PollUntilContextTimeout(
```
xueli181114, line 452: How about using Eventually rather than import another package?

yuwang-RH: Sure I can use Eventually.

PR #858, opened 2024-12-02
OCM-12057 | feat: Added adjustable custom wait times

den-rgb:
Adding custom wait time adjusters for cluster and hcp machinepool creation. 
**Which issue(s) this PR fixes** 
Fixes # [OCM-12057](https://issues.redhat.com/browse/OCM-12057)

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


provider/clusterrosa/classic/cluster_rosa_classic_resource.go
```diff
@@ -882,7 +886,9 @@ func (r *ClusterRosaClassicResource) Create(ctx context.Context, request resourc
 	}
 
 	if common.HasValue(state.WaitForCreateComplete) && state.WaitForCreateComplete.ValueBool() {
-		object, err = r.ClusterWait.WaitForClusterToBeReady(ctx, object.ID(), rosa.DefaultWaitTimeoutInMinutes)
+		timeOut := common.OptionalInt64(state.ClusterWaitTimeoutInMinutes)
+		timeOut = common.ValidateTimeout(timeOut, rosa.DefaultClusterWaitTimeoutInMinutes)
```
gdbranco, line 890: For classic the default should still be 60 :thinking:

provider/clusterrosa/hcp/resource.go
```diff
@@ -320,6 +320,14 @@ func (r *ClusterRosaHcpResource) Schema(ctx context.Context, req resource.Schema
 				Description: "Wait until the cluster standard compute pools are created. The waiter has a timeout of 60 minutes, with the default value set to false. This can only be provided when also waiting for create completion.",
 				Optional:    true,
 			},
+			"cluster_wait_timeout_in_minutes": schema.Int64Attribute{
+				Description: "This value sets the maximum duration in minutes to wait for the cluster to be in a ready state.",
+				Optional:    true,
+			},
+			"resource_wait_timeout_in_minutes": schema.Int64Attribute{
+				Description: "This value sets the maximum duration in minutes to wait for the cluster resource to be in a ready state.",
```
gdbranco, line 328: I think this might be confusing for customers, as the cluster resource in terraform lingo means the cluster itself
Maybe subresources? But I still feel like we should highlight that it is for the default machine pool somehow as it might still be confused with other sub resources such as idp etc

provider/common/cluster_waiter.go
```diff
@@ -173,3 +173,10 @@ func pollClusterState(clusterId string, ctx context.Context, timeout int64, clus
 
 	return object, nil
 }
+
+func ValidateTimeout(timeOut *int64, defaultTimeout int64) *int64 {
+	if timeOut == nil || *timeOut <= 0 {
```
gdbranco, line 178: Would it be confusing for the customer to get the default when configured below zero? I feel like it should throw an error instead. We could add a validation directly to the attribute declaration instead of having that validation here and just handling nil -> default instead

provider/clusterrosa/hcp/resource.go
den-rgb: I'll change it back to `Default_wait_timeout_in_minutes` as I'll be using that value for classic clusters also

provider/clusterrosa/classic/cluster_rosa_classic_resource.go
```diff
@@ -457,6 +457,10 @@ func (r *ClusterRosaClassicResource) Schema(ctx context.Context, req resource.Sc
 				Description: "Wait until the cluster is either in a ready state or in an error state. The waiter has a timeout of 60 minutes, with the default value set to false",
 				Optional:    true,
 			},
+			"default_wait_timeout_in_minutes": schema.Int64Attribute{
```
gdbranco, line 460: Should this be called max_wait_timeout instead?

provider/clusterrosa/classic/cluster_rosa_classic_state.go
```diff
@@ -76,7 +76,8 @@ type ClusterRosaClassicState struct {
 
 	UpgradeAcksFor types.String `tfsdk:"upgrade_acknowledgements_for"`
 
-	DisableWaitingInDestroy types.Bool  `tfsdk:"disable_waiting_in_destroy"`
-	DestroyTimeout          types.Int64 `tfsdk:"destroy_timeout"`
-	WaitForCreateComplete   types.Bool  `tfsdk:"wait_for_create_complete"`
+	DisableWaitingInDestroy     types.Bool  `tfsdk:"disable_waiting_in_destroy"`
+	DestroyTimeout              types.Int64 `tfsdk:"destroy_timeout"`
+	WaitForCreateComplete       types.Bool  `tfsdk:"wait_for_create_complete"`
+	ClusterWaitTimeoutInMinutes types.Int64 `tfsdk:"cluster_wait_timeout_in_minutes"`
```
gdbranco, line 82: You called it default in the resource but using cluster here in the state, will it work as expected?

docs/resources/cluster_rosa_hcp.md
```diff
@@ -75,6 +75,8 @@ resource "rhcs_cluster_rosa_hcp" "rosa_sts_cluster" {
 - `host_prefix` (Number) Length of the prefix of the subnet assigned to each node. After the creation of the resource, it is not possible to update the attribute value.
 - `kms_key_arn` (String) Used to encrypt root volume of compute node pools. The key ARN is the Amazon Resource Name (ARN) of a AWS Key Management Service (KMS) Key. It is a unique, fully qualified identifier for the AWS KMS Key. A key ARN includes the AWS account, Region, and the key ID(optional). After the creation of the resource, it is not possible to update the attribute value.
 - `machine_cidr` (String) Block of IP addresses for nodes. After the creation of the resource, it is not possible to update the attribute value.
+- `max_hcp_cluster_wait_timeout_in_minutes` (Number) This value sets the maximum duration in minutes to wait for a HCP cluster to be in a ready state.
```
gdbranco, line 78: Might be a personal preference, but I think `max_wait_timeout_in_minutes` should be for the cluster control plane so that it is closer to classic. Than the second value we add max_machine_pool_wait...

den-rgb: but then wouldn't ` max_machine_pool_wait` value be used for classic cluster creation, unless we duplicate ` max_machine_pool_wait` (60) and rename it for classic cluster creation it wouldn't make sense to have it named ` max_machine_pool_wait` and then use it for the wait time of a cluster.

gdbranco: It would be the same attribute for waiting the cluster for both :thinking: , waiting the cluster on classic means it also waits for pools

PR #860, opened 2024-12-06
OCM-12705 | feat: Add min/max replica values to docs

den-rgb:
Adding Min/Max info to machinepool replicas for classic and hcp clusters
**Which issue(s) this PR fixes** [OCM-12705](https://issues.redhat.com/browse/OCM-12705)

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


davidleerh:
@den-rgb @gdbranco 
After internal discussion it was deemed that listing the exact requirements within the TF docs and/or data source is not favored. It will be cumbersome to manage if any of these values get updated.
Instead it would be best if we link / provide URL to the official openshift document showing the requirements.

PR #867, opened 2025-01-17
OCM-13503 | feat: ROSA GovCloud Support

andykrohg:
This PR introduces support for ROSA GovCloud environments. 

It adds explicit token refreshes during the cluster wait process, which is required for support in ROSA GovCloud installations, since FedRAMP requirements prohibit a refresh token from having a lifespan greater than 15 minutes (compared to 10 hours on the commercial side). Without this feature, the token times out while waiting for a ROSA cluster to come up.

It also adds GovCloud OCM AWS Account IDs and a GovCloud-only operator policy (openshift_aws_vpce_operator_avo_aws_creds_policy)

**Change type**
- [x] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**

### DIFF
--- a/provider/info/consts.go
+++ b/provider/info/consts.go
@@ -18,13 +18,19 @@ package info
 type ocmEnv string
 
 const (
-	ocmEnvProd  ocmEnv = "production"
-	ocmEnvStage ocmEnv = "stage"
-	ocmEnvInt   ocmEnv = "integration"
+	ocmEnvProd         ocmEnv = "production"
+	ocmEnvStage        ocmEnv = "stage"
+	ocmEnvInt          ocmEnv = "integration"
+	ocmEnvFedRAMPProd  ocmEnv = "fedramp-production"
+	ocmEnvFedRAMPStage ocmEnv = "fedramp-stage"
+	ocmEnvFedRAMPInt   ocmEnv = "fedramp-int"
 )
 
 var ocmAWSAccounts = map[ocmEnv]string{
-	ocmEnvProd:  "710019948333",
-	ocmEnvStage: "644306948063",
-	ocmEnvInt:   "896164604406",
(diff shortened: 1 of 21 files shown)


provider/info/consts.go
```diff
@@ -18,13 +18,19 @@ package info
...
 )
 
 var ocmAWSAccounts = map[ocmEnv]string{
-	ocmEnvProd:  "710019948333",
-	ocmEnvStage: "644306948063",
-	ocmEnvInt:   "896164604406",
+	ocmEnvProd:         "710019948333",
+	ocmEnvStage:        "644306948063",
+	ocmEnvInt:          "896164604406",
+	ocmEnvFedRAMPProd:  "448648337690",
```
hunterkepley, line 33: Is it fine to have these on GitHub? Is there another way we can store them if not?

hunterkepley (commented):
mostly LGTM, but I am unsure about if we should have gov AWS account IDs on GitHub

provider/info/consts.go
andykrohg: Does it help to know that the `rosa` folks did it first? 😄 https://github.com/openshift/rosa/blob/cca9ff58f9eeffdd5de120defd61e0fe7d096fd0/pkg/fedramp/config.go#L37

hunterkepley:
Need `TICKET-#### | type: XXXX` format for MR title and commit

provider/info/consts.go
hunterkepley: yes it does 😛

andykrohg:
Hmm, I could be wrong, but the latest failures seem to be unrelated to my changes

PR #946, opened 2025-08-01
OCM-17531 | feat: Enable refresh token support in OCM API auth flow

ironcladlou:
Before this commit, it was impossible to authenticate to OCM API endpoints which require refresh tokens, such as the GovCloud internal OCM api. This made it impossible to use the provider with ROSA GovCloud internally.

This commit adds support by:

- Exposing a new `refresh_token` provider variable which is appended to the OCM authentication builder token list

- Tweaks the client secret handling to unconditionally pass the client secret variable to the auth builder, even when empty. From the OCM SDK docs:

        // Note that some OpenID providers (Keycloak, for example) require the client identifier also for
        // the resource owner password grant. In that case use the set only the identifier, and let the
        // secret blank. For example:

  This change should thus support all workflows in a backwards compatible way.

After this change, the provider can successfully authenticate to ROSA GovCloud OCM API endpoints when appropriately configured.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -70,6 +70,7 @@ type Config struct {
 	URL          types.String `tfsdk:"url"`
 	TokenURL     types.String `tfsdk:"token_url"`
 	Token        types.String `tfsdk:"token"`
+	RefreshToken types.String `tfsdk:"refresh_token"`
 	ClientID     types.String `tfsdk:"client_id"`
 	ClientSecret types.String `tfsdk:"client_secret"`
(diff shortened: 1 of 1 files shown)


hunterkepley:
QE said the e2e test failures are unrelated, so if the unit tests pass (and the E2E test failure is still the same) ill merge 🙂

hunterkepley:
We've been low manpower since the re-org and these e2e tests have eluded us for some time

ironcladlou:
Great, thanks!

PR #947, opened 2025-08-08
OCM-17691 | fix: Make billing account ID optional to enable GovCloud HCP cluster creation

ironcladlou:
Before this commit, HCP clusters on ROSA GovCloud couldn't be created because the
module required a value for the billing account variable, which is invalid to
supply to the OCM API when creating HCP GovCloud clusters[1]. In these cases,
the billing ID passed to the OCM `/clusters` API should be empty.

This commit makes billing account ID optional so that users can create HCP clusters
on ROSA GovCloud.

This is consistent with the server side validation OCM Cluster Service applies to
the billing ID, which are dynamic based on many criteria[2]. Instead of trying to
replicate the server side validation in the provider, this change simply makes the
field always optional and the server can perform those complex validations.

[1] https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd.go#L1268
[2] https://github.com/openshift-online/ocm-cluster-service/blob/main/cmd/clusters-service/service/cluster_service.go


ironcladlou:
@jerichokeyne thanks, I put this in a WIP status and haven't had a chance to go over any changes to tests, make the Jira issue, etc. I'll add the doc generation stuff to my list. #946 merged so thats rebased out of this PR now

ironcladlou:
@jerichokeyne I created and linked an upstream bug for this, so I think the next step would be to validate/accept the bug itself (if people agree). Thank you!

PR #957, opened 2025-09-09
Image mirrorOCM-18302 | feat: Support for image registry mirrors 

philipwu08:
Support for image registry mirrors

Fixes #

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**


docs/resources/image_mirror.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `cluster_id` (String) The ID of the ROSA cluster
+- `mirrors` (List of String) List of mirror registries that will serve content for the source
+- `source` (String) The source registry that will be mirrored
```
davidleerh, line 22: Should we imply that this is required during creation only?

hunterkepley:
This could also be an issue with the job itself. We can change the fetch depth to `0` on the `Validate Commit Messages` job (not sure if it's like this already), but we should check that the BASE SHA exists in the first place

PR #1032, opened 2026-01-20
OCM-18335 | fix: machinepool EC version upgrades

olucasfreitas:
**Change type**
- [ ] New feature
- [x] Bug fix
- [ ] Build
- [ ] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**

# Details

This PR fixes a bug where HCP machine pool upgrades between early candidate (EC) versions would fail with the error "cluster version is already above the requested version", even when upgrading to a newer version (e.g., from `4.20.0-ec.5` to `4.20.0-ec.6`).

---

## Fixed Behavior

1. Upgrade a machine pool from one EC version to a newer EC version:

    ```hcl
    resource "rhcs_hcp_machine_pool" "mps" {
      # ...
      version = "4.20.0-ec.6"  # Upgrading from 4.20.0-ec.5
    }
    ```

2. Now correctly identifies that `4.20.0-ec.6` is newer than `4.20.0-ec.5` and schedules the upgrade:

    ```
    rhcs_hcp_machine_pool.mps[0]: Modifying... [id=np-72513-uu]
    rhcs_hcp_machine_pool.mps[0]: Modifications complete
    ```

---

## Previous Behavior

Before this fix, upgrading between EC versions would fail because the channel group suffix (e.g., `-candidate`) was not being trimmed from the version ID stored in state:

```bash
terraform apply
```

Output:
```
Error: Can't upgrade machine pool

  with rhcs_hcp_machine_pool.mps[0],
  on main.tf line 28, in resource "rhcs_hcp_machine_pool" "mps":
  28: resource "rhcs_hcp_machine_pool" "mps" {

Can't upgrade machine pool version with identifier: `np-72513-uu`, cluster
version is already above the requested version
```

The root cause was that the state stored `4.20.0-ec.5-candidate` instead of `4.20.0-ec.5`, causing the version comparison to incorrectly determine that `ec.5-candidate` > `ec.6` (lexicographically).

---

# Ticket

Closes [OCM-18335](https://issues.redhat.com/browse/OCM-18335)

### DIFF
--- a/provider/machinepool/hcp/machine_pool_resource.go
+++ b/provider/machinepool/hcp/machine_pool_resource.go
@@ -1340,6 +1340,8 @@ func populateState(ctx context.Context, object *cmv1.NodePool, state *HcpMachine
 		version, ok := object.Version().GetID()
 		// If we're using a non-default channel group, it will have been appended to
 		// the version ID. Remove it before saving state.
+		channelGroup, _ := cluster.Version().GetChannelGroup()
+		version = strings.TrimSuffix(version, fmt.Sprintf("-%s", channelGroup))
 		version = strings.TrimPrefix(version, rosa.VersionPrefix)
 		if ok {
 			state.CurrentVersion = types.StringValue(version)


olucasfreitas:
@hunterkepley I checked the cluster upgrade, the `rc` and `ec` versions, all seemed to work correctly after the fix, bellow is an explanation why I did the implementation like this.

The reason I did it like this, it that the API returns version IDs like `openshift-v4.21.0-ec.0-candidate` for candidate channel clusters. The code striped `openshift-v` but was missing the code to strip `-candidate`, so the stored version becomes `4.21.0-ec.0-candidate` instead of `4.21.0-ec.0`.
The library we use seemed to think that the candidate was part of the version resulting in the provider incorrectly believing that the current version is higher than the requested upgrade version, so it failed.

here is the output from calling the ocm api

$ ocm get /api/clusters_mgmt/v1/clusters/2nvsnm55v9c8jpvdimjf9t6j0cgueqah | jq '.version.id'
"openshift-v4.21.0-ec.3-candidate"

hunterkepley:
> @hunterkepley I checked the cluster upgrade, the `rc` and `ec` versions, all seemed to work correctly after the fix, bellow is an explanation why I did the implementation like this.
> 
> The reason I did it like this, it that the API returns version IDs like `openshift-v4.21.0-ec.0-candidate` for candidate channel clusters. The code striped `openshift-v` but was missing the code to strip `-candidate`, so the stored version becomes `4.21.0-ec.0-candidate` instead of `4.21.0-ec.0`. The library we use seemed to think that the candidate was part of the version resulting in the provider incorrectly believing that the current version is higher than the requested upgrade version, so it failed.
> 
> here is the output from calling the ocm api
> 
> $ ocm get /api/clusters_mgmt/v1/clusters/2nvsnm55v9c8jpvdimjf9t6j0cgueqah | jq '.version.id' "openshift-v4.21.0-ec.3-candidate"


> The library we use seemed to think 
The library is our OCM SDK

Could you provide output of the cluster upgrade(s) you tried out? You posted this comment about 30 minutes after I asked and it takes a bit longer than 30 minutes to upgrade clusters

Also, [please have a look at this Go Playground example I made](https://goplay.tools/snippet/OSB5Z7sQEdO) - I think this may not actually be fixing the issue

PR #1070, opened 2026-02-19
OCM-22297 | ci: Fixing id:72514,id:72478, upgrade tests, disabling IMDS tests for now

jerichokeyne:
**What this PR does / why we need it**: This is fixing a few CI failures that have been coming up lately
- `id:72514` - assumed that you can't create a node pool with autoscaling enabled and `min_replicas` set to `0`, but that actually seems to work fine
- `id:72478` - tried to change the channel group to `eus`, but that channel group only has even numbered y-stream versions. I switched it to try to use either the `fast` or `candidate` channel groups as those both seem to have the same versions as `stable` (which is required for the edit to succeed)
- A lot of the tests under `tests/e2e/hcp_image_mirror_test.go` have been very flaky since they've been introduced. Until we have time to re-write those tests I'm excluding the entire test suite to minimize the number of false negatives we get
- The upgrades tests needed updating to set the `channel` property on the cluster before listing the y-stream upgrades due to a recent API change. I also noticed that we had separate test cases for HCP clusters vs classic clusters, but they only seemed to really differ by a single line, so I combined the 4 test cases down to 2 test cases (y-stream vs z-stream) for now
  - I've tested that both z-stream and y-stream upgrades work well for HCP clusters
  - I've tested y-stream upgrades for classic clusters, but ran into an issue where it doesn't work on the first apply, but does work on the second apply. I think there may be either a provider bug or an API bug where the `upgrade_acknowledgements_for` field isn't actually working properly. I'm still trying to narrow down the issue, but I don't think it's related to my testing code changes at all as this was a known issue before as well. I tested z-stream upgrades for classic clusters, and didn't see any issues
  - I also had to update the `ocm-sdk-go` library version to be able to set the new `channel` property

Fixes #[OCM-22297](https://issues.redhat.com/browse/OCM-22297)

**Change type**
- [ ] New feature
- [ ] Bug fix
- [ ] Build
- [x] CI
- [ ] Documentation
- [ ] Performance
- [ ] Refactor
- [ ] Style
- [ ] Unit tests
- [ ] Subsystem tests

**Checklist**

**Local logs:**
- [id:72478](https://github.com/user-attachments/files/25423227/72478-log.txt)
- [id:72514](https://github.com/user-attachments/files/25423229/72514-log.txt)
- [HCP y-stream upgrade](https://github.com/user-attachments/files/25423230/hcp-up-y-log.txt)
- [HCP z-stream upgrade](https://github.com/user-attachments/files/25423231/hcp-up-z-log.txt)
- [Classic y-stream upgrade](https://github.com/user-attachments/files/25493367/sts-up-y-log.txt)
- [Classic z-stream upgrade](https://github.com/user-attachments/files/25493369/sts-up-z-log.txt)

## Summary by CodeRabbit

* **Tests**
  * Refactored cluster upgrade and edit test suites for improved organization and reduced duplication
  * Added new testing utilities for version management and cluster channel updates
  * Disabled HCP ImageMirror test suite due to known flakiness
  * Removed outdated autoscaling validation test case

### DIFF
--- a/tests/e2e/cluster_upgrade_test.go
+++ b/tests/e2e/cluster_upgrade_test.go
@@ -25,12 +25,14 @@ var _ = Describe("Upgrade", func() {
 	defer GinkgoRecover()
 
 	var (
-		targetV        string
-		clusterID      string
-		profileHandler profilehandler.ProfileHandler
-		profile        profilehandler.ProfileSpec
-		clusterArgs    *exec.ClusterArgs
-		clusterService exec.ClusterService
+		targetV           string
+		availableUpgrades []string
+		clusterID         string
+		profileHandler    profilehandler.ProfileHandler
+		profile           profilehandler.ProfileSpec
+		clusterArgs       *exec.ClusterArgs
+		clusterService    exec.ClusterService
+		channelGroup      string
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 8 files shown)


tests/e2e/cluster_upgrade_test.go
```diff
@@ -46,235 +48,164 @@ var _ = Describe("Upgrade", func() {
...
+				return clusterDetails.Body().Version().AvailableUpgrades()
+			}).WithTimeout(time.Minute * 5).WithPolling(time.Second * 1).Should(ContainElement(targetV))
+
+			Logger.Infof("Starting cluster upgrade to version '%s' on channel '%s'", targetV, targetChannel)
+
+			By("Apply the cluster Upgrade")
+			clusterArgs.OpenshiftVersion = helper.StringPointer(targetV)
+			clusterArgs.UpgradeAcknowledgementsFor = helper.StringPointer(majorVersion)
+			_, err = clusterService.Apply(clusterArgs)
+			Expect(err).ToNot(HaveOccurred())
```
olucasfreitas: Nice cleanup on consolidating the upgrade cases. I'm still worried about the classic Y-stream path though: the PR description says classic can fail on the first `Apply()` because `upgrade_acknowledgements_for` isn't being honored reliably yet. With the shared HCP/classic flow here we still assert a single success path for classic, so this can keep producing false negatives. Could we split classic back out or add a temporary known-issue guard/retry until that provider/API issue is resolved?

tests/e2e/cluster_upgrade_test.go
```diff
@@ -47,235 +49,170 @@ var _ = Describe("Upgrade", func() {
...
+				return clusterDetails.Body().Version().AvailableUpgrades()
+			}).WithTimeout(time.Minute * 5).WithPolling(time.Second * 1).Should(ContainElement(targetV))
+
+			Logger.Infof("Starting cluster upgrade to version '%s' on channel '%s'", targetV, targetChannel)
+
+			By("Apply the cluster Upgrade")
+			clusterArgs.OpenshiftVersion = helper.StringPointer(targetV)
+			clusterArgs.UpgradeAcknowledgementsFor = helper.StringPointer(majorVersion)
+			_, err = clusterService.Apply(clusterArgs)
+			if err != nil && !profile.GetClusterType().HCP {
```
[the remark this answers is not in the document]
jerichokeyne: I'm finally getting back to this now, and it appears that this is now working for classic upgrades just fine without a retry, so I removed the retry now

[classic-up-y.txt](https://github.com/user-attachments/files/27495044/classic-up-y.txt)

tests/e2e/cluster_upgrade_test.go
```diff
@@ -47,235 +49,164 @@ var _ = Describe("Upgrade", func() {
...
+				Skip("The test is configured only for Y-stream upgrade")
+			}
 
-				By("Run the cluster update")
-				clusterArgs.OpenshiftVersion = helper.StringPointer(targetV)
+			By("Get cluster info")
+			clusterResp, err := cms.RetrieveClusterDetail(cms.RHCSConnection, clusterID)
+			Expect(err).ToNot(HaveOccurred())
+			targetV, err = cms.GetVersionUpgradeTarget(clusterResp.Body().Version().RawID(),
+				constants.Y, availableUpgrades)
```
olucasfreitas, line 131: This still computes `targetV` from `availableUpgrades` loaded before `ChangeClusterChannel()`. If Y-stream upgrades only appear after the channel update, this can fail here before the new flow is exercised.

jerichokeyne: It's getting `availableUpgrades` from the new version API that should be independent of the cluster's channel setting. It kinda needs to happen before the channel update happens because this is how it finds what channel to update to

```
// tests/utils/cms/versions.go
func GetVersionUpgrades(client *client.Connection, originalVersion string) (upgrades []string, err error) {
	res, err := client.ClustersMgmt().V1().Versions().Version(originalVersion).Get().Send()
	if err != nil {
		return nil, err
	}
	return res.Body().AvailableUpgrades(), nil
}
```

```
// tests/e2e/cluster_upgrade_test.go
availableUpgrades, err = cms.GetVersionUpgrades(cms.RHCSConnection, clusterResp.Body().Version().ID())
```

jerichokeyne:
/override ci/prow/e2e-presubmits-rosa-hcp-advanced-critical-high-presubmit

PR #1131, opened 2026-04-22
OCM-21915 | feat: link ocm role resource

gdbranco:
## PR Summary
With the soon to be enabled enforcement of the OCM role workflow all clients require an option to allow the creation of the role. TF RHCS still lacks such a resource.

The current PR allows to link an OCM role to OCM, the creation of the role itself will be conducted at the module level since it requires AWS interaction

## Related Issues and PRs

- Jira: [OCM-21915](https://redhat.atlassian.net/browse/OCM-21915)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

## How to Test (Step-by-Step)
```
terraform {
  required_providers {
    rhcs = {
      version = ">= 1.1.0"
      source  = "terraform.local/local/rhcs"
    }
  }
}

provider "rhcs" {
}

resource "rhcs_link_ocm_role" "link_ocm_role" {
  role_arn = var.role_arn
}

```
### Preconditions
Create a proper OCM role in AWS, it is possible to use ROSA CLI for instance

### Test Steps
1. rosa create ocm-role -y
2. Save role arn
3. Forward role arn to link role resource

### Expected Results
AMS label for sts_ocm_role should include the new OCM role association

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [x] `make build` passes.
- [x] Documentation was added/updated where appropriate.
- [ ] Any risk, limitation, or follow-up work is documented.

## Summary by CodeRabbit

* **New Features**
  * Added a Terraform resource and a data source to link AWS IAM role ARNs to OCM organizations (create/read/delete/import, ARN validation, composite IDs).

* **Documentation**
  * New docs page describing the resource, schema, and required inputs.

* **Tests**
  * New unit, subsystem and end-to-end suites, manifests and test helpers; e2e tests can run without a cluster profile and include non-cluster test labeling.

* **Chores**
  * Updated ignore rules to exclude local envrc files.


provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,474 @@
...
+const (
+	OCMRoleLabel = "sts_ocm_role"
+)
+
+var (
+	RoleArnRE = regexp.MustCompile(`^arn:(?:aws|aws-us-gov|aws-cn):iam::\d{12}:role\/.+$`)
+	// labelMutex protects label read-modify-write operations to prevent
+	// concurrent updates from causing lost writes when multiple resources
+	// modify the same sts_ocm_role label simultaneously
+	labelMutex sync.Mutex
```
amandahla, line 34: @coderabbitai 
I'm concerned about potential concurrency issues in this Terraform provider implementation and would like your expert analysis on whether these concerns are valid:

**Context:**
This is a Terraform provider using terraform-plugin-framework (not the legacy SDK). The resource `rhcs_link_ocm_role` manages linking AWS IAM roles to OCM organizations by updating a label via the OCM API. The label is a comma-separated list of role ARNs.

The reviewer flagged this code at line 116:
```go
var (
    RoleArnRE = regexp.MustCompile(`^arn:(?:aws|aws-us-gov|aws-cn):iam::\d{12}:role\/.+$`)
    labelMutex sync.Mutex  // Package-level global
)
The mutex is used in Create() (line 512) and Delete() (line 308) to protect read-modify-write operations on the OCM organization label.

Reviewer's critique:
- Multiple Terraform provider instances (separate processes) won't share this mutex, so it doesn't prevent cross-process races
- Within a single provider process, this mutex blocks ALL instances of this resource, even when managing different organizations
- Breaks testability for parallel tests
Suggests: Move mutex to struct, use per-org locking, or rely on OCM API atomic operations

Questions (consider that this will be used by modules defined in https://github.com/terraform-redhat/terraform-rhcs-rosa-hcp)
1) For Terraform Plugin Framework providers, does Terraform spawn multiple provider processes that could race on the same OCM organization label?
2) Is a package-level mutex protecting label operations across different orgs an anti-pattern, or is it acceptable given Terraform's execution model?
3) Does the OCM SDK (github.com/openshift-online/ocm-sdk-go) provide any optimistic locking, ETags, or atomic update mechanisms for label updates that would be better than client-side mutex?
4) What's the correct pattern for protecting read-modify-write operations in Terraform providers when the API doesn't guarantee atomicity?
5) Should we warn the user about it?
6) If we decide to keep the mutex, how do we handle timeouts inside of it?

File: provider/ocmrolelink/ocm_role_link_resource.go [hunk 2]
```diff
@@ -0,0 +1,474 @@
...
+
+	return exists, linkedArns, nil
+}
+
+func (r *LinkOCMRoleResource) linkRole(ctx context.Context, orgID, roleArn, awsAccountID string) error {
+	// Lock to prevent concurrent label modifications from causing lost updates
+	labelMutex.Lock()
+	defer labelMutex.Unlock()
+
+	exists, linkedArns, err := r.checkIfRoleLinked(ctx, orgID, roleArn, awsAccountID)
```
amandahla, line 499: ROSA [cli](https://github.com/openshift/rosa/blob/b197921c1f61d35afa86eeb01d3230bb7239960e/cmd/link/userrole/cmd.go#L120) checks if the role exists.
Does the OCM API returns an error if the provider try to link to non-existent role?

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+		resp.Diagnostics.AddError(
+			"Failed to verify role link",
+			fmt.Sprintf("Unable to check if role %s is linked: %s", roleArn, err.Error()),
+		)
+		return
+	}
+
+	// Check if the specific role is in the label
+	linkedArns := strings.Split(labelsResp.Body().Value(), ",")
```
amandahla, line 398: Shouldn import checks also if the org already has role from AWS account?

provider/ocmrolelink/ocm_role_links_data_source.go
```diff
@@ -0,0 +1,135 @@
...
+	tflog.Debug(ctx, "Fetching current account organization")
+	acctResp, err := d.connection.AccountsMgmt().V1().CurrentAccount().Get().Send()
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to get current account",
+			fmt.Sprintf("Unable to retrieve current user's organization: %s", err.Error()),
+		)
+		return
+	}
+	orgID := acctResp.Body().Organization().ID()
```
amandahla, line 89: Lets say that I ran the terraform apply with a token in a specific org and then I decided to change it. Now I'm running again terraform apply but in a different org.
Would make sense checking that the current account org is the same as the one in the state and if not warn the user with something like "The OCM role link was created in organization blabla but the current user is in organization  blublu" ?

provider/ocmrolelink/ocm_role_link_resource.go
gdbranco: No way to currently do it through OCM AMS, I was also considering this. Even if the customer links to a wrong role/value the actual role also includes another layer of security which is an external-id. If that does not match the trust is broken and OCM is still not able to assume it and will report that the role does not match the expected pattern `$PREFIX-OCM-ROLE-$EXTERNAL_ORG_ID` and if it does match but trust isn't there (different external-id) it will fail reporting that there's no trust to the account

Yes, rosa cli can check it due to having access to local aws creds. But if the customer wants they can also use OCM API for AMS directly right now and bypass that

provider/ocmrolelink/ocm_role_links_data_source.go
gdbranco: We could, not sure if that's an expectation or not. With ROSA CLI you wouldn't get such a message :thinking: 

Not a problem at all if would you like to provide more info to the customer on this

provider/ocmrolelink/ocm_role_link_resource.go
amandahla: @gdbranco human comment here:
I understand the reason of using a mutex here and makes sense accordingly to how other providers approach this.
1) My main concern is that the OCM call is not using context so if a timeout happens or if the user cancel the operation, this operation will hang/leak.
Is there a reason for not using the SendContext here?
2) wdyt about adding debug logs here like "Attempting to acquire mutex" ?

amandahla:
Minor: usually a verb does not start a terraform resource name. wdyt about inverting to ocm_role_link?

provider/ocmrolelink/ocm_role_link_resource.go
amandahla: Yes, I'm afraid that we might only document that as "must exist" or something similar so the user will be aware of that.

provider/ocmrolelink/ocm_role_links_data_source.go
amandahla: If is not a complex check, I think its worthy :)

olucasfreitas (commented):
A few follow-ups from code review (excluding items already handled or that were local-environment noise).

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,474 @@
...
+		)
+		return
+	}
+
+	if labelsResp.Status() == http.StatusNotFound {
+		tflog.Info(ctx, fmt.Sprintf("OCM role label not found for organization %s, considering already unlinked", orgID))
+		return
+	}
+
+	linkedArns := strings.Split(labelsResp.Body().Value(), ",")
```
olucasfreitas, line 286: Comma-delimited label values are fragile if an ARN ever contains a comma (unlikely for IAM ARNs today, but brittle as a serialization format). Consider documenting this limitation explicitly in provider docs, or switching to a safer encoding if OCM ever allows broader values.

provider/linkocmrole/link_ocm_role_resource.go
```diff
@@ -0,0 +1,474 @@
...
+	defer labelMutex.Unlock()
+
+	labelsResp, err := r.connection.AccountsMgmt().V1().Organizations().
+		Organization(orgID).
+		Labels().
+		Labels(OCMRoleLabel).
+		Get().
+		Send()
+
+	if err != nil && labelsResp != nil && labelsResp.Status() != http.StatusNotFound {
```
olucasfreitas: If `Get().Send()` returns a transport error where `labelsResp` is nil, this condition short-circuits and the `err != nil` branch won’t run, so we may fall through with a nil `labelsResp` and mis-handle the failure mode. Consider structuring this like `Read()`/`ImportState()` (check `err` first, then branch on `labelsResp != nil && labelsResp.Status() == 404`, etc.).

provider/linkocmrole/link_ocm_role_resource.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	tflog.Info(ctx, fmt.Sprintf("Successfully unlinked role %s from organization %s", roleArn, orgID))
+}
+
+func (r *LinkOCMRoleResource) ImportState(ctx context.Context,
+	req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	roleArn := req.ID
+
+	_, err := arn.Parse(roleArn)
```
olucasfreitas: `provider/ocmrolelink/ocm_role_link_resource.go`: the remaining mismatch is the delimiter guard. `Schema()` rejects commas in `role_arn`, but `ImportState()` only checks `RoleArnRE`, so import can still accept a value that normal apply would reject and that read/delete later split on `,`. Reusing one shared validator/helper for schema and import would keep both paths aligned.

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,474 @@
...
+
+		linkedArns = append(linkedArns, trimmed)
+	}
+
+	return exists, linkedArns, nil
+}
+
+func (r *LinkOCMRoleResource) linkRole(ctx context.Context, orgID, roleArn, awsAccountID string) error {
+	// Lock to prevent concurrent label modifications from causing lost updates
+	labelMutex.Lock()
```
olucasfreitas, line 496: While `labelMutex` is held, OCM calls use `Send()` rather than `SendContext(ctx)`. That means cancellation/deadline from Terraform won’t propagate, and a hung request can block other link/unlink operations for longer than necessary. Prefer `SendContext(ctx)` for network calls under the mutex (including the nested `checkIfRoleLinked` GET).

tests/e2e/e2e_suite_test.go
```diff
@@ -28,9 +29,18 @@ var _ = BeforeSuite(func() {
 	err = helper.AlignRHCSSourceVersion(config.GetManifestsDir())
 	Expect(err).ToNot(HaveOccurred())
 
-	profileHandler, err := profilehandler.NewProfileHandlerFromYamlFile()
-	Expect(err).ToNot(HaveOccurred())
-	clusterID, err = profileHandler.RetrieveClusterID()
-	Expect(err).ToNot(HaveOccurred())
+	// Check if running tests have the NonClusterTest label
+	// If so, skip loading cluster profile (tests will create their own standalone handlers)
+	filter := GinkgoLabelFilter()
```
olucasfreitas: `GinkgoLabelFilter()` returns the raw filter string (e.g. `day1 && non-cluster-test`), not a boolean expression evaluator, so `ci.NonClusterTest.MatchesLabelFilter(filter)` won’t reliably detect non-cluster suites. Net effect: `BeforeSuite` still loads `CLUSTER_PROFILE` even when running only `ci.NonClusterTest` specs, which contradicts the intent described in `link_ocm_role_test.go`.

Consider using `Label(`ci.NonClusterTest`)` on the `Describe` blocks that must skip cluster profile setup, or a more robust label filter parse.

tests/non_cluster_e2e/ocm_role_link_test.go
```diff
@@ -0,0 +1,102 @@
...
+		profileHandler, err = profilehandler.NewStandaloneProfileHandler("link-ocm-role")
+		Expect(err).ToNot(HaveOccurred())
+
+		linkOCMRoleService, err = profileHandler.Services().GetLinkOCMRoleService()
+		Expect(err).ToNot(HaveOccurred())
+
+		// Generate a test role ARN
+		// In a real e2e test, this should be an actual AWS IAM role ARN that exists
+		testRoleName := "rhcse2e-OCM-Role-" + helper.GenerateRandomStringWithSymbols(8)
+		testRoleArn = "arn:aws:iam::123456789012:role/" + testRoleName
```
olucasfreitas, line 50: These tests generate a plausible IAM role ARN, but it won’t exist in AWS/OCM for real org linking. That’s OK for plan/validation coverage, but the `Apply`/`Destroy` path likely only works meaningfully against a stubbed/local stack.

If the goal is CI signal, consider splitting: keep fast validation tests, and gate true link/unlink integration behind an opt-in env var + real role ARN fixture.

provider/ocmrolelink/ocm_role_link_resource.go
gdbranco: Yeah, can't do much since this is how OCM AMS decided to store the values. Not sure where we could document it in TF though, since this is not customer facing 🤔

provider/linkocmrole/link_ocm_role_resource.go
gdbranco: We can, I don't think that should happen. But it is fair

tests/non_cluster_e2e/ocm_role_link_test.go
gdbranco: We can improve the e2e tests later with a real OCM role once we start building those on the fly, I'll add a module change that will assist in that

tests/e2e/e2e_suite_test.go
gdbranco: I can separate the suite for NonClusterTest entirely to guarantee the cluster_profile is never used in that suite. I don't see a better way around it though since ginkgo does not provide some form of contains label 🤔

gdbranco: hmm, I'm trying to workaround what ginkgo provides and just compare against string slice. I'll update briefly

olucasfreitas (commented):
A couple of additional follow-ups after reloading the current head and re-running the non-cluster suite locally.

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,520 @@
...
+
+func (r *OCMRoleLinkResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Link an OCM role to an OCM organization",
+		Attributes: map[string]schema.Attribute{
+			"role_arn": schema.StringAttribute{
+				Description: "ARN of the AWS IAM role to link to the organization",
+				Required:    true,
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(
```
olucasfreitas, line 62: I think this is still worth tightening. IAM allows commas in role names and paths, and we still split sts_ocm_role on commas, so a valid ARN may not round-trip safely. Would rejecting commas in role_arn for now make sense?

tests/e2e/e2e_suite_test.go
```diff
@@ -28,9 +30,34 @@ var _ = BeforeSuite(func() {
...
+	// Skip only if CLUSTER_PROFILE is not set AND we're running non-cluster tests only
+	clusterProfile := config.GetClusterProfile()
+	labelFilter := GinkgoLabelFilter()
+
+	// Extract the label string from the constant (Labels is []string)
+	nonClusterTestLabel := ci.NonClusterTest[0]
+
+	// Simple heuristic: skip cluster profile if not set and filter explicitly includes non-cluster-test
+	// without negation (!) - handles "non-cluster-test", "non-cluster-test && day1", etc.
+	skipClusterProfile := clusterProfile == "" && labelFilter != "" &&
```
olucasfreitas, line 43: I reran the focus-only path on the current head with go test ./tests/e2e -run TestRHCSProvider -args -ginkgo.focus="Link OCM Role", and it still panics on missing CLUSTER_PROFILE in BeforeSuite. Could we either support focus-only runs here or update the local instructions to require --label-filter=non-cluster-test?

docs/resources/ocm_role_link.md
```diff
@@ -0,0 +1,24 @@
...
+Link an OCM role to an OCM organization
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `role_arn` (String) ARN of the AWS IAM role to link to the organization
```
olucasfreitas: I may be missing it, but I do not see a note that the IAM role must already exist and that this resource only links or unlinks it. Could we add a short note around here?

tests/non_cluster_e2e/ocm_role_link_test.go
```diff
@@ -0,0 +1,102 @@
...
+		profileHandler, err = profilehandler.NewStandaloneProfileHandler("ocm-role-link")
+		Expect(err).ToNot(HaveOccurred())
+
+		ocmRoleLinkService, err = profileHandler.Services().GetOCMRoleLinkService()
+		Expect(err).ToNot(HaveOccurred())
+
+		// Generate a test role ARN
+		// In a real e2e test, this should be an actual AWS IAM role ARN that exists
+		testRoleName := "rhcse2e-OCM-Role-" + helper.GenerateRandomStringWithSymbols(8)
+		testRoleArn = "arn:aws:iam::123456789012:role/" + testRoleName
```
olucasfreitas, line 50: I think this is still worth tightening. Since AMS does not reject a non-existent role here, this test can change a real org with a generated ARN without proving a real end-to-end link. Would gating create and destroy on a real ARN from env or config work here?

subsystem/common/ocm_role_link_resource_test.go
```diff
@@ -0,0 +1,904 @@
...
+				CombineHandlers(
+					VerifyRequest(http.MethodGet, "/api/accounts_mgmt/v1/current_account"),
+					RespondWithOcmObjectMarshal(http.StatusOK, account, amsv1.MarshalAccount),
+				),
+				CombineHandlers(
+					VerifyRequest(http.MethodGet, "/api/accounts_mgmt/v1/organizations/"+orgID+"/labels/sts_ocm_role"),
+					RespondWithOcmObjectMarshal(http.StatusOK, existingLabel, amsv1.MarshalLabel),
+				),
+				CombineHandlers(
+					VerifyRequest(http.MethodGet, "/api/accounts_mgmt/v1/organizations/"+orgID+"/labels/sts_ocm_role"),
```
olucasfreitas: Small test cleanup: I still see duplicate GET handlers for the same label endpoint here. Would you mind trimming the extras so the test stays closer to the actual call count?

gdbranco: I see it, I had the expectation that gomega should check for unconsumed handlers but it only checks if there are missing calls to them instead. I've addressed that by setting up a new test context struct that will hold the instance values so that we can move away from global variables and also included a new verification method to check for unconsumed handlers. Let me know if that works for you

tests/non_cluster_e2e/ocm_role_link_test.go
gdbranco: I'll be creating and destroying a real role once we get the updated module so we can rely on it instead of adhoc-ly recreating the behavior

tests/e2e/e2e_suite_test.go
gdbranco: Oh, I was miss interpreting how the labels are fetched, I thought it would pick up which labels the loaded tests include. But it actually just sets it to what was supplied through ``--label-filter` param. So it would only work if you use `--label-filter=non-cluster-test` instead of focus ...

hmm I don't really have a good way around this, ginkgo does not allow me to see labels from the set of tests in execution. I think I would rather have a separate e2e suite with only non cluster tests then

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,540 @@
...
+		_, err = r.connection.AccountsMgmt().V1().Organizations().
+			Organization(orgID).
+			Labels().
+			Labels(OCMRoleLabel).
+			Update().
+			Body(labelBuilder).
+			SendContext(ctx)
+	}
+
+	if err != nil {
```
olucasfreitas, line 335: `provider/ocmrolelink/ocm_role_link_resource.go`: `Delete()` only turns `403` into the actionable "organization admin" error on the initial label `GET`. If the API allows the read but rejects the actual `PATCH`/`DELETE`, users only get the generic `Failed to unlink OCM role` message even though the code already knows the intended remediation.

docs/data-sources/ocm_role_links.md
olucasfreitas, line 1: `docs/data-sources/ocm_role_links.md`: the PR adds a new `rhcs_ocm_role_links` data source, but the scoped test additions only exercise the `rhcs_ocm_role_link` resource (`subsystem/common/ocm_role_link_resource_test.go`, `tests/non_cluster_e2e/ocm_role_link_test.go`). There is no subsystem/e2e coverage for the data source read path, empty-on-404 behavior, or comma-splitting of `role_arns`, so that new surface ships unvalidated.

provider/ocmrolelink/ocm_role_link_resource.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+
+	labelsResp, err := r.connection.AccountsMgmt().V1().Organizations().
+		Organization(orgID).
+		Labels().
+		Labels(OCMRoleLabel).
+		Get().
+		Send()
+
+	if err != nil {
```
olucasfreitas, line 196: `provider/ocmrolelink/ocm_role_link_resource.go`: `Read()` warns when the current token belongs to a different org than `organization_id` in state, but it still treats a `404` on the old org as "resource gone" and removes state; `Delete()` does not do the mismatch check at all and also treats `404` as "already unlinked". If the user changes OCM token/org between applies, an inaccessible old-org label can be misclassified as deleted, so Terraform can silently forget a still-linked role or recreate it in the new org.

olucasfreitas:
closing this in favor of #1156

PR #1162, opened 2026-05-21
OCM-25158 | fix(hcp): auto_node causes "Provider produced inconsistent result after apply" on every cluster create

paulczar:
## PR Summary

Fix `auto_node` causing "Provider produced inconsistent result after apply" on every ROSA HCP cluster create. The OCM API never returns `auto_node.mode` in GET responses, so `populateRosaHcpClusterState` always set `state.AutoNode = nil`, making Terraform see an inconsistency against the planned value. This PR infers `mode` from `role_arn` presence and issues a post-create PATCH to activate AutoNode mode via the OCM Day-2 API.

Every `terraform apply` that creates a ROSA HCP cluster with `auto_node` configured fails with:

```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.cluster.rhcs_cluster_rosa_hcp.main[0],
│ provider "provider["registry.terraform.io/terraform-redhat/rhcs"]" produced
│ an unexpected new value: .auto_node:
│ was cty.ObjectVal(map[string]cty.Value{"mode":cty.StringVal("enabled"),
│ "role_arn":cty.StringVal("arn:aws:iam::808082629126:role/my-autonode-role")}),
│ but now null.
│
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

The cluster **is** successfully created in OCM with AutoNode configured — the error occurs during Terraform's post-apply state reconciliation, leaving the cluster orphaned in OCM with no state entry.

**Root cause — two separate issues were found:**

### Issue 1: `auto_node.mode` is never returned by OCM GET

`populateRosaHcpClusterState` reads `auto_node.mode` from the top-level cluster object and `aws.auto_node.role_arn` from the nested AWS object, then only populates `state.AutoNode` if **both** fields are present:

```go
if !autoNodeState.Mode.IsNull() && !autoNodeState.RoleARN.IsNull() {
    state.AutoNode = autoNodeState
} else {
    state.AutoNode = nil  // ← always hit: Mode is always null from GET
}
```

The OCM API never returns `auto_node.mode` in GET responses — verified with 20 consecutive requests against a ready, AutoNode-enabled cluster:

```
{"top_level_auto_node":null,"aws_auto_node":{"role_arn":"arn:aws:iam::808082629126:role/cz-demo-autonode-operator-role"}}
# identical across all 20 requests
```

### Issue 2: AutoNode `mode` is a Day-2 only operation in OCM

The OCM create API accepts `aws.auto_node.role_arn` but silently ignores `auto_node.mode`. AutoNode mode must be activated via a separate PATCH after the cluster is ready — equivalent to what `rosa edit cluster --autonode enabled` does.

## Related Issues and PRs

- Jira: N/A (community-reported)
- Fixes: N/A
- Related PR(s): #1048 ([OCM-22063](https://redhat.atlassian.net/browse/OCM-22063)) — similar fix for `KMSKeyArn`, `EtcdKmsKeyArn`, `AuditLogArn`, `ExternalAuthProvidersEnabled` which were missing the same nil-guard pattern
- Related design/docs: N/A

## Type of Change

- [ ] feat - adds a new user-facing capability.
- [x] fix - resolves an incorrect behavior or bug.

## Previous Behavior

Every `terraform apply` creating a ROSA HCP cluster with `auto_node` set fails with "Provider produced inconsistent result after apply". The cluster is created in OCM but never saved to Terraform state, leaving it orphaned. There is no workaround at the Terraform configuration level — `lifecycle { ignore_changes = [auto_node] }` does not suppress this error as it fires during Terraform core's post-apply consistency check.

## Behavior After This Change

`terraform apply` with `auto_node` succeeds end-to-end:

1. Cluster is created in OCM with `aws.auto_node.role_arn` set
2. Provider issues a post-create PATCH to activate `auto_node.mode` (OCM Day-2 requirement)
3. Provider reads the cluster back and correctly infers `mode = "enabled"` from the presence of `role_arn` in the GET response
4. Terraform state is saved cleanly with no inconsistency error
5. Subsequent `terraform plan` shows no drift

### Preconditions

- ROSA HCP cluster credentials and a VPC with subnets
- A Karpenter IRSA IAM role ARN for AutoNode
- Provider built locally from this branch

### Test Steps

1. Configure a ROSA HCP cluster resource with `auto_node`:

```hcl
resource "rhcs_cluster_rosa_hcp" "main" {
  name    = "my-autonode-cluster"
  # ... other required fields ...

  auto_node = {
    mode     = "enabled"
    role_arn = "<karpenter-irsa-role-arn>"
  }
}
```

2. Run `terraform apply`
3. Verify apply completes without error
4. Run `terraform plan` — confirm no drift is shown for `auto_node`
5. Verify in OCM that `auto_node.mode` is `"enabled"`:

```bash
ocm get /api/clusters_mgmt/v1/clusters/<id> | jq '{auto_node, aws_auto_node: .aws.auto_node}'
```

### Expected Results

- `terraform apply` exits 0 with no inconsistency error
- `terraform plan` after apply shows no changes
- OCM confirms `auto_node.mode = "enabled"` and `aws.auto_node.role_arn` is set

## Proof of the Fix

OCM GET response before and after the post-create PATCH:

```
# Immediately after cluster create (before PATCH):
{"top_level_auto_node":null,"aws_auto_node":{"role_arn":"arn:aws:iam::808082629126:role/cz-demo-autonode-operator-role"}}

# After post-create PATCH issued by provider:
{"top_level_auto_node":{"mode":"enabled"},"aws_auto_node":{"role_arn":"arn:aws:iam::808082629126:role/cz-demo-autonode-operator-role"}}
```

`terraform apply` output:

```
Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
# No "Provider produced inconsistent result" error
```

## Breaking Changes

- [x] No breaking changes

- [ ] Relevant Jira/GitHub issues and related PRs are linked.
- [ ] Tests were added/updated where appropriate.
- [ ] `make pre-push-checks` passes.
- [ ] `make fmt-check` passes.
- [ ] `make build` passes.

---

### Notes for reviewers

- The post-create PATCH is only issued when `state.AutoNode != nil` and `mode` is non-empty, so it's a no-op for clusters without `auto_node` configured.
- If OCM changes its GET API to return `auto_node.mode` in future, the inference logic is still safe — the `if autoNodeState.Mode.IsNull()` guard means an API-returned value takes precedence over the inferred one.
- Follow-up: consider whether `activeDeadlineSeconds` on the `installplan-approver` Helm hook job should be increased to handle slow OLM catalog startup on fresh clusters (separate issue).

## Summary by CodeRabbit

* **New Features**
  * AutoNode activation is applied as a follow-up cluster update after creation when configured; activation failures are reported as warnings and won't block creation.
* **Bug Fixes**
  * Creation now validates AutoNode settings: when AutoNode mode is set, creation must wait for completion.
  * Cluster state sync now preserves AutoNode when the API returns only the provider role ARN.

[Review Change Stack](https://app.coderabbit.ai/change-stack/terraform-redhat/terraform-provider-rhcs/pull/1162?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/provider/clusterrosa/hcp/resource.go
+++ b/provider/clusterrosa/hcp/resource.go
@@ -736,18 +736,12 @@ func createHcpClusterObject(ctx context.Context,
 		return nil, err
 	}
 
-	autoNodeRoleArn := optionalAutoNodeRoleARN(state.AutoNode)
-
 	if err := ocmClusterResource.CreateAWSBuilder(rosaTypes.Hcp, awsTags, ec2MetadataHttpTokens,
 		kmsKeyARN, etcdKmsKeyArn, auditLogArn,
 		isPrivate, awsAccountID, awsBillingAccountId, stsBuilder, awsSubnetIDs,
 		ingressHostedZoneId, route53RoleArn, internalCommunicationHostedZoneId, vpceRoleArn,
-		awsAdditionalComputeSecurityGroupIds, nil, nil, awsAdditionalAllowedPrincipals, autoNodeRoleArn); err != nil {
+		awsAdditionalComputeSecurityGroupIds, nil, nil, awsAdditionalAllowedPrincipals, nil); err != nil {
 		return nil, err
-	}
-
-	if autoNodeMode := optionalAutoNodeMode(state.AutoNode); autoNodeMode != nil {
-		builder.AutoNode(cmv1.NewClusterAutoNode().Mode(*autoNodeMode))
 	}
 
 	if !common.IsStringAttributeUnknownOrEmpty(state.BaseDNSDomain) {
@@ -949,6 +943,17 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		return
 	}
 
+	hasAutoNodeMode := state.AutoNode != nil && !common.IsStringAttributeUnknownOrEmpty(state.AutoNode.Mode)
+
+	if hasAutoNodeMode && !shouldWaitCreationComplete {
+		response.Diagnostics.AddError(
+			summary,
+			"When configuring auto_node.mode, wait_for_create_complete = true is required "+
+				"(AutoNode activation is a post-create operation)",
+		)
+		return
+	}
+
 	hasEtcdEncrpytion := common.BoolWithFalseDefault(state.EtcdEncryption)
 	hasEtcdKmsKeyArn := common.HasValue(state.EtcdKmsKeyArn) && state.EtcdKmsKeyArn.ValueString() != ""
 	if (!hasEtcdEncrpytion && hasEtcdKmsKeyArn) || (hasEtcdEncrpytion && !hasEtcdKmsKeyArn) {
@@ -1023,6 +1028,14 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 	}
 	object = add.Body()
 
+	// Capture the planned AutoNode values before populateRosaHcpClusterState overwrites them
+	var plannedAutoNodeMode string
+	var plannedAutoNodeRoleARN string
+	if state.AutoNode != nil {
+		plannedAutoNodeMode = state.AutoNode.Mode.ValueString()
+		plannedAutoNodeRoleARN = state.AutoNode.RoleARN.ValueString()
+	}
+
 	// Save initial state:
 	err = populateRosaHcpClusterState(ctx, object, state)
 	if err != nil {
@@ -1056,27 +1069,57 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 				response.Diagnostics.Append(diags...)
 				return
 			}
-		}
-		if shouldWaitComputeNodesComplete {
-			tflog.Info(ctx, "Waiting for standard compute nodes to get ready")
-			timeOut := common.OptionalInt64(state.MaxMachinePoolWaitTimeoutInMinutes)
-			timeOut, err = common.ValidateTimeout(timeOut, rosa.MaxMachinePoolWaitTimeoutInMinutes)
-			if err != nil {
-				response.Diagnostics.AddError(
-					"Waiting for cluster creation finished with error",
-					fmt.Sprintf("Waiting for cluster creation finished with the error %v", err),
-				)
+		} else {
+			if shouldWaitComputeNodesComplete {
+				tflog.Info(ctx, "Waiting for standard compute nodes to get ready")
+				timeOut := common.OptionalInt64(state.MaxMachinePoolWaitTimeoutInMinutes)
+				timeOut, err = common.ValidateTimeout(timeOut, rosa.MaxMachinePoolWaitTimeoutInMinutes)
+				if err != nil {
+					response.Diagnostics.AddError(
+						"Waiting for cluster creation finished with error",
+						fmt.Sprintf("Waiting for cluster creation finished with the error %v", err),
+					)
+				}
+				object, err = r.ClusterWait.WaitForStdComputeNodesToBeReady(ctx, object.ID(), *timeOut)
+				if err != nil {
+					response.Diagnostics.AddError(
+						"Waiting for std compute nodes completion finished with error",
+						fmt.Sprintf("Waiting for std compute nodes completion finished with the error %v", err),
+					)
+					if object == nil {
+						diags = response.State.Set(ctx, state)
+						response.Diagnostics.Append(diags...)
+						return
+					}
+				}
 			}
-			object, err = r.ClusterWait.WaitForStdComputeNodesToBeReady(ctx, object.ID(), *timeOut)
-			if err != nil {
-				response.Diagnostics.AddError(
-					"Waiting for std compute nodes completion finished with error",
-					fmt.Sprintf("Waiting for std compute nodes completion finished with the error %v", err),
-				)
-				if object == nil {
-					diags = response.State.Set(ctx, state)
-					response.Diagnostics.Append(diags...)
-					return
+
+			// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+			// silently ignores auto_node.mode. A follow-up PATCH is required to activate mode
+			// after the cluster is ready, mirroring what `rosa edit cluster --autonode enabled` does.
+			if plannedAutoNodeMode != "" {
+				tflog.Info(ctx, "Activating AutoNode mode via post-create PATCH (OCM Day-2 requirement)")
+				autoNodePatch, err := cmv1.NewCluster().
+					AutoNode(cmv1.NewClusterAutoNode().Mode(plannedAutoNodeMode)).
+					AWS(cmv1.NewAWS().AutoNode(cmv1.NewAwsAutoNode().RoleArn(plannedAutoNodeRoleARN))).
+					Build()
+				if err != nil {
+					response.Diagnostics.AddWarning(
+						"Can't build AutoNode patch",
+						fmt.Sprintf("Can't build AutoNode patch for cluster '%s': %v", state.ID.ValueString(), err),
+					)
+				} else {
+					update, err := r.ClusterCollection.Cluster(state.ID.ValueString()).Update().
+						Body(autoNodePatch).
+						SendContext(ctx)
+					if err != nil {
+						response.Diagnostics.AddWarning(
+							"Can't activate AutoNode mode",
+							fmt.Sprintf("Can't activate AutoNode mode for cluster '%s': %v", state.ID.ValueString(), err),
+						)
+					} else {
+						object = update.Body()
+					}
 				}
 			}
 		}
@@ -1893,20 +1936,35 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
 	}
 
 	autoNodeState := &AutoNode{
-		Mode:    types.StringNull(),
 		RoleARN: types.StringNull(),
 	}
+	hasMode := false
+	hasRoleARN := false
+
 	if autoNode, ok := object.GetAutoNode(); ok {
 		if mode, ok := autoNode.GetMode(); ok && mode == autoNodeModeEnabled {
 			autoNodeState.Mode = types.StringValue(mode)
+			hasMode = true
 		}
 	}
 	if awsAutoNode, ok := object.AWS().GetAutoNode(); ok {
 		if roleArn, ok := awsAutoNode.GetRoleArn(); ok && roleArn != "" {
 			autoNodeState.RoleARN = types.StringValue(roleArn)
+			hasRoleARN = true
 		}
 	}
-	if !autoNodeState.Mode.IsNull() && !autoNodeState.RoleARN.IsNull() {
+
+	// Only populate state.AutoNode if at least one of the fields is present
+	// Mode requires RoleARN to be meaningful, but RoleARN can exist alone
+	if hasMode {
+		// Mode must have RoleARN
+		if hasRoleARN {
+			state.AutoNode = autoNodeState
+		} else {
+			state.AutoNode = nil
+		}
+	} else if hasRoleARN {
+		// RoleARN alone is OK (mode can be added later)
 		state.AutoNode = autoNodeState
 	} else {
 		state.AutoNode = nil
--- a/provider/clusterrosa/hcp/resource.go
+++ b/provider/clusterrosa/hcp/resource.go
@@ -1031,6 +1031,7 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 	// Capture the planned AutoNode values before populateRosaHcpClusterState overwrites them
 	var plannedAutoNodeMode string
 	var plannedAutoNodeRoleARN string
+	autoNodePatchFailed := false
 	if state.AutoNode != nil {
 		plannedAutoNodeMode = state.AutoNode.Mode.ValueString()
 		plannedAutoNodeRoleARN = state.AutoNode.RoleARN.ValueString()
@@ -1108,6 +1109,7 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 						"Can't build AutoNode patch",
 						fmt.Sprintf("Can't build AutoNode patch for cluster '%s': %v", state.ID.ValueString(), err),
 					)
+					autoNodePatchFailed = true
 				} else {
 					update, err := r.ClusterCollection.Cluster(state.ID.ValueString()).Update().
 						Body(autoNodePatch).
@@ -1117,6 +1119,7 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 							"Can't activate AutoNode mode",
 							fmt.Sprintf("Can't activate AutoNode mode for cluster '%s': %v", state.ID.ValueString(), err),
 						)
+						autoNodePatchFailed = true
 					} else {
 						object = update.Body()
 					}
@@ -1135,6 +1138,16 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 			),
 		)
 		return
+	}
+
+	// If the AutoNode PATCH failed, the OCM API won't return mode in its response, which would
+	// cause Terraform to see a diff between plan (mode=enabled) and state (mode=null). Restore
+	// the planned values so the user sees the warning but state stays consistent with the plan.
+	if autoNodePatchFailed && plannedAutoNodeMode != "" {
+		state.AutoNode = &AutoNode{
+			Mode:    types.StringValue(plannedAutoNodeMode),
+			RoleARN: types.StringValue(plannedAutoNodeRoleARN),
+		}
 	}
 
 	// Fetch log forwarder IDs
--- a/provider/clusterrosa/hcp/resource_test.go
+++ b/provider/clusterrosa/hcp/resource_test.go
@@ -233,7 +233,7 @@ var _ = Describe("Rosa HCP Sts cluster", func() {
 
 			Expect(rosaClusterObject.FIPS()).To(BeTrue())
 		})
-		It("Sets AutoNode mode and AWS role ARN when provided", func() {
+		It("Does not set AutoNode mode or role ARN on create (Day-2 operation)", func() {
 			clusterState := generateBasicRosaHcpClusterState()
 			clusterState.AutoNode = &AutoNode{
 				Mode:    types.StringValue(autoNodeModeEnabled),
@@ -242,13 +242,13 @@ var _ = Describe("Rosa HCP Sts cluster", func() {
 			rosaClusterObject, err := createHcpClusterObject(context.Background(), clusterState, diag.Diagnostics{})
 			Expect(err).ToNot(HaveOccurred())
 
-			autoNode, ok := rosaClusterObject.GetAutoNode()
-			Expect(ok).To(BeTrue())
-			Expect(autoNode.Mode()).To(Equal(autoNodeModeEnabled))
+			// AutoNode mode and role_arn are not set in the create payload
+			// They are set via post-create PATCH after cluster is ready
+			_, ok := rosaClusterObject.GetAutoNode()
+			Expect(ok).To(BeFalse())
 
-			awsAutoNode, ok := rosaClusterObject.AWS().GetAutoNode()
-			Expect(ok).To(BeTrue())
-			Expect(awsAutoNode.RoleArn()).To(Equal(autoNodeRoleArn))
+			_, ok = rosaClusterObject.AWS().GetAutoNode()
+			Expect(ok).To(BeFalse())
 		})
 	})
 	It("Throws an error when version format is invalid", func() {
@@ -435,7 +435,7 @@ var _ = Describe("Rosa HCP Sts cluster", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(clusterState.AutoNode).To(BeNil())
 		})
-		It("Ignores auto_node when mode is missing", func() {
+		It("Populates role_arn without mode when only role_arn is present in API response", func() {
 			clusterState := &ClusterRosaHcpState{}
 			clusterJson := generateBasicRosaHcpClusterJson()
 			clusterJson["aws"].(map[string]interface{})["auto_node"] = map[string]interface{}{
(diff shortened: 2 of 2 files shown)


olucasfreitas:
@paulczar please squash the commits and fix the small issues coming up in the pipelines, I'll give his a review shortly

provider/clusterrosa/hcp/resource.go
```diff
@@ -1082,6 +1082,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+			SendContext(ctx)
+		if err != nil {
+			response.Diagnostics.AddError(
+				"Can't activate AutoNode mode",
+				fmt.Sprintf("Can't activate AutoNode mode for cluster '%s': %v", state.ID.ValueString(), err),
+			)
+			return
+		}
+		object = update.Body()
+	}
```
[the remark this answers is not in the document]
olucasfreitas: This looks like it still runs when wait_for_create_complete is false, even though AutoNode enablement needs the cluster to be ready first. It would be safer to only issue this PATCH after a successful ready wait.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1904,9 +1938,15 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
 	if awsAutoNode, ok := object.AWS().GetAutoNode(); ok {
 		if roleArn, ok := awsAutoNode.GetRoleArn(); ok && roleArn != "" {
 			autoNodeState.RoleARN = types.StringValue(roleArn)
+			// The OCM API does not return auto_node.mode in GET responses even when AutoNode
+			// is enabled. Since "enabled" is the only supported value, infer it from the
+			// presence of role_arn so the state block is not incorrectly nil-ified.
+			if autoNodeState.Mode.IsNull() {
+				autoNodeState.Mode = types.StringValue(autoNodeModeEnabled)
```
olucasfreitas: This seems to treat role_arn as enough to infer enabled, but the create path above suggests role_arn can exist before AutoNode is actually activated. That could leave state showing enabled when the Day-2 PATCH did not complete.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1082,6 +1082,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+		}
+		update, err := r.ClusterCollection.Cluster(state.ID.ValueString()).Update().
+			Body(autoNodePatch).
+			SendContext(ctx)
+		if err != nil {
+			response.Diagnostics.AddError(
+				"Can't activate AutoNode mode",
+				fmt.Sprintf("Can't activate AutoNode mode for cluster '%s': %v", state.ID.ValueString(), err),
+			)
+			return
```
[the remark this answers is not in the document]
olucasfreitas: If this PATCH build or send fails, Create exits before state is written even though the cluster was already created. It may be worth persisting state here so the resource does not get orphaned.

amandahla:
Hi @paulczar , thanks for submitting the PR.

Could you give more details about "The OCM create API accepts aws.auto_node.role_arn but silently ignores auto_node.mode. AutoNode mode must be activated via a separate PATCH after the cluster is ready" ?
Isnt something that could be fixed in the API instead of repetitive calls in the provider?

Also, please enable sign-off in the commit `git commit -s`

provider/clusterrosa/hcp/resource.go
```diff
@@ -1904,9 +1938,15 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
 	if awsAutoNode, ok := object.AWS().GetAutoNode(); ok {
 		if roleArn, ok := awsAutoNode.GetRoleArn(); ok && roleArn != "" {
 			autoNodeState.RoleARN = types.StringValue(roleArn)
+			// The OCM API does not return auto_node.mode in GET responses even when AutoNode
+			// is enabled. Since "enabled" is the only supported value, infer it from the
+			// presence of role_arn so the state block is not incorrectly nil-ified.
+			if autoNodeState.Mode.IsNull() {
```
amandahla: This if is not needed.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1082,6 +1082,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		}
 	}
 
+	// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+	// accepts role_arn under aws.auto_node but silently ignores auto_node.mode. A follow-up
+	// PATCH is required to activate mode after the cluster is ready, mirroring what
+	// `rosa edit cluster --autonode enabled` does.
+	if state.AutoNode != nil && !common.IsStringAttributeUnknownOrEmpty(state.AutoNode.Mode) {
```
amandahla: There is nothing to be done on the update side, right? If the arn is changed, for example.

provider/clusterrosa/hcp/resource.go
amandahla: @coderabbitai wdyt

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
paulczar: this is a good catch!  we should only allow setting autonode_true if "wait_for_create_complete=true". in fact we should erroe out if autonode=true and wait_for_create_complete=false.

provider/clusterrosa/hcp/resource.go
paulczar: good catch! I've updated to not infer

paulczar:
> Hi @paulczar , thanks for submitting the PR.
> 
> Could you give more details about "The OCM create API accepts aws.auto_node.role_arn but silently ignores auto_node.mode. AutoNode mode must be activated via a separate PATCH after the cluster is ready" ? Isnt something that could be fixed in the API instead of repetitive calls in the provider?
> 
> Also, please enable sign-off in the commit `git commit -s`

Currently its considered a day-2 only operation to enable auto-node in the API.   I disagree with that, but its not my call.     It certainly causes a quandry for us here in TF as we don't want users to have to do two TF runs and flip variables on the way through, so we have to do the patching.   I'm raising the "day-2" only nature of the feature to try and have that reconsidered at the same time as planning for the worst here in TF.

provider/clusterrosa/hcp/resource.go
paulczar: I believe I've addressed this in a followup commit to handle updates to both properly.

provider/clusterrosa/hcp/resource.go
```diff
@@ -949,6 +949,25 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+
+	if hasAutoNodeMode && !shouldWaitCreationComplete {
+		response.Diagnostics.AddError(
+			summary,
+			"When configuring auto_node.mode it is also required to set wait_for_create_complete = true (AutoNode activation is a post-create operation)",
+		)
+		return
+	}
+
+	if hasAutoNodeMode && !hasAutoNodeRoleARN {
```
robpblake: @paulczar I think it makes sense to validate that customers are setting the `mode` if they are providing the `role_arn`. Both are required to use autonode.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1080,6 +1112,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 				}
 			}
 		}
+
+		// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+		// accepts role_arn under aws.auto_node but silently ignores auto_node.mode. A follow-up
+		// PATCH is required to activate mode after the cluster is ready, mirroring what
+		// `rosa edit cluster --autonode enabled` does.
+		if plannedAutoNodeMode != "" {
+			tflog.Info(ctx, "Activating AutoNode mode via post-create PATCH (OCM Day-2 requirement)")
```
robpblake: @paulczar I would remove setting the `role_arn` as part of the original cluster payload and do everything here. So at this point given the validations, we know the customer will have supplied us a `role_arn` and set the `mode=enabled`, so we can supply both at the same time to the cluster PATCH. 

That keeps all our AutoNode flows in a single location at this time

provider/clusterrosa/hcp/resource.go
```diff
@@ -1896,17 +1974,33 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
...
 	if awsAutoNode, ok := object.AWS().GetAutoNode(); ok {
 		if roleArn, ok := awsAutoNode.GetRoleArn(); ok && roleArn != "" {
 			autoNodeState.RoleARN = types.StringValue(roleArn)
+			hasRoleARN = true
 		}
 	}
-	if !autoNodeState.Mode.IsNull() && !autoNodeState.RoleARN.IsNull() {
+
+	// Only populate state.AutoNode if at least one of the fields is present
+	// Mode requires RoleARN to be meaningful, but RoleARN can exist alone
```
robpblake, line 1971: @paulczar This is a bug in the backend CS implementation. Both `role_arn` and `mode` are required but we don't currently block supplying the `role_arn` as part of the day 1 call to create the cluster. I can't remember why PM decided it was a day 2 operation, but I can fix that bug which simplifies things here.

provider/clusterrosa/hcp/resource.go
amandahla: Both field are marked as required in the schema, is this check needed?

amandahla: @coderabbitai do your thing

provider/clusterrosa/hcp/resource.go
olucasfreitas: Even if the backend bug is fixed later, the provider still exposes that partial role_arn-only shape into Terraform state today. I would avoid sending role_arn on create and only write both fields together in the ready-time PATCH.

provider/clusterrosa/hcp/resource.go
olucasfreitas: This still feels like the right direction. The create payload is still sending role_arn on day 1, and that is what leaves the provider with a role_arn-only intermediate state on read.

provider/clusterrosa/hcp/resource.go
olucasfreitas: I do not think this is a runtime bug anymore. Both fields are already required in the auto_node schema, so Terraform should reject a partial block before Create runs.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1896,17 +1974,33 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
...
+	if hasMode {
+		// Mode must have RoleARN
+		if hasRoleARN {
+			state.AutoNode = autoNodeState
+		} else {
+			state.AutoNode = nil
+		}
+	} else if hasRoleARN {
+		// RoleARN alone is OK (mode can be added later)
```
olucasfreitas, line 1980: We still send role_arn in the day-1 create payload, then keep a role_arn-only auto_node state on read. That leaves an ambiguous partial state, maybe it would be cleaner to move role_arn into this post-ready PATCH and activate both fields together here.

provider/clusterrosa/hcp/resource.go
olucasfreitas: I think this create-time check is redundant for the schema case. The wait_for_create_complete validation still makes sense because that cross-field rule is not enforced by the nested schema.

amandahla:
@robpblake @paulczar Just a heads-up that our next release is scheduled for next week on June 3. Since we'll need to leave some time for testing before then, do you think we’ll be able to merge this PR in time? Let me know if there’s anything we can do to help move it forward.

robpblake:
> @robpblake @paulczar Just a heads-up that our next release is scheduled for next week on June 3. Since we'll need to leave some time for testing before then, do you think we’ll be able to merge this PR in time? Let me know if there’s anything we can do to help move it forward.

@amandahla I'm committed to getting this in ASAP. We need it in the next release.

provider/clusterrosa/hcp/resource.go
paulczar: 👍

provider/clusterrosa/hcp/resource.go
```diff
@@ -1023,6 +1027,19 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 	}
 	object = add.Body()
 
+	// Capture the planned AutoNode values before populateRosaHcpClusterState overwrites them
+	plannedAutoNode := state.AutoNode
+	var plannedAutoNodeMode string
+	var plannedAutoNodeRoleARN string
+	if plannedAutoNode != nil {
+		if !common.IsStringAttributeUnknownOrEmpty(plannedAutoNode.Mode) {
```
robpblake: @paulczar I think these if statements are technically unnecessary as the schema for the `auto_node` configuration makes them required if the customer supplies the `auto_node` part.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1080,6 +1097,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+
+		// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+		// accepts role_arn under aws.auto_node but silently ignores auto_node.mode. A follow-up
+		// PATCH is required to activate mode after the cluster is ready, mirroring what
+		// `rosa edit cluster --autonode enabled` does.
+		if plannedAutoNodeMode != "" {
+			tflog.Info(ctx, "Activating AutoNode mode via post-create PATCH (OCM Day-2 requirement)")
+			autoNodePatchBuilder := cmv1.NewCluster()
+			autoNodePatchBuilder.AutoNode(cmv1.NewClusterAutoNode().Mode(plannedAutoNodeMode))
+			if plannedAutoNodeRoleARN != "" {
```
robpblake: @paulczar The role arn should always be set as the schema marks the field as `required`

provider/clusterrosa/hcp/resource.go
```diff
@@ -1893,20 +1944,35 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
 	}
 
 	autoNodeState := &AutoNode{
-		Mode:    types.StringNull(),
```
robpblake, line 1896: @paulczar Is this removal correct?

provider/clusterrosa/hcp/resource.go
```diff
@@ -1893,20 +1944,35 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
...
 		if roleArn, ok := awsAutoNode.GetRoleArn(); ok && roleArn != "" {
 			autoNodeState.RoleARN = types.StringValue(roleArn)
+			hasRoleARN = true
 		}
 	}
-	if !autoNodeState.Mode.IsNull() && !autoNodeState.RoleARN.IsNull() {
+
+	// Only populate state.AutoNode if at least one of the fields is present
+	// Mode requires RoleARN to be meaningful, but RoleARN can exist alone
+	if hasMode {
```
robpblake, line 1972: @paulczar Do we need this if statement now that we are patching both fields after the cluster is ready?

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
olucasfreitas: There is still one gap here. If WaitForClusterToBeReady returns an error with a non-nil object, we still fall through and send the AutoNode PATCH, so I would gate this on a successful ready wait or an explicit Ready state check.

provider/clusterrosa/hcp/resource.go
```diff
@@ -1080,6 +1097,40 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 				}
 			}
 		}
+
+		// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+		// accepts role_arn under aws.auto_node but silently ignores auto_node.mode. A follow-up
+		// PATCH is required to activate mode after the cluster is ready, mirroring what
+		// `rosa edit cluster --autonode enabled` does.
+		if plannedAutoNodeMode != "" {
```
olucasfreitas: This still falls through to the AutoNode PATCH when the ready wait returns an error with a non-nil object. The PATCH should only run after a successful ready wait or an explicit Ready state check.

provider/clusterrosa/hcp/resource.go
```diff
@@ -949,6 +943,16 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		return
 	}
 
+	hasAutoNodeMode := state.AutoNode != nil && !common.IsStringAttributeUnknownOrEmpty(state.AutoNode.Mode)
+
+	if hasAutoNodeMode && !shouldWaitCreationComplete {
+		response.Diagnostics.AddError(
+			summary,
+			"When configuring auto_node.mode it is also required to set wait_for_create_complete = true (AutoNode activation is a post-create operation)",
```
robpblake: @paulczar This line is too long and is causing the pre-merge tests to fail. Needs to be less than 120 characters.

paulczar:
the two failing tests appear to be classic related which this PR doesn't touch

provider/clusterrosa/hcp/resource.go
```diff
@@ -1056,27 +1069,57 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
-		if shouldWaitComputeNodesComplete {
-			tflog.Info(ctx, "Waiting for standard compute nodes to get ready")
-			timeOut := common.OptionalInt64(state.MaxMachinePoolWaitTimeoutInMinutes)
-			timeOut, err = common.ValidateTimeout(timeOut, rosa.MaxMachinePoolWaitTimeoutInMinutes)
-			if err != nil {
-				response.Diagnostics.AddError(
-					"Waiting for cluster creation finished with error",
-					fmt.Sprintf("Waiting for cluster creation finished with the error %v", err),
-				)
+		} else {
```
amandahla, line 1073: Why this else?

provider/clusterrosa/hcp/resource.go
```diff
@@ -1056,27 +1069,57 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
+			// silently ignores auto_node.mode. A follow-up PATCH is required to activate mode
+			// after the cluster is ready, mirroring what `rosa edit cluster --autonode enabled` does.
+			if plannedAutoNodeMode != "" {
+				tflog.Info(ctx, "Activating AutoNode mode via post-create PATCH (OCM Day-2 requirement)")
+				autoNodePatch, err := cmv1.NewCluster().
+					AutoNode(cmv1.NewClusterAutoNode().Mode(plannedAutoNodeMode)).
+					AWS(cmv1.NewAWS().AutoNode(cmv1.NewAwsAutoNode().RoleArn(plannedAutoNodeRoleARN))).
+					Build()
+				if err != nil {
+					response.Diagnostics.AddWarning(
```
amandahla, line 1108: Is a warning enough? The configuration will be considered valid and will try to apply again during the update.
@coderabbitai

provider/clusterrosa/hcp/resource.go
```diff
@@ -1056,27 +1069,57 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
...
-					diags = response.State.Set(ctx, state)
-					response.Diagnostics.Append(diags...)
-					return
+
+			// AutoNode (auto_node.mode) is a Day-2 operation in the OCM API: the create request
+			// silently ignores auto_node.mode. A follow-up PATCH is required to activate mode
+			// after the cluster is ready, mirroring what `rosa edit cluster --autonode enabled` does.
+			if plannedAutoNodeMode != "" {
+				tflog.Info(ctx, "Activating AutoNode mode via post-create PATCH (OCM Day-2 requirement)")
+				autoNodePatch, err := cmv1.NewCluster().
```
amandahla, line 1103: Do we have a test that covers this patch?

provider/clusterrosa/hcp/resource.go
```diff
@@ -1893,20 +1936,35 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
...
-	if !autoNodeState.Mode.IsNull() && !autoNodeState.RoleARN.IsNull() {
+
+	// Only populate state.AutoNode if at least one of the fields is present
+	// Mode requires RoleARN to be meaningful, but RoleARN can exist alone
+	if hasMode {
+		// Mode must have RoleARN
+		if hasRoleARN {
+			state.AutoNode = autoNodeState
+		} else {
+			state.AutoNode = nil
```
amandahla, line 1977: Is nil safe to use instead of types.ObjectNull?

provider/clusterrosa/hcp/resource.go
robpblake: @coderabbitai Moving this to error rather than warning is better behaviour?

amandahla: As we discussed, this doesnt impact the user since the cluster is created and we can leave it as a warning.

amandahla: @coderabbitai you are right, populateRosaHcpClusterState overwritten the  values so we should restore them

olucasfreitas: The warning path still loses the planned auto_node values. If this PATCH fails, populateRosaHcpClusterState refreshes from the pre-patch object and can drop auto_node before State.Set, which brings back the same apply inconsistency.

amandahla:
@paulczar @robpblake I think this one can be closed in favor of https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1170 right?

provider/clusterrosa/hcp/resource.go
paulczar: The `else` was added based on feedback from @olucasfreitas — the compute node wait and the AutoNode PATCH should only run when `WaitForClusterToBeReady` succeeds. If the cluster wait fails but we still have a non-nil `object`, we fall through to populate state from the partial result and let Terraform handle it; running the PATCH against a cluster that isn't ready would just compound the failure. So the `else` gates both operations on a clean ready signal.

provider/clusterrosa/hcp/resource.go
paulczar: `AutoNode` is a Go pointer (`*AutoNode` in the state struct), so `nil` is the idiomatic representation of "not set" for a nested object — no need for `types.ObjectNull()`. `types.ObjectNull()` is for `types.Object` typed fields in the framework; here the struct embedding handles nullability through the pointer itself. The existing nil checks on `state.AutoNode` before accessing its fields cover the safety concern.

robpblake:
> @paulczar @robpblake I think this one can be closed in favor of #1170 right?

@amandahla Yes. It builds on top of @paulczar initial work and simplifies a little.

It is now merged, so I'll close this one.

cheers,

Rob

PR #1164, opened 2026-05-23
spike: OSD-on-GCP support in rhcs

naeemarsalan:
> This PR was written in part with the assistance of generative AI. All changes were validated against real OCM + GCP and pass the local pre-push hook. AI was not used to respond to review comments.

Spike. Not for merge yet.

## why

- rhcs has no OSD path today; no Terraform module on top of rhcs for OSD either
- this brings OSD-GCP in as first-class so rhcs is the single provider for every cluster type
- uplift is bounded: OCM endpoints are cluster-type-agnostic (the IdP work proved that), so it's schema + wiring

## what's new

- `rhcs_wif_config` resource + data source — `provider/wifconfig/`
- `rhcs_cluster_osd_gcp` — `provider/clusterosdgcp/`. Composes WIF + CMEK + PSC + Shared VPC + private API. Cluster-admin htpasswd user is folded inline like `rhcs_cluster_rosa_classic`
- `rhcs_gcp_machine_pool` — `provider/machinepool/gcp/`. Distinct TypeName mirroring `rhcs_hcp_machine_pool`
- runbook example at `examples/create_osd_gcp_cluster_full/`. GCP IAM ships as a bundled sub-module that could graduate into a `terraform-rhcs-osd-gcp` repo
- per-resource templates + `example_1.tf` for the docs Example Usage section, regenerated via `make docs`
- unit + internal regression tests on the new packages, Ginkgo-style like `provider/imagemirror/`. Internal tests pinned to the bugs caught during smoke testing

## what we tested

- `make pre-push-checks` green (build, fmt, lint, docs-lint, coverage, all 4 subsystem suites: classic 236/236, hcp 196/196, osdgcp 1/1, framework 0)
- `make unit-test` and `make e2e-unit-test` pass (Prow `ci/prow/unit` equivalents)
- `docker build .` succeeds (Prow `ci/prow/e2e-images` equivalent)
- provider loads via `dev_overrides`, `terraform validate` clean
- **full apply, real OCM + real GCP:** WIF config registered, GCP-side WIF IAM provisioned (KMS + workload identity pool + 7 SAs + 8 custom roles + ~30 IAM bindings), cluster created, cluster reached OCM `ready` state. API URL + console URL live; inline admin htpasswd IDP + cluster-admins group both confirmed via OCM
- **full destroy, real OCM + real GCP:** 68 resources destroyed cleanly, no manual `ocm delete` intervention needed
- **WIF in-use retry path validated live.** Captured `[INFO] WIF delete retry pending` log lines from the live destroy showing retry firing 4 times with linear backoff (15s / 30s / 45s / 60s) on OCM `CLUSTERS-MGMT-400 ... is in use and cannot be deleted` — then succeeded on attempt 5 once OCM released the cluster->wif reference

### bugs caught + fixed during smoke testing (now folded into the single commit)

| bug | fix |
|---|---|
| `marketplace-gcp` billing_model rejected by OCM for OSD Trial | `billing_model` now pure pass-through, OCM picks per-product default |
| Shared VPC pre-install needs host-project IAM bindings on WIF SAs | added `host_iam.tf` to the example with dynamic SA prefix filter |
| imported clusters break `admin_credentials` outputs (OCM doesn't return passwords on Read) | wrapped outputs in `try()` |
| `populateState` was nilling input-only fields (`gcp_encryption_key`, `gcp_network`) when OCM omitted them on Get | only assign when OCM returns non-nil |
| subsystem scaffold test asserted `BeZero()` on the whole RunOutput struct | assert `ExitCode` only |
| cluster/wif/machine_pool Delete treated 404 as failure (not idempotent), and WIF Delete raced cluster destroy with OCM's eventually-consistent reference cleanup | 404 → success in all three; wif Delete retries with linear backoff up to 12 attempts |
| no-op Update wrote plan into state (false-drift on immutable fields) | refresh from OCM, write the server view |

## gaps remaining before production

These are tracked in workspace docs (not in-tree).

1. **Subsystem CRUD coverage** — `subsystem/osdgcp/osdgcp_test.go` is scaffold-only. Need to mirror `subsystem/classic/cluster_resource_rosa_create_test.go` (~479 lines, hand-rolled mock OCM JSON).
2. **Cluster Create orphan-on-timeout** — when `WaitForClusterToBeReady` exceeds the wait_timeout (default 60min, hit twice in our smoke test), Terraform discards the resource but OCM keeps the cluster. Recovery needs `terraform import`. Fix: write minimal state with the cluster id before the wait.
3. **Three mock-based regression tests** still TODO — `TestClusterDelete_IdempotentOn404`, `TestWifConfigDelete_IdempotentOn404` + retry, `TestMachinePoolDelete_IdempotentOn404`. Need a fake OCM HTTP server.
4. **Real OCM jira ticket** to replace the `OSDGCP-1` placeholder in the commit subject.

## known limitations not in our control

- **OCM cluster delete does not propagate to most GCP cleanup.** Reproducible in both rhcs and mobb destroy cycles: OCM deletes the cluster cleanly but leaves the GCP compute instances, their boot disks, the private-API load balancer stack (forwarding rule, internal address, backend service, regional health check), and the cluster's private DNS managed zone in the service project. They count against quota indefinitely. Workaround during smoke runs was a REST-API sweep after destroy. Affects every TF provider that drives OSD-GCP via OCM; needs to be taken up with OCM.
- **GCP KMS keyrings cannot be deleted** (GCP design), and crypto keys go to DESTROY_SCHEDULED on destroy with a 24h delay. Re-applying with the same `cluster_name` after destroy hits 409 on the keyring and 400 on the key when the cluster's install tries to encrypt VM disks. Workaround: `terraform import` the keyring/key, then `cryptoKeyVersions/<n>:restore` + PATCH `state:ENABLED`. Will be addressed in the example with `lifecycle.prevent_destroy = true` or a random suffix in a follow-up.

## not changed

ROSA Classic + HCP code paths untouched. New resources are additive.

`AGENTS.md` says rhcs is ROSA-only by policy. Putting this up early so the scope conversation can happen before more time goes in.

## Summary by CodeRabbit

* **New Features**
  * Added GCP-focused Terraform support: WIF data source, WIF resource, OSD cluster on GCP resource, and GCP machine pool — with authentication options (WIF or service account), CMEK, Private Service Connect, Shared VPC, secure boot, admin credential creation, validations, imports, and lifecycle behavior.

* **Documentation**
  * Added comprehensive docs, templates, module README, and end-to-end examples showing multi-phase WIF setup, IAM wiring, CMEK, cluster provisioning, outputs, and usage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/terraform-redhat/terraform-provider-rhcs/pull/1164?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


shreyansvm:
@naeemarsalan : thanks for leading this effort. Read appreciate your contributions. 

Regarding the first bug, it is accurate that the GCP Marketplace is not a supported billing model for the OSD Trial. OSD Trial is a legacy trial offering based on annual subscriptions, and there is no way to convert an OSD Trial cluster to a GCP Marketplace billing model. Hence, customers who want to natively transact through the Marketplace (and utilize GCP committed discounts) should transact through the GCP Marketplace from the beginning.

PR #1174, opened 2026-06-01
ROSAENG-6793 | feat: read noproxy default domains from ocm api

marcolan018:

Current in production env, for zero egress enabled clusters, client can query the default domains in noproxy thru OCM API. This PR removes the hardcoded domain templates in TF provider, changes to read the default domain list from OCM API.

## Related Issues and PRs

- Jira: [ROSAENG-6793](https://redhat.atlassian.net/browse/ROSAENG-6793)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

### Test Steps
1. Create a hcp cluster with zero egress enabled. Related blocks in the main.tf as below
```
resource "rhcs_cluster_rosa_hcp" "rosa_sts_hcp_cluster" {
...
...
  properties = {
    rosa_creator_arn = data.aws_caller_identity.current.arn,
    zero_egress = "true"
  }
  proxy = {
    http_proxy = "http://test.com",
    no_proxy = "127.0.0.1"
  }
}
```
2. Run command `terraform apply` 
3.

### Expected Results

the `terraform apply` command should complete successfully

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [ ] Commit subject/title follows `[JIRA-TICKET] | [TYPE][(scope)][!]: <MESSAGE>`.
- [ ] PR description clearly explains both **what** changed and **why**.
- [ ] Relevant Jira/GitHub issues and related PRs are linked.
- [ ] Tests were added/updated where appropriate.
- [ ] I manually tested the change.
- [ ] `make pre-push-checks` passes.
- [ ] `make fmt-check` passes.
- [ ] `make build` passes.
- [ ] Any risk, limitation, or follow-up work is documented.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy handling for zero-egress clusters: user-defined proxy exclusion entries are preserved while automatic default domains are removed during create, update, and refresh/read flows; API-returned proxy values are stored raw and de-duplication is applied before state is persisted. Helper failures now emit warnings without overwriting existing proxy settings.

### DIFF
--- a/provider/clusterrosa/hcp/resource.go
+++ b/provider/clusterrosa/hcp/resource.go
@@ -1031,6 +1031,13 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 	plannedAutoNode := state.AutoNode
 	clusterReady := false
 
+	var plannedNoProxy types.String
+	if state.Proxy != nil {
+		plannedNoProxy = state.Proxy.NoProxy
+	} else {
+		plannedNoProxy = types.StringValue("")
+	}
+
 	// Save initial state:
 	err = populateRosaHcpClusterState(ctx, object, state)
 	if err != nil {
@@ -1149,6 +1156,16 @@ func (r *ClusterRosaHcpResource) Create(ctx context.Context, request resource.Cr
 		return
 	}
 
+	// Revise zero egress no proxy if needed
+	err = r.reviseZeroEgressNoProxy(ctx, object, state)
+	if err != nil {
+		response.Diagnostics.AddWarning(
+			"Can't revise zero egress no proxy",
+			fmt.Sprintf("Received error %v", err),
+		)
+		state.Proxy.NoProxy = plannedNoProxy
+	}
+
 	diags = response.State.Set(ctx, state)
 	response.Diagnostics.Append(diags...)
 }
@@ -1162,6 +1179,13 @@ func (r *ClusterRosaHcpResource) Read(ctx context.Context, request resource.Read
 	response.Diagnostics.Append(diags...)
 	if response.Diagnostics.HasError() {
 		return
+	}
+
+	var priorNoProxy types.String
+	if state.Proxy != nil {
+		priorNoProxy = state.Proxy.NoProxy
+	} else {
+		priorNoProxy = types.StringValue("")
 	}
 
 	// Find the cluster:
@@ -1206,6 +1230,16 @@ func (r *ClusterRosaHcpResource) Read(ctx context.Context, request resource.Read
 			fmt.Sprintf("Received error %v", err),
 		)
 		return
+	}
+
+	// Revise zero egress no proxy if needed
+	err = r.reviseZeroEgressNoProxy(ctx, object, state)
+	if err != nil {
+		response.Diagnostics.AddWarning(
+			"Can't revise zero egress no proxy",
+			fmt.Sprintf("Received error %v", err),
+		)
+		state.Proxy.NoProxy = priorNoProxy
 	}
 
 	diags = response.State.Set(ctx, state)
@@ -1568,6 +1602,13 @@ func (r *ClusterRosaHcpResource) Update(ctx context.Context, request resource.Up
 
 	object := update.Body()
 
+	var plannedNoProxy types.String
+	if plan.Proxy != nil {
+		plannedNoProxy = plan.Proxy.NoProxy
+	} else {
+		plannedNoProxy = types.StringValue("")
+	}
+
 	// Update the state:
 	err = populateRosaHcpClusterState(ctx, object, plan)
 	if err != nil {
@@ -1588,6 +1629,16 @@ func (r *ClusterRosaHcpResource) Update(ctx context.Context, request resource.Up
 			fmt.Sprintf("Received error %v", err),
 		)
 		return
+	}
+
+	// Revise zero egress no proxy if needed
+	err = r.reviseZeroEgressNoProxy(ctx, object, plan)
+	if err != nil {
+		response.Diagnostics.AddWarning(
+			"Can't revise zero egress no proxy",
+			fmt.Sprintf("Received error %v", err),
+		)
+		plan.Proxy.NoProxy = plannedNoProxy
 	}
 
 	diags = response.State.Set(ctx, plan)
@@ -2019,8 +2070,7 @@ func populateRosaHcpClusterState(ctx context.Context, object *cmv1.Cluster, stat
 
 		noProxy, ok := proxyObj.GetNoProxy()
 		if ok && noProxy != "" {
-			deDuplicatedNoProxy := proxy.RemoveNoProxyZeroEgressDefaultDomains(noProxy, ",")
-			state.Proxy.NoProxy = types.StringValue(deDuplicatedNoProxy)
+			state.Proxy.NoProxy = types.StringValue(noProxy)
 		}
 	} else {
 		// We cannot set the proxy to nil because the attribute state.Proxy.AdditionalTrustBundle might contain a value.
@@ -2200,6 +2250,47 @@ func (r *ClusterRosaHcpResource) waitTillClusterIsNotFoundWithTimeout(ctx contex
 	}
 
 	return false, nil
+}
+
+func (r *ClusterRosaHcpResource) reviseZeroEgressNoProxy(ctx context.Context,
+	object *cmv1.Cluster, state *ClusterRosaHcpState) error {
+	properties, ok := object.GetProperties()
+	if !ok {
+		return nil
+	}
+	zeroEgress, ok := properties["zero_egress"]
+	if !ok || zeroEgress != "true" {
+		return nil
+	}
+	proxyObj, ok := object.GetProxy()
+	if !ok {
+		return nil
+	}
+	noProxy, ok := proxyObj.GetNoProxy()
+	if !ok || noProxy == "" {
+		return nil
+	}
+	get, err := r.ClusterCollection.Cluster(object.ID()).Get().Parameter("fetch_service_inquiries", true).SendContext(ctx)
+	if err != nil {
+		return err
+	}
+	aws, ok := get.Body().GetAWS()
+	if !ok {
+		return fmt.Errorf("AWS configuration not found in cluster response")
+	}
+	zeroEgressConfig, ok := aws.GetZeroEgress()
+	if !ok {
+		return fmt.Errorf("zero egress configuration not found in cluster response")
+	}
+	defaultDomains := zeroEgressConfig.NoProxyDefaultDomains()
+	if len(defaultDomains) == 0 {
+		return nil
+	}
+	deDuplicatedNoProxy := proxy.RemoveNoProxyZeroEgressDefaultDomains(
+		noProxy, ",", defaultDomains, state.AWSAccountID.ValueString(),
+	)
+	state.Proxy.NoProxy = types.StringValue(deDuplicatedNoProxy)
+	return nil
 }
 
 func validateAutoNodeRoleARN() validator.String {
--- a/provider/proxy/utils.go
+++ b/provider/proxy/utils.go
@@ -4,37 +4,42 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"slices"
 	"strings"
 )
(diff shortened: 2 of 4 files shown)


provider/clusterrosa/hcp/resource.go
```diff
@@ -2202,6 +2231,34 @@ func (r *ClusterRosaHcpResource) waitTillClusterIsNotFoundWithTimeout(ctx contex
...
+	}
+	proxyObj, ok := object.GetProxy()
+	if !ok {
+		return nil
+	}
+	noProxy, ok := proxyObj.GetNoProxy()
+	if !ok || noProxy == "" {
+		return nil
+	}
+	get, err := r.ClusterCollection.Cluster(object.ID()).Get().Parameter("fetch_service_inquiries", true).SendContext(ctx)
```
olucasfreitas, line 2273: This extra `fetch_service_inquiries=true` GET is now on the success path of `Create()` and `Update()`. If it flakes after POST/PATCH succeeds, Terraform returns an error before state is written, which can orphan a just-created cluster.

provider/proxy/utils.go
```diff
@@ -1,40 +1,25 @@
...
-		}
+func RemoveNoProxyZeroEgressDefaultDomains(input string, separator string, defaultDomains []string) string {
+	domains := strings.Split(input, separator)
+	defaultDomainsMap := make(map[string]string)
+	for _, item := range defaultDomains {
+		defaultDomainsMap[item] = ""
 	}
-	return strings.Join(result, ",")
+	domains = slices.DeleteFunc(domains, func(item string) bool {
+		_, exists := defaultDomainsMap[item]
```
olucasfreitas: Switching from regex/suffix matching to exact string equality drops the old/new ECR compatibility the provider previously preserved. A service-inquiry entry like `.dkr.ecr.us-east-1.amazonaws.com` should still filter `123456789012.dkr.ecr.us-east-1.amazonaws.com`, but this map lookup won’t.

provider/clusterrosa/hcp/resource.go
marcolan018: That's a valid concern. Seems it's a design flaw in the current implementation. Some existing code has already updated the state using this way, such as autoNode, logforwarding.
imho, the risk is minor because the scenario is rare. Maybe we can create a new issue to track this?

provider/proxy/utils.go
marcolan018: In fact, the current change is the better one. The function should only remove the default domains injected by CS. If some domain like `123456789012.dkr.ecr.us-east-1.amazonaws.com` is not included in the default domains, but complies with the regex, that's an error which should be exposed

provider/proxy/utils.go
```diff
@@ -1,40 +1,25 @@
...
+		return exists
+	})
+	return strings.Join(domains, ",")
 }
```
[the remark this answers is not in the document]
olucasfreitas: I agree this should be fixed. If the helper keeps a generic `separator` argument, the join should use that same delimiter or the function contract is narrower than the signature suggests.

subsystem/hcp/cluster_resource_test.go
```diff
@@ -4535,6 +4535,139 @@ var _ = Describe("HCP Cluster", func() {
 				runOutput = Terraform.Apply()
 				Expect(runOutput.ExitCode).To(BeZero())
 			})
+
+			It("Filters zero-egress default domains from no_proxy during read", func() {
```
olucasfreitas: This spec is named as a `Read()` case, but the only handlers here are POST plus the helper's `fetch_service_inquiries=true` GET, so it actually exercises create-time filtering. Since the helper was added to `Create`, `Read`, and `Update`, I'd add at least one real refresh/update case so the suite matches the changed paths.

provider/proxy/utils.go
olucasfreitas: If no_proxy_default_domains is guaranteed to return the exact appended entries for every cluster, including older ones that may still carry account-prefixed ECR hosts, then exact match is the right behavior. Can we point to that CS/OCM contract explicitly here? Otherwise this looks like a regression from the backward compatibility PR #1103 added.

provider/clusterrosa/hcp/resource.go
olucasfreitas: I don't think the logforwarding/autoNode precedents are equivalent: fetchLogForwarderIds() is best-effort and the AutoNode patch only warns, while this extra GET is a hard error before State.Set. Even if the flake is rare, the resulting failure mode is a successful cluster operation with a failed Terraform apply/refresh, so I'd address it in this PR rather than defer it.

marcolan018: I changed the AddError to AddWarning for reviseZeroEgressNoProxy(). Now if the GET request failed, user will see the warning message and the terraform command will fail finally due to state/plan inconsistent error. But the cluster's state is persistent so no orphan ones. Users can refresh or destroy later.
does this look good? @olucasfreitas

provider/proxy/utils.go
marcolan018: good point. I added the code path to gurantee the backward compability

subsystem/hcp/cluster_resource_test.go
marcolan018: I added the subsystem test for `Read`

provider/clusterrosa/hcp/resource.go
olucasfreitas: The `AddWarning` change avoids the orphan-cluster risk, which is an improvement. One remaining concern: when the auxiliary GET fails, `populateRosaHcpClusterState` has already written the raw API `no_proxy` (with platform defaults appended) into state. The warning fires but `State.Set` still persists that unfiltered value, which can cause `Provider produced inconsistent result after apply` on the next plan -- the same drift ROSAENG-8051 described. Could we preserve the prior/planned `no_proxy` value on this failure path instead of letting the raw API value through?

provider/clusterrosa/hcp/resource.go
```diff
@@ -2202,6 +2228,34 @@ func (r *ClusterRosaHcpResource) waitTillClusterIsNotFoundWithTimeout(ctx contex
...
+	if err != nil {
+		return err
+	}
+	defaultDomains := get.Body().AWS().ZeroEgress().NoProxyDefaultDomains()
```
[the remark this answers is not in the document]
olucasfreitas: This is a real risk -- the SDK `AWS()` and `ZeroEgress()` accessors return nil when unset, so if the service-inquiry response is missing those fields, this chain panics rather than returning an error to the `AddWarning` path.

subsystem/hcp/cluster_resource_test.go
```diff
@@ -4535,6 +4535,353 @@ var _ = Describe("HCP Cluster", func() {
...
+				)
+
+				// Trigger refresh by running apply again (no changes expected)
+				runOutput = Terraform.Apply()
+				Expect(runOutput.ExitCode).To(BeZero())
+
+				resource := Terraform.Resource("rhcs_cluster_rosa_hcp", "my_cluster")
+				// Verify Read path also filters default domains correctly
+				Expect(resource).To(MatchJQ(`.attributes.proxy.no_proxy`, "custom.domain"))
+			})
```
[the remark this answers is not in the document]
olucasfreitas: Agree this would be good to add. The refresh test needs to be fixed first (it's currently failing at head), and an update case can follow.

subsystem/hcp/cluster_resource_test.go
olucasfreitas: This spec fails at the current head with a `RawQuery mismatch` -- the handler ordering during the refresh phase doesn't match the actual request sequence. The Read path issues a plain GET first, then the helper does the `fetch_service_inquiries=true` GET.

provider/clusterrosa/hcp/resource.go
marcolan018: added the logic to restore the value of no_proxy to the planned one

subsystem/hcp/cluster_resource_test.go
marcolan018: I cannot find the failed test.
To have a plain GET firstly, then have a `fetch_service_inquiries` GET, I think that's the desired flow.
Can you explain more? @olucasfreitas

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
marcolan018: fixed

provider/proxy/utils.go
olucasfreitas: This regex strips any account-prefixed ECR hostname unconditionally, not just CS-injected defaults. Low risk in practice, but worth a comment noting the trade-off since `no_proxy` can contain user-specified ECR endpoints.

provider/clusterrosa/hcp/resource.go
olucasfreitas: The `AddWarning` change fixes the orphan-resource case, but the fallback is still incomplete: when the prior/planned `no_proxy` is null, the raw API defaults still stay in state. I think the restore should be unconditional so the warning path also preserves an intentionally omitted `no_proxy`.

provider/proxy/utils.go
olucasfreitas: The backward-compat intent makes sense, but this regex is broader than the Jira/API contract because it strips any account-prefixed ECR hostname, not just a legacy CS default. Since OpenShift treats `no_proxy` as arbitrary destination domains, I’d narrow this to the specific legacy default mapping rather than a blanket regex.

subsystem/hcp/cluster_resource_test.go
olucasfreitas: I reran the focused scope on `809441d` and `subsystem/hcp` still fails in this spec with `RawQuery mismatch`. The plain cluster GET is arriving before the `fetch_service_inquiries=true` GET in the refresh phase, so the handler order still needs adjustment.

provider/clusterrosa/hcp/resource.go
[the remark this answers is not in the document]
olucasfreitas: Using `GetAWS()` and `GetZeroEgress()` makes the accessor chain nil-safe here, so the original panic concern looks addressed. I don’t see this thread as blocking anymore.

subsystem/hcp/cluster_resource_test.go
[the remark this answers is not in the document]
olucasfreitas: Create and refresh are covered now, but `Update()` still isn’t. Since the helper runs in `Update()` too, I’d keep this thread open until there’s one zero-egress PATCH case.

provider/proxy/utils.go
marcolan018: updated to only filter the domain with specific account prefix

provider/clusterrosa/hcp/resource.go
marcolan018: changed the restore to unconditional

subsystem/hcp/cluster_resource_test.go
marcolan018: @olucasfreitas can you provide the command or any other details for the failed test case? 
I tried to run all hcp subsystem test locally, no errors found

```
[1781065161] HCP - 199/199 specs ••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• SUCCESS! 1m4.12592125s PASS

Ginkgo ran 1 suite in 1m6.642772875s
Test Suite Passed
```

subsystem/hcp/cluster_resource_test.go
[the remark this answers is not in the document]
marcolan018: added the subsystem test for update path

subsystem/hcp/cluster_resource_test.go
olucasfreitas: Sorry, I was running this in the wrong commit, my bad @marcolan018

PR #1187, opened 2026-06-09
ROSAENG-9174 | feat(hcp): add component_routes support to HCP default ingress resource

reedcort:
## PR Summary

Add `component_routes` support to the `rhcs_hcp_default_ingress` resource, enabling ROSA HCP customers to configure custom hostnames and TLS certificates for Console and Downloads routes — matching feature parity with the Classic `rhcs_default_ingress` resource.

ROSA HCP customers cannot customize hostnames for Console or Downloads routes. This blocks customers migrating from ROSA Classic who rely on custom component routes, and those with strict enterprise domain/compliance requirements (e.g., GovCloud). The Classic default ingress resource already supports `component_routes`, but the HCP resource only exposed `listening_method`.

## Related Issues and PRs
- Jira: [ROSAENG-9174](https://redhat.atlassian.net/browse/ROSAENG-9174)
- Feature: [ROSA-62](https://issues.redhat.com/browse/ROSA-62)
- Clusters-service MR: 11393 (merged)
- HyperShift PR: [openshift/hypershift#8838](https://github.com/openshift/hypershift/pull/8838) (merged)
- HyperShift bug: [OCPBUGS-92791](https://redhat.atlassian.net/browse/OCPBUGS-92791)
- ROSA CLI PR: [openshift/rosa#3376](https://github.com/openshift/rosa/pull/3376) (merged)

## Type of Change

## Previous Behavior

The `rhcs_hcp_default_ingress` resource only supported the `listening_method` attribute. Customers could not configure `component_routes` for HCP clusters via Terraform. The field was silently ignored.

## Behavior After This Change

The `rhcs_hcp_default_ingress` resource now supports an optional `component_routes` map attribute for Console and Downloads routes (OAuth is not supported on HCP — the OAuth server runs on the management cluster). Customers can set custom hostnames and reference TLS secrets:

```hcl
resource "rhcs_hcp_default_ingress" "default" {
  cluster          = rhcs_cluster_rosa_hcp.example.id
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.company.com"
      tls_secret_ref = "console-tls"
    }
    downloads = {
      hostname       = "downloads.company.com"
      tls_secret_ref = "downloads-tls"
    }
  }
}
```

For HCP clusters, only the routes specified by the user are sent to the API (no `ResetComponentRoutes()`). This avoids sending an empty oauth route which the backend correctly rejects on HCP.

### Preconditions
- A running ROSA HCP cluster
- The `hypershift-component-routes` feature toggle enabled
- HyperShift operator >= v0.1.79 deployed
- A TLS secret created in `openshift-config` namespace on the hosted cluster
- ROSA CLI change merged ([openshift/rosa#3376](https://github.com/openshift/rosa/pull/3376))

### Test Steps
1. Configure the `rhcs_hcp_default_ingress` resource with `component_routes` for console and/or downloads
2. Run `terraform plan` — should show the component_routes as a new attribute
3. Run `terraform apply` — should update the ingress via the OCM API
4. Verify on the cluster: `oc get ingresses.config.openshift.io cluster -o jsonpath='{.spec.componentRoutes}'`
5. Verify custom routes: `oc get routes -n openshift-console`

### Expected Results
- `terraform plan` shows the component_routes diff correctly
- `terraform apply` succeeds
- The cluster's ingress config reflects the custom hostnames
- Custom routes are created (e.g. `console-custom`, `downloads-custom`)
- Original routes redirect (301) to the custom hostnames
- Setting an OAuth component route returns an error from the backend

## Proof of the Fix
- Verified end-to-end in integration on cluster `2ri09onv6bra4994h79m4eu2t2oh9spq`
- `terraform plan` correctly shows console and downloads component_routes
- `terraform apply` succeeds, componentRoutes propagate to hosted cluster ingress config
- Both `console-custom` and `downloads-custom` routes created on the hosted cluster
- Console accessible at custom hostname via curl

## Breaking Changes
- [x] No breaking changes

- [x] `make install-hooks` has been run in this clone.
- [x] `make build` passes.
- [x] Documentation was added/updated where appropriate.

## Summary by CodeRabbit

- **New Features**
  - Added optional `component_routes` support to the HCP default ingress resource for `console` and `downloads`.
  - Allows configuring `hostname` and `tls_secret_ref` per route, with validation requiring both values together.
  - Component routes are applied during create/update, and removed routes are reflected in state.

- **Documentation**
  - Documented the `component_routes` attribute schema and noted OAuth is not supported on HCP clusters.

- **Tests**
  - Added coverage for PATCH payloads and state changes when setting and clearing `component_routes`.

### DIFF
--- a/provider/defaultingress/hcp/resource.go
+++ b/provider/defaultingress/hcp/resource.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -18,15 +20,70 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common/attrvalidators"
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/defaultingress"
 )
 
 var validListeningMethods = []string{string(cmv1.ListeningMethodExternal), string(cmv1.ListeningMethodInternal)}
+
+var requiredHcpComponentRouteKeys = []string{"console", "downloads"}
+
+func validateComponentRoutes(ctx context.Context, req validator.MapRequest, resp *validator.MapResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+	elements := req.ConfigValue.Elements()
+	if len(elements) == 0 {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Empty component routes",
+			"component_routes must not be empty. Provide at least one route (console or downloads), or remove the attribute entirely.",
+		)
+		return
+	}
+	for key := range elements {
+		valid := false
+		for _, k := range requiredHcpComponentRouteKeys {
+			if key == k {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			resp.Diagnostics.AddAttributeError(
+				req.Path,
+				"Invalid component route key",
+				fmt.Sprintf("'%s' is not a valid component route key. Valid keys are: console, downloads", key),
+			)
+		}
+	}
+	for key, val := range elements {
+		obj, ok := val.(types.Object)
+		if !ok {
+			continue
+		}
+		var route defaultingress.ComponentRoute
+		d := obj.As(ctx, &route, basetypes.ObjectAsOptions{})
+		if d.HasError() {
+			continue
+		}
+		hostname := route.Hostname.ValueString()
+		tlsSecret := route.TlsSecretRef.ValueString()
+		if (hostname == "" && tlsSecret != "") || (hostname != "" && tlsSecret == "") {
+			resp.Diagnostics.AddAttributeError(
+				req.Path,
+				"Incomplete component route",
+				fmt.Sprintf("'%s' must have both hostname and tls_secret_ref, or both empty", key),
+			)
+		}
+	}
+}
 
 type DefaultIngressResource struct {
 	collection  *cmv1.ClustersClient
@@ -71,6 +128,20 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
 				Required:   true,
 				Validators: []validator.String{attrvalidators.EnumValueValidator(validListeningMethods)},
 			},
+			"component_routes": schema.MapAttribute{
+				Description: "Component route parameters for console and downloads. " +
+					"OAuth is not supported on HCP clusters.",
+				ElementType: basetypes.ObjectType{
+					AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+				},
+				Optional: true,
+				Validators: []validator.Map{
+					attrvalidators.NewMapValidator(
+						"component_routes must contain exactly 'console' and 'downloads' keys",
+						validateComponentRoutes,
+					),
+				},
+			},
 		},
 	}
 	return
@@ -113,7 +184,7 @@ func (r *DefaultIngressResource) Create(ctx context.Context, req resource.Create
 		)
 		return
 	}
-	err = r.updateIngress(ctx, nil, plan, plan.Cluster.ValueString(), r.collection)
+	err = r.updateIngress(ctx, nil, plan, plan.Cluster.ValueString(), r.collection, &resp.Diagnostics)
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Failed building cluster default ingress",
@@ -175,24 +246,20 @@ func (r *DefaultIngressResource) Update(ctx context.Context, req resource.Update
 	}
 
 	// assert cluster attribute wasn't changed:
-	common.ValidateStateAndPlanEquals(state.Cluster, plan.Cluster, "cluster", &diags)
+	common.ValidateStateAndPlanEquals(state.Cluster, plan.Cluster, "cluster", &resp.Diagnostics)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection)
+	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection, &resp.Diagnostics)
 	if err != nil {
-		diags.AddError(
+		resp.Diagnostics.AddError(
 			"Failed to update default ingress",
 			fmt.Sprintf(
 				"Cannot update default ingress for "+
 					"cluster '%s': %v", state.Cluster.ValueString(), err,
 			),
 		)
-	}
-
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
 		return
 	}
 
@@ -276,11 +343,30 @@ func (r *DefaultIngressResource) populateState(ingress *cmv1.Ingress, state *Def
 	state.Id = types.StringValue(ingress.ID())
 	state.ListeningMethod = types.StringValue(string(ingress.Listening()))
 
+	componentRoutes, ok := ingress.GetComponentRoutes()
+	if ok && len(componentRoutes) > 0 {
+		elements := map[string]attr.Value{}
+		for k, v := range componentRoutes {
+			elements[k] = defaultingress.FlattenComponentRoute(v.Hostname(), v.TlsSecretRef())
+		}
+		mapValue, diags := types.MapValue(types.ObjectType{
+			AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+		}, elements)
+		if diags != nil && diags.HasError() {
+			return fmt.Errorf("failed to convert component routes to MapType: %v", diags.Errors()[0].Detail())
+		}
+		state.ComponentRoutes = mapValue
+	} else {
+		state.ComponentRoutes = types.MapNull(types.ObjectType{
+			AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+		})
+	}
+
 	return nil
 }
 
 func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan *DefaultIngress,
-	clusterId string, clusterCollection *cmv1.ClustersClient) error {
+	clusterId string, clusterCollection *cmv1.ClustersClient, diags *diag.Diagnostics) error {
 
 	if state == nil {
 		state = &DefaultIngress{Cluster: plan.Cluster}
@@ -299,7 +385,7 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan 
 			plan = &DefaultIngress{}
 		}
 
-		ingressBuilder := getDefaultIngressBuilder(ctx, state, plan)
+		ingressBuilder := getDefaultIngressBuilder(ctx, state, plan, diags)
 
 		ingress, err := ingressBuilder.Build()
 		if err != nil {
@@ -320,10 +406,34 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan 
 	return nil
 }
 
-func getDefaultIngressBuilder(ctx context.Context, state, plan *DefaultIngress) *cmv1.IngressBuilder {
+func getDefaultIngressBuilder(
+	ctx context.Context, state, plan *DefaultIngress, diags *diag.Diagnostics,
+) *cmv1.IngressBuilder {
 	ingressBuilder := cmv1.NewIngress()
 	if !common.IsStringAttributeUnknownOrEmpty(plan.ListeningMethod) && state.ListeningMethod != plan.ListeningMethod {
 		ingressBuilder.Listening(cmv1.ListeningMethod(plan.ListeningMethod.ValueString()))
+	}
+	if !state.ComponentRoutes.Equal(plan.ComponentRoutes) {
+		componentRoutes := map[string]*cmv1.ComponentRouteBuilder{}
+		if !plan.ComponentRoutes.IsNull() {
+			for k, v := range plan.ComponentRoutes.Elements() {
+				componentRouteBuilder := cmv1.NewComponentRoute()
+				hostname, tlsSecretRef := defaultingress.ExpandComponentRoute(ctx, v.(types.Object), diags)
+				componentRouteBuilder.Hostname(hostname)
+				componentRouteBuilder.TlsSecretRef(tlsSecretRef)
+				componentRoutes[k] = componentRouteBuilder
+			}
+			for k := range state.ComponentRoutes.Elements() {
+				if _, exists := componentRoutes[k]; !exists {
+					componentRoutes[k] = cmv1.NewComponentRoute().Hostname("").TlsSecretRef("")
+				}
+			}
+		} else {
+			for k := range state.ComponentRoutes.Elements() {
+				componentRoutes[k] = cmv1.NewComponentRoute().Hostname("").TlsSecretRef("")
+			}
+		}
+		ingressBuilder.ComponentRoutes(componentRoutes)
 	}
 	return ingressBuilder
 }
--- a/provider/defaultingress/hcp/resource.go
+++ b/provider/defaultingress/hcp/resource.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -18,15 +20,81 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common"
 	"github.com/terraform-redhat/terraform-provider-rhcs/provider/common/attrvalidators"
+	"github.com/terraform-redhat/terraform-provider-rhcs/provider/defaultingress"
 )
 
 var validListeningMethods = []string{string(cmv1.ListeningMethodExternal), string(cmv1.ListeningMethodInternal)}
+
+var validHcpComponentRouteKeys = []string{"console", "downloads"}
+
+func resetHcpComponentRoutes() map[string]*cmv1.ComponentRouteBuilder {
+	componentRoutes := make(map[string]*cmv1.ComponentRouteBuilder, len(validHcpComponentRouteKeys))
+	for _, key := range validHcpComponentRouteKeys {
+		componentRoutes[key] = cmv1.NewComponentRoute().
+			Hostname("").
+			TlsSecretRef("")
+	}
+	return componentRoutes
+}
+
+func validateComponentRoutes(ctx context.Context, req validator.MapRequest, resp *validator.MapResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+	elements := req.ConfigValue.Elements()
+	if len(elements) == 0 {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Empty component routes",
+			"component_routes must not be empty. "+
+				"Provide at least one route (console or downloads), or remove the attribute entirely.",
+		)
+		return
+	}
+	for key := range elements {
+		valid := false
+		for _, k := range validHcpComponentRouteKeys {
+			if key == k {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			resp.Diagnostics.AddAttributeError(
+				req.Path,
+				"Invalid component route key",
+				fmt.Sprintf("'%s' is not a valid component route key. Valid keys are: console, downloads", key),
+			)
+		}
+	}
+	for key, val := range elements {
+		obj, ok := val.(types.Object)
+		if !ok {
+			continue
+		}
+		var route defaultingress.ComponentRoute
+		d := obj.As(ctx, &route, basetypes.ObjectAsOptions{})
+		if d.HasError() {
+			continue
+		}
+		hostname := route.Hostname.ValueString()
+		tlsSecret := route.TlsSecretRef.ValueString()
+		if (hostname == "" && tlsSecret != "") || (hostname != "" && tlsSecret == "") {
+			resp.Diagnostics.AddAttributeError(
+				req.Path,
+				"Incomplete component route",
+				fmt.Sprintf("'%s' must have both hostname and tls_secret_ref, or both empty", key),
+			)
+		}
+	}
+}
 
 type DefaultIngressResource struct {
 	collection  *cmv1.ClustersClient
@@ -71,6 +139,20 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
 				Required:   true,
 				Validators: []validator.String{attrvalidators.EnumValueValidator(validListeningMethods)},
 			},
+			"component_routes": schema.MapAttribute{
+				Description: "Component route parameters for console and downloads. " +
+					"OAuth is not supported on HCP clusters.",
+				ElementType: basetypes.ObjectType{
+					AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+				},
+				Optional: true,
+				Validators: []validator.Map{
+					attrvalidators.NewMapValidator(
+						"component_routes supports 'console' and 'downloads' keys",
+						validateComponentRoutes,
+					),
+				},
+			},
 		},
 	}
 	return
@@ -113,7 +195,7 @@ func (r *DefaultIngressResource) Create(ctx context.Context, req resource.Create
 		)
 		return
 	}
-	err = r.updateIngress(ctx, nil, plan, plan.Cluster.ValueString(), r.collection)
+	err = r.updateIngress(ctx, nil, plan, plan.Cluster.ValueString(), r.collection, &resp.Diagnostics)
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Failed building cluster default ingress",
@@ -175,24 +257,20 @@ func (r *DefaultIngressResource) Update(ctx context.Context, req resource.Update
 	}
 
 	// assert cluster attribute wasn't changed:
-	common.ValidateStateAndPlanEquals(state.Cluster, plan.Cluster, "cluster", &diags)
+	common.ValidateStateAndPlanEquals(state.Cluster, plan.Cluster, "cluster", &resp.Diagnostics)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection)
+	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection, &resp.Diagnostics)
 	if err != nil {
-		diags.AddError(
+		resp.Diagnostics.AddError(
 			"Failed to update default ingress",
 			fmt.Sprintf(
 				"Cannot update default ingress for "+
 					"cluster '%s': %v", state.Cluster.ValueString(), err,
 			),
 		)
-	}
-
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
 		return
 	}
 
@@ -276,11 +354,30 @@ func (r *DefaultIngressResource) populateState(ingress *cmv1.Ingress, state *Def
 	state.Id = types.StringValue(ingress.ID())
 	state.ListeningMethod = types.StringValue(string(ingress.Listening()))
 
+	componentRoutes, ok := ingress.GetComponentRoutes()
+	if ok && len(componentRoutes) > 0 {
+		elements := map[string]attr.Value{}
+		for k, v := range componentRoutes {
+			elements[k] = defaultingress.FlattenComponentRoute(v.Hostname(), v.TlsSecretRef())
+		}
+		mapValue, diags := types.MapValue(types.ObjectType{
+			AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+		}, elements)
+		if diags != nil && diags.HasError() {
+			return fmt.Errorf("failed to convert component routes to MapType: %v", diags.Errors()[0].Detail())
+		}
+		state.ComponentRoutes = mapValue
+	} else {
+		state.ComponentRoutes = types.MapNull(types.ObjectType{
+			AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+		})
+	}
+
 	return nil
 }
 
 func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan *DefaultIngress,
-	clusterId string, clusterCollection *cmv1.ClustersClient) error {
+	clusterId string, clusterCollection *cmv1.ClustersClient, diags *diag.Diagnostics) error {
 
 	if state == nil {
 		state = &DefaultIngress{Cluster: plan.Cluster}
@@ -299,7 +396,7 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan 
 			plan = &DefaultIngress{}
 		}
 
-		ingressBuilder := getDefaultIngressBuilder(ctx, state, plan)
+		ingressBuilder := getDefaultIngressBuilder(ctx, state, plan, diags)
 
 		ingress, err := ingressBuilder.Build()
 		if err != nil {
@@ -320,10 +417,23 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan 
 	return nil
 }
 
-func getDefaultIngressBuilder(ctx context.Context, state, plan *DefaultIngress) *cmv1.IngressBuilder {
+func getDefaultIngressBuilder(
+	ctx context.Context, state, plan *DefaultIngress, diags *diag.Diagnostics,
+) *cmv1.IngressBuilder {
 	ingressBuilder := cmv1.NewIngress()
 	if !common.IsStringAttributeUnknownOrEmpty(plan.ListeningMethod) && state.ListeningMethod != plan.ListeningMethod {
 		ingressBuilder.Listening(cmv1.ListeningMethod(plan.ListeningMethod.ValueString()))
+	}
+	if !state.ComponentRoutes.Equal(plan.ComponentRoutes) {
+		componentRoutes := resetHcpComponentRoutes()
+		for k, v := range plan.ComponentRoutes.Elements() {
+			componentRouteBuilder := cmv1.NewComponentRoute()
+			hostname, tlsSecretRef := defaultingress.ExpandComponentRoute(ctx, v.(types.Object), diags)
+			componentRouteBuilder.Hostname(hostname)
+			componentRouteBuilder.TlsSecretRef(tlsSecretRef)
+			componentRoutes[k] = componentRouteBuilder
+		}
+		ingressBuilder.ComponentRoutes(componentRoutes)
 	}
 	return ingressBuilder
 }
--- a/subsystem/hcp/default_ingress_resource_test.go
+++ b/subsystem/hcp/default_ingress_resource_test.go
@@ -116,6 +116,220 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
 		Expect(runOutput.ExitCode).To(BeZero())
 	})
 
+	It("When component_routes with console and downloads are set it should update successfully", func() {
+		TestServer.AppendHandlers(
+			CombineHandlers(
+				VerifyRequest(http.MethodGet, "/api/clusters_mgmt/v1/clusters/123"),
+				RespondWithJSON(http.StatusOK, clusterReady),
+			),
+			CombineHandlers(
+				VerifyRequest(http.MethodGet, "/api/clusters_mgmt/v1/clusters/123/ingresses"),
+				RespondWithJSON(http.StatusOK, defaultDay1Template),
+			),
+			CombineHandlers(
+				VerifyRequest(http.MethodPatch, "/api/clusters_mgmt/v1/clusters/123/ingresses/d6z2"),
+				VerifyJQ(".component_routes.console.hostname", "console.custom.example.com"),
+				VerifyJQ(".component_routes.console.tls_secret_ref", "console-tls-secret"),
+				VerifyJQ(".component_routes.downloads.hostname", "downloads.custom.example.com"),
+				VerifyJQ(".component_routes.downloads.tls_secret_ref", "downloads-tls-secret"),
+				RespondWithJSON(http.StatusOK, `
+				{
+					"kind": "Ingress",
+					"href": "/api/clusters_mgmt/v1/clusters/123/ingresses/d6z2",
+					"id": "d6z2",
+					"listening": "external",
+					"default": true,
+					"dns_name": "redhat.com",
+					"component_routes": {
+						"console": {
+							"hostname": "console.custom.example.com",
+							"tls_secret_ref": "console-tls-secret"
+						},
+						"downloads": {
+							"hostname": "downloads.custom.example.com",
+							"tls_secret_ref": "downloads-tls-secret"
+						}
+					}
+				}`),
+			),
+		)
+		Terraform.Source(`
+			resource "rhcs_hcp_default_ingress" "default_ingress" {
+				cluster = "123"
+				listening_method = "external"
+				component_routes = {
+					"console" = {
+						"hostname"       = "console.custom.example.com"
+						"tls_secret_ref" = "console-tls-secret"
+					}
+					"downloads" = {
+						"hostname"       = "downloads.custom.example.com"
+						"tls_secret_ref" = "downloads-tls-secret"
+					}
+				}
+			}`)
+		runOutput := Terraform.Apply()
+		Expect(runOutput.ExitCode).To(BeZero())
+
+		resource := Terraform.Resource("rhcs_hcp_default_ingress", "default_ingress")
+		Expect(resource).To(MatchJQ(".attributes.component_routes.console.hostname", "console.custom.example.com"))
+		Expect(resource).To(MatchJQ(".attributes.component_routes.console.tls_secret_ref", "console-tls-secret"))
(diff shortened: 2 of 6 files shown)


amandahla:
Hi @reedcort, thanks for submitting the PR.

[ROSAENG-60551](https://redhat.atlassian.net/browse/ROSAENG-60551) covers CLI and OCM UI support as well as Terraform. Are those pieces ready to merge alongside this?

Today `rosa edit ingress --component-routes` still rejects HCP clusters, so I'm concerned that landing the provider first could confuse users. We usually treat the ROSA CLI as the design source of truth and land features there before the provider.

reedcort:
@amandahla The CLI change is in progress now — I'll put this TF PR on hold until the CLI change lands.

For OCM UI support, we don't have plans to add that at this time.

olucasfreitas:
@reedcort @amandahla ROSA CLI PR openshift/rosa#3376 merged on 2026-07-15. The hold was released. CLI alignment is confirmed.

provider/defaultingress/hcp/resource.go
```diff
@@ -177,7 +188,7 @@ func (r *DefaultIngressResource) Update(ctx context.Context, req resource.Update
 		return
 	}
 
-	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection)
+	err := r.updateIngress(ctx, state, plan, plan.Cluster.ValueString(), r.collection, diags)
```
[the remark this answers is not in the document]
olucasfreitas: This was marked "✅ Addressed in commit 1760edf" but the code still passes `diag.Diagnostics` by value to `updateIngress` and `getDefaultIngressBuilder`. The function signatures have not changed to `*diag.Diagnostics`, so `ExpandComponentRoute` errors are still silently lost if the slice reallocates.

provider/defaultingress/hcp/resource.go
```diff
@@ -180,7 +191,7 @@ func (r *DefaultIngressResource) Update(ctx context.Context, req resource.Update
...
```
olucasfreitas: `ValidateStateAndPlanEquals` adds errors to `diags` via the pointer, but the next check is `resp.Diagnostics.HasError()` — those errors aren't in `resp.Diagnostics` yet, so validation is silently bypassed. Either check `diags.HasError()` or append to resp first.

provider/defaultingress/hcp/resource.go
```diff
@@ -71,6 +75,13 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
 				Required:   true,
 				Validators: []validator.String{attrvalidators.EnumValueValidator(validListeningMethods)},
 			},
+			"component_routes": schema.MapAttribute{
```
olucasfreitas, line 142: Consider adding a map key validator to restrict keys to `console` and `downloads` — currently any string is accepted and the error only comes from the API.

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,178 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
+							"hostname": "console.custom.example.com",
+							"tls_secret_ref": "console-tls-secret"
+						}
+					}
+				}`),
+			),
+		)
+
+		Terraform.Source("")
+		runOutput = Terraform.Apply()
```
olucasfreitas, line 462: This test only destroys the resource. Add a case that removes `component_routes` from the HCL while keeping the resource, and assert the PATCH sends an empty `component_routes` map — that's the clearing logic this PR specifically implements.

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,178 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
+						"tls_secret_ref" = "console-tls-secret"
+					}
+					"downloads" = {
+						"hostname"       = "downloads.custom.example.com"
+						"tls_secret_ref" = "downloads-tls-secret"
+					}
+				}
+			}`)
+		runOutput := Terraform.Apply()
+		Expect(runOutput.ExitCode).To(BeZero())
```
olucasfreitas, line 172: Consider also asserting on the resulting Terraform state (e.g., `Terraform.Resource(...).AttributeValue("component_routes.console.hostname")`) to catch `populateState` bugs.

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,203 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
+
+		resource := Terraform.Resource("rhcs_hcp_default_ingress", "default_ingress")
+		Expect(resource).To(MatchJQ(".attributes.component_routes.console.hostname", "console.custom.example.com"))
+		Expect(resource).To(MatchJQ(".attributes.component_routes.console.tls_secret_ref", "console-tls-secret"))
+		Expect(resource).To(MatchJQ(".attributes.component_routes.downloads.hostname", "downloads.custom.example.com"))
+		Expect(resource).To(MatchJQ(".attributes.component_routes.downloads.tls_secret_ref", "downloads-tls-secret"))
+	})
+
+	It("When component_routes with only console is set it should update successfully", func() {
```
amandahla, line 181: This test allows a single console route. ROSA CLI rejects that for HCP in "When only one route is provided it should fail with wrong count (cmd/edit/ingress/cmd_test.go)" and requires both console and downloads. Is the Terraform behavior intentionally more flexible, or should we align with the CLI and require both keys when component_routes is set?

@coderabbitai validate, I'm using https://github.com/openshift/rosa/pull/3376 as reference.

provider/defaultingress/hcp/resource.go
amandahla: seems addressed in line 91
@coderabbitai validate

provider/defaultingress/hcp/resource.go
```diff
@@ -71,6 +75,33 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
...
+				Description: "Component route parameters for console and downloads. OAuth is not supported on HCP clusters.",
```
amandahla: Should we also validate that each route has both hostname and tls_secret_ref (or both empty)? Right now a missing field may only fail later with a backend 400. WDYT?

@coderabbitai validate, I'm using the https://github.com/openshift/rosa/pull/3376 as reference, business rules/validations must be the same.

provider/defaultingress/hcp/resource.go
```diff
@@ -71,6 +75,33 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
...
+				ElementType: basetypes.ObjectType{
+					AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+				},
+				Optional: true,
```
amandahla, line 148: Classic rejects an empty component_routes map (validateDefaultIngress in provider/defaultingress/classic/resource.go). Should HCP do the same for component_routes = {}?

@coderabbitai validate, I'm using the openshift/rosa#3376 as reference, business rules/validations must be the same. I'm also comparing with Classic resource just in case there is something that must be the same for both clusters.

provider/defaultingress/hcp/resource.go
amandahla: @reedcort wdyt? what should be the validation?

provider/defaultingress/hcp/resource.go
```diff
@@ -320,10 +366,29 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan
...
+	ctx context.Context, state, plan *DefaultIngress, diags *diag.Diagnostics,
+) *cmv1.IngressBuilder {
 	ingressBuilder := cmv1.NewIngress()
 	if !common.IsStringAttributeUnknownOrEmpty(plan.ListeningMethod) && state.ListeningMethod != plan.ListeningMethod {
 		ingressBuilder.Listening(cmv1.ListeningMethod(plan.ListeningMethod.ValueString()))
 	}
+	if !state.ComponentRoutes.Equal(plan.ComponentRoutes) {
+		componentRoutes := map[string]*cmv1.ComponentRouteBuilder{}
+		if !plan.ComponentRoutes.IsNull() {
+			for k, v := range plan.ComponentRoutes.Elements() {
```
amandahla: Shrink map: apply with console + downloads, then apply with only console. Does downloads get cleared, or can it remain on the cluster?

Example
1)
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.example.com"
      tls_secret_ref = "console-tls"
    }
    downloads = {
      hostname       = "downloads.example.com"
      tls_secret_ref = "downloads-tls"
    }
  }
}
```

then
2)
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.example.com"
      tls_secret_ref = "console-tls"
    }
  }
}
```

@coderabbitai validate if this is what is happening here and compare with the classic implementation
the main point is: if something is removed, should it be cleared?

provider/defaultingress/hcp/resource.go
```diff
@@ -320,10 +366,29 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan
...
+				componentRouteBuilder := cmv1.NewComponentRoute()
+				hostname, tlsSecretRef := defaultingress.ExpandComponentRoute(ctx, v.(types.Object), *diags)
+				componentRouteBuilder.Hostname(hostname)
+				componentRouteBuilder.TlsSecretRef(tlsSecretRef)
+				componentRoutes[k] = componentRouteBuilder
+			}
+		} else {
+			for k := range state.ComponentRoutes.Elements() {
```
amandahla: Omit attribute: apply with component_routes, then remove the attribute. I expect that to mean “I don’t want this config anymore” and clear prior keys. Also: is component_routes = {} treated as clear, rejected, or a no-op?

example
1)
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.example.com"
      tls_secret_ref = "console-tls"
    }
    downloads = {
      hostname       = "downloads.example.com"
      tls_secret_ref = "downloads-tls"
    }
  }
}
```

2)
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"
}
```

@coderabbitai validate, compare with the classic implementation.

subsystem/hcp/default_ingress_resource_test.go
amandahla, line 181: Related to my previous comment about shrinking, wdyt about adding a test that applies both and then console only?
@coderabbitai evaluate, related to https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1187/changes#r3616356951

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,200 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
+						"console": {
+							"hostname": "console.custom.example.com",
+							"tls_secret_ref": "console-tls-secret"
+						}
+					}
+				}`),
+			),
+			CombineHandlers(
+				VerifyRequest(http.MethodPatch, "/api/clusters_mgmt/v1/clusters/123/ingresses/d6z2"),
+				RespondWithJSON(http.StatusOK, `
```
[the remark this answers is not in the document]
amandahla: @coderabbitai I think Cursor reviewed almost the same, confirm:
"In this clear scenario (When component_routes block is removed it should clear routes), the second PATCH is only checked with VerifyRequest — there is no assertion on the request body (e.g. empty hostname / tls_secret_ref for the previously managed keys).

Classic covers the equivalent with VerifyJSON on the clear PATCH in subsystem/classic/default_ingress_resource_test.go around the component_routes reset fully case (~line 979).

Could we add a VerifyJQ (or VerifyJSON) here so the test locks that omit → clear sends empty values, not only that state ends up null?"

subsystem/hcp/default_ingress_resource_test.go
amandahla: @reedcort wdyt?

provider/defaultingress/hcp/resource.go
amandahla: @reedcort could you confirm how the api handles this? what should be the behavior?

provider/defaultingress/hcp/resource.go
amandahla: @reedcort Could you align the behavior with the Classic resource? Is this expected I believe, right?

amandahla:
@reedcort please consider the comments related to difference between rosa CLI x Classic x HCP. I believe the CLI must be the source of truth.

@coderabbitai Please validate the summary below. Respond ONLY using the following format for each item, with no extra introductory or concluding text:
- [Item 1]: Confirmed / [Short correction if incorrect]
- [Item 2]: Confirmed / [Short correction if incorrect]

Summary:

- Single route vs both required — ROSA CLI (HCP) rejects one route and requires both console and downloads; this PR allows console-only (and tests that).
- hostname + tls_secret_ref as a pair — ROSA CLI validates both together; this PR does not plan-validate the pair (may only fail with a backend 400).
- Empty map component_routes = {} — Classic rejects it; this PR does not.
- Shrink map (drop a key) — Classic clears removed keys (reset + overlay; tested); this PR only PATCHes remaining plan keys — unclear if dropped routes stay on the cluster.
- Omit component_routes — Classic clears all routes on full reset (PATCH body asserted); this PR intends clear-on-omit, but clear PATCH body is not asserted in tests ({} clear vs reject vs no-op still open).
- oauth key — ROSA CLI / HCP reject oauth; this PR has a key allowlist (you noted that as addressed).
- Missing tests vs Classic — Classic has shrink + full-clear with VerifyJSON empty hostname/tls; this PR lacks shrink (both → console-only) and clear-body VerifyJQ/VerifyJSON.

subsystem/hcp/default_ingress_resource_test.go
amandahla: related https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1187#discussion_r3616376568

provider/defaultingress/hcp/resource.go
```diff
@@ -320,10 +366,29 @@ func (r *DefaultIngressResource) updateIngress(ctx context.Context, state, plan
...
 	ingressBuilder := cmv1.NewIngress()
 	if !common.IsStringAttributeUnknownOrEmpty(plan.ListeningMethod) && state.ListeningMethod != plan.ListeningMethod {
 		ingressBuilder.Listening(cmv1.ListeningMethod(plan.ListeningMethod.ValueString()))
 	}
+	if !state.ComponentRoutes.Equal(plan.ComponentRoutes) {
+		componentRoutes := map[string]*cmv1.ComponentRouteBuilder{}
+		if !plan.ComponentRoutes.IsNull() {
+			for k, v := range plan.ComponentRoutes.Elements() {
+				componentRouteBuilder := cmv1.NewComponentRoute()
+				hostname, tlsSecretRef := defaultingress.ExpandComponentRoute(ctx, v.(types.Object), *diags)
```
olucasfreitas: The HCP resource now passes `*diag.Diagnostics` down one level, but `ExpandComponentRoute()` still takes `diag.Diagnostics` by value, so decode diagnostics can still disappear on append. Please thread the pointer all the way into the shared helper and update both HCP and Classic callers.

subsystem/hcp/default_ingress_resource_test.go
olucasfreitas, line 181: The merged ROSA CLI contract for HCP requires exactly `console` + `downloads` when `component_routes` is set, while this PR accepts a console-only map and codifies that in tests. If the CLI is the source of truth, we should align the provider validation here rather than ship two different client contracts.

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,203 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
```
olucasfreitas, line 445: This now covers the state side of omit-to-clear, but the second PATCH still isn't asserting the serialized clear payload. Could we mirror the Classic test here and verify empty hostname/TLS values for the previously managed keys.

provider/defaultingress/hcp/resource.go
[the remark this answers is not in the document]
olucasfreitas: Still valid on `a8165d6`: the pointer now reaches `getDefaultIngressBuilder()`, but `ExpandComponentRoute()` still takes `diag.Diagnostics` by value, so appended decode diagnostics can still be lost.

provider/defaultingress/hcp/resource.go
olucasfreitas: Resolved in the current head. `ValidateStateAndPlanEquals` now writes directly to `resp.Diagnostics` and the update path exits immediately on error.

subsystem/hcp/default_ingress_resource_test.go
olucasfreitas: Validated against merged `openshift/rosa#3376`: the CLI requires the full two-route HCP set and rejects a one-route input. Unless we plan to relax the CLI in a sibling change, the provider should align here and drop the console-only success case.

provider/defaultingress/hcp/resource.go
olucasfreitas: I agree with rejecting `{}` here. Classic already tells users to omit the attribute to clear routes, and the HCP CLI path does not accept an empty route set either.

subsystem/hcp/default_ingress_resource_test.go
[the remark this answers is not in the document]
olucasfreitas: Agreed. The omit-to-clear path is implemented, but the second PATCH should assert the empty hostname/TLS values so the contract matches the Classic test instead of relying only on returned state.

subsystem/hcp/default_ingress_resource_test.go
reedcort: Aligned with the CLI — the validator now requires exactly both console and downloads keys. The console-only success test has been removed. Verified on live cluster.

provider/defaultingress/hcp/resource.go
reedcort: Added plan-time validation matching the ROSA CLI contract: both console and downloads required, empty map rejected, each route must have both hostname and tls_secret_ref (or both empty), oauth rejected. All verified on live cluster via terraform plan.

subsystem/hcp/default_ingress_resource_test.go
reedcort: Shrink (both → console-only) is now rejected by the validator since both keys are required. No shrink test needed — it's caught at plan time before any API call.

provider/defaultingress/hcp/resource.go
reedcort: The OCM API accepts partial updates — sending only console leaves downloads unchanged on the cluster. However, when I implemented the ROSA CLI change (openshift/rosa#3376), I followed the classic CLI framework which requires all routes to be set (classic: oauth+console+downloads, HCP: console+downloads). The provider validator follows the same contract.

The question is whether requiring both keys is the right approach, or if we should allow partial updates in both the CLI and Terraform. The API supports it, but the classic CLI has always required the full set. I'll confirm with my team that we want to require all routes to be set.

reedcort:
## Latest push addresses all feedback

**Changes:**
- `ExpandComponentRoute` → `*diag.Diagnostics` (shared helper + both HCP and classic callers)
- `ValidateStateAndPlanEquals` → `&resp.Diagnostics` directly
- Map key validator: requires both `console` + `downloads`, rejects oauth, empty map, and hostname/tls_secret_ref pair mismatch
- `.Equal()` instead of `reflect.DeepEqual` for component_routes comparison
- Permadiff fix for empty API response
- Clearing logic sends empty hostname/tls_secret_ref for prior routes
- Shrink protection clears removed keys
- Subsystem tests: happy path with `MatchJQ` state assertions, clearing test verifies PATCH sent + state nil

**Validation verified on live cluster (`terraform plan`):**

| Validation | Error |
|---|---|
| Console-only (missing downloads) | `Missing required component route: 'downloads'` |
| Empty map `{}` | `Empty component routes` |
| Hostname without tls_secret_ref | `Incomplete component route` |
| OAuth key | `Invalid component route key` |

**Subsystem test limitation:** The clearing PATCH body assertion (`VerifyJSON`) can't be tested in subsystem tests — the Terraform framework substitutes prior state into the plan for Optional attributes when the block is removed, so `state.ComponentRoutes.Equal(plan.ComponentRoutes)` returns true and the clearing path is skipped in the builder. Classic avoids this because its struct has many more differing fields. Clearing verified end-to-end on live cluster with `TF_LOG=TRACE` confirming the correct PATCH payload.

provider/defaultingress/hcp/resource.go
reedcort: Done. Both behaviors are now aligned with Classic.

reedcort:
After confirming with the team, we want to allow partial component route updates rather than requiring the full set. The CLI is being updated to match (openshift/rosa#3403). Once that merges, the provider validator here will also be updated to allow partial updates — removing the requirement for both console and downloads keys.

/hold

provider/defaultingress/hcp/resource.go
amandahla: The concept of "partial" for the terraform provider might be tricky. 

Lets see this scenario:
1)
User create the terraform configuration for the first time.
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.example.com"
      tls_secret_ref = "console-tls"
    }
  }
}
```
First request to the OCM API, sending "console".

2)
User change the terraform configuration and apply:
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    console = {
      hostname       = "console.example.com"
      tls_secret_ref = "console-tls"
    }
    downloads = {
      hostname       = "downloads.example.com"
      tls_secret_ref = "downloads-tls"
    }
  }
}
```
Second request to the OCM API, since partial updates are allowed, can send only "downloads".

3) What if the user removes console?
```
resource "rhcs_hcp_default_ingress" "default_ingress" {
  cluster          = "<cluster-id>"
  listening_method = "external"

  component_routes = {
    downloads = {
      hostname       = "downloads.example.com"
      tls_secret_ref = "downloads-tls"
    }
  }
}
```

Now, this part might be different from the CLI because terraform will reconcile the configuration with the real world. So, removing console, for example, means that the component_routes must contain only downloads. Its different than:
```
Set only console: rosa edit ingress -c <cluster> --component-routes 'console: hostname=console.example.com;tlsSecretRef=console-tls' <ingress-id>
Verify console route updated, downloads unchanged
Set only downloads: rosa edit ingress -c <cluster> --component-routes 'downloads: hostname=downloads.example.com;tlsSecretRef=downloads-tls' <ingress-id>
Verify downloads route updated, console unchanged
```
Makes sense?

IMHO a safer rule is: always send desired state for managed keys (or clear-all-managed + overlay plan). 

wdyt?

@coderabbitai validate and post a short summary.

Please check my comment above against:

- HashiCorp Terraform / Plugin Framework best practices — config as desired state for optional map attributes; what should happen when a key is removed from a map or the attribute is omitted after it was set (Update vs merge-style partial PATCH); whether “send only changed keys” is safe when the API merges by key.
- This repository’s existing patterns — especially Classic rhcs_default_ingress component_routes (provider/defaultingress/classic/resource.go, ResetComponentRoutes, validateDefaultIngress) and related subsystem tests under subsystem/classic/default_ingress_resource_test.go (reset single / reset fully). Compare to this PR’s HCP update path in provider/defaultingress/hcp/resource.go.
In the summary, explicitly answer:

- Is “partial add” OK but “remove key / omit attribute must clear” the right Terraform model?
- Does Classic already erase/reset then send the plan (preferred) vs diff-only PATCH?
- Does this PR match that, or diverge in a way that can leave stale map keys on the cluster?
- What concrete change (if any) would align HCP with provider norms?

Keep the summary concise (bullet points). Cite file paths/lines where relevant.

amandahla: @reedcort feel free to ping me on slack in case you want to discuss about it :)

reedcort:
/hold cancel

@amandahla @olucasfreitas The CLI partial updates PR (openshift/rosa#3403) has been merged. This PR is now aligned with the CLI contract:

- Partial updates allowed (console-only or downloads-only)
- Empty map `{}` rejected
- OAuth key rejected
- hostname/tls_secret_ref pair validation at plan time
- Clearing via block removal sends empty values for prior routes
- `ExpandComponentRoute` threaded with `*diag.Diagnostics` through shared helper and both HCP/classic callers

Ready for review.

provider/defaultingress/hcp/resource.go
```diff
@@ -71,6 +129,20 @@ func (r *DefaultIngressResource) Schema(ctx context.Context, req resource.Schema
...
+			"component_routes": schema.MapAttribute{
+				Description: "Component route parameters for console and downloads. " +
+					"OAuth is not supported on HCP clusters.",
+				ElementType: basetypes.ObjectType{
+					AttrTypes: defaultingress.ComponentRouteAttributeTypes,
+				},
+				Optional: true,
+				Validators: []validator.Map{
+					attrvalidators.NewMapValidator(
+						"component_routes must contain exactly 'console' and 'downloads' keys",
```
olucasfreitas: The validator body now allows partial HCP updates, but the summary string still says `component_routes must contain exactly 'console' and 'downloads' keys`. Could we update the summary text so plan-time errors match the new CLI/provider contract?

provider/defaultingress/hcp/resource.go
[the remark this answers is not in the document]
olucasfreitas: Resolved in `f4f250a`: `ExpandComponentRoute()` now accepts `*diag.Diagnostics`, and both HCP and Classic callers pass the pointer through.

provider/defaultingress/hcp/resource.go
olucasfreitas: Also addressed in the current head: `{}` is rejected, and incomplete hostname/tls_secret_ref pairs now fail during plan-time validation.

provider/defaultingress/hcp/resource.go
olucasfreitas: The builder now clears keys present in prior state but missing from the plan. I still think a sequential both→console subsystem case would be valuable so that path is covered explicitly.

subsystem/hcp/default_ingress_resource_test.go
```diff
@@ -116,6 +116,170 @@ var _ = Describe("rhcs_cluster_rosa_hcp - default ingress", func() {
...
+						"downloads": {
+							"hostname": "downloads.custom.example.com",
+							"tls_secret_ref": "downloads-tls-secret"
+						}
+					}
+				}`),
+			),
+			CombineHandlers(
+				VerifyRequest(http.MethodPatch, "/api/clusters_mgmt/v1/clusters/123/ingresses/d6z2"),
+				RespondWithJSON(http.StatusOK, `
```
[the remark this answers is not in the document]
olucasfreitas: I agree this is the remaining high-value test gap. The clear behavior is implemented, but the subsystem suite still does not assert the second PATCH payload the way the Classic test does.

provider/defaultingress/hcp/resource.go
amandahla: @coderabbitai validate if my concern described in this thread was addressed specially "the HCP implementation does not currently guarantee that downloads is cleared in the console + downloads → console transition." + terraform examples run as expected by a terraform user/sre.

amandahla:
Terraform config is desired state, not an imperative patch. 

If TF only sets console while downloads was set outside Terraform, a partial PATCH leaves downloads on the cluster and can miss it forever if it never enters state — silent drift.

Classic avoids that by resetting managed keys, then applying the plan. HCP should do the same for console/downloads. CLI partial/merge is fine for rosa edit; it is not the Terraform model unless we explicitly document merge-only (I’d rather not).

**See the issue?** Read mirrors the full component_routes map from OCM into state. So if downloads was set in the CLI and appears in the GET/PATCH body, the next plan should propose clearing it when config only has console — not permanent silent drift.
The real issue remains: partial update does not reconcile on the first apply. Config is not applied as **full** desired state in one shot; Classic’s reset-then-send avoids that. We should still follow Classic so one apply matches config.

@coderabbitai confirm if thats correct and show how to fix it.

provider/defaultingress/hcp/resource.go
amandahla: Closing this in favor of the comment posted on the PR

amandahla:
> Terraform config is desired state, not an imperative patch.
> 
> If TF only sets console while downloads was set outside Terraform, a partial PATCH leaves downloads on the cluster and can miss it forever if it never enters state — silent drift.
> 
> Classic avoids that by resetting managed keys, then applying the plan. HCP should do the same for console/downloads. CLI partial/merge is fine for rosa edit; it is not the Terraform model unless we explicitly document merge-only (I’d rather not).
> 
> **See the issue?** Read mirrors the full component_routes map from OCM into state. So if downloads was set in the CLI and appears in the GET/PATCH body, the next plan should propose clearing it when config only has console — not permanent silent drift. The real issue remains: partial update does not reconcile on the first apply. Config is not applied as **full** desired state in one shot; Classic’s reset-then-send avoids that. We should still follow Classic so one apply matches config.
> 
> @coderabbitai confirm if thats correct and show how to fix it.

@reedcort We should apply the same behavior as per Classic by always resetting managed keys and then applying the plan. This is the scenario that I want to prevent from happening:

**UPDATED!**

### Scenario: CLI sets `downloads`, then Terraform manages only `console`

**Setup**
1. User sets custom `downloads` via ROSA CLI / UI.
2. New Terraform config (no state yet) with only `console` in `component_routes`.

---

#### 1) First `terraform plan` (create)

No state yet → plan is create-from-config only. Existing CLI `downloads` is **not** shown.

```hcl
# rhcs_hcp_default_ingress.default_ingress will be created
+ resource "rhcs_hcp_default_ingress" "default_ingress" {
    + cluster          = "<cluster-id>"
    + listening_method = "external"
    + id               = (known after apply)

    + component_routes = {
        + "console" = {
            + hostname       = "console.example.com"
            + tls_secret_ref = "console-tls"
          }
      }
  }

Plan: 1 to add, 0 to change, 0 to destroy.
```

**UX gap:** plan does not indicate that `downloads` will be cleared or kept.

---

#### 2) First `terraform apply` (current code)

On create, `updateIngress` calls `populateDefaultIngress` before building the PATCH (no ingress id yet). That read is flattened into `state.ComponentRoutes`.

- If OCM returns CLI `downloads` → state already has `downloads` → first PATCH includes configured `console` **and** empty `downloads` (clear). **No second apply required** in that case.
- If OCM **omits** `downloads` from that initial read → first PATCH may send only `console` → CLI `downloads` can remain → later plans may stay clean (**silent drift**).

So the problem is not “only visible on a second update”; it is **misleading first plan** + **correctness depending on a complete initial API read**.

---

#### Ask

Please follow Classic reset-and-overlay for managed HCP keys (`console`, `downloads`), excluding `oauth`. Do not depend on whatever keys happened to be present in the first list/GET. Add create-path tests for (a) list returns pre-existing `downloads` and (b) list omits `downloads`, both asserting the first PATCH clears `downloads` via empty values.

@coderabbitai confirm and update the comment https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1187#issuecomment-5104494290 if needed.

amandahla:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== HCP resource outline ==='
> ast-grep outline provider/defaultingress/hcp/resource.go --items all || true
> 
> echo '=== Focused HCP create/update/builder/component-route code ==='
> rg -n -C 18 \
>   'func \(.*\) Create|func \(.*\) Update|func updateIngress|func getDefaultIngressBuilder|ComponentRoutes|component_routes|populateState|GetComponentRoutes|ExpandComponentRoute' \
>   provider/defaultingress/hcp/resource.go provider/defaultingress/hcp/state.go \
>   provider/defaultingress/component_routes.go || true
> 
> echo '=== Classic component-route reset/overlay implementation ==='
> rg -n -C 18 \
>   'func getDefaultIngressBuilder|ComponentRoutes|component_routes|ExpandComponentRoute|Hostname\(""\)|TlsSecretRef\(""\)' \
>   provider/defaultingress/classic/resource.go || true
> 
> echo '=== HCP component-route tests ==='
> ast-grep outline subsystem/hcp/default_ingress_resource_test.go --items all || true
> rg -n -C 12 \
>   'component_routes|ComponentRoutes|VerifyJSON|VerifyJQ|MatchJQ|PATCH|patch' \
>   subsystem/hcp/default_ingress_resource_test.go || true
> 
> echo '=== CodeRabbit comments on PR `#1187` ==='
> gh api --paginate \
>   repos/terraform-redhat/terraform-provider-rhcs/issues/1187/comments \
>   --jq '.[] | select(.user.login == "coderabbitai[bot]" or .user.login == "coderabbitai") | {id,url,body}' || true
> ```
> 
> Length of output: 65321
> 
> `@amandahla` Confirmed that the desired-state/reset model is preferable, but the stated first-apply sequence is not accurate for the current code.
> 
> On create, `updateIngress` calls `populateDefaultIngress` before building the PATCH when the resource has no ingress ID (Line 378). That list response is flattened into `state.ComponentRoutes` (Lines 347-359). Therefore, if it contains the CLI-created `downloads` route, with config containing only `console`, Lines 427-430 include:
> 
> ```json
> "downloads": {
>   "hostname": "",
>   "tls_secret_ref": ""
> }
> ```
> 
> in the **first** PATCH. The second apply described in the scenario is not required in that API-response case.
> 
> The remaining concern is valid: correctness currently depends on the initial API read containing every route. HCP should match Classic’s fixed reset-and-overlay behavior for its managed keys (`console`, `downloads`), while deliberately excluding unsupported `oauth`:
> 
> ```go
> func resetHcpComponentRoutes() map[string]*cmv1.ComponentRouteBuilder {
> 	componentRoutes := make(map[string]*cmv1.ComponentRouteBuilder, len(validHcpComponentRouteKeys))
> 	for _, key := range validHcpComponentRouteKeys {
> 		componentRoutes[key] = cmv1.NewComponentRoute().
> 			Hostname("").
> 			TlsSecretRef("")
> 	}
> 	return componentRoutes
> }
> ```
> 
> Then, in `getDefaultIngressBuilder`:
> 
> ```go
> if !state.ComponentRoutes.Equal(plan.ComponentRoutes) {
> 	componentRoutes := resetHcpComponentRoutes()
> 
> 	if !plan.ComponentRoutes.IsNull() {
> 		for key, value := range plan.ComponentRoutes.Elements() {
> 			hostname, tlsSecretRef := defaultingress.ExpandComponentRoute(
> 				ctx, value.(types.Object), diags,
> 			)
> 			componentRoutes[key] = cmv1.NewComponentRoute().
> 				Hostname(hostname).
> 				TlsSecretRef(tlsSecretRef)
> 		}
> 	}
> 
> 	ingressBuilder.ComponentRoutes(componentRoutes)
> }
> ```
> 
> Add a create-path regression test that returns a pre-existing `downloads` route from the initial ingress list, configures only `console`, and asserts the first PATCH includes the configured `console` plus empty `downloads` values. Also add a test where the list response omits `downloads`, asserting the same reset payload.
> 
> The prior comment is already consistent with this: it explicitly says the first-apply case is handled **when OCM returns the route**, while recommending the fixed-key reset because it does not depend on a complete API read. No update to comment `#issuecomment-5104494290` is needed.

Thanks @coderabbitai — agreed. I’ll correct the sequence: on create, populateDefaultIngress can load CLI downloads into state before the first PATCH, so clear can happen on first apply when the API returns that key.
My ask stands: please use Classic-style reset for managed HCP keys (console, downloads) then overlay the plan, so we don’t depend on a complete initial read. Happy with the suggested resetHcpComponentRoutes + create-path tests.

amandahla:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@amandahla` Confirmed. HCP `component_routes` should follow Classic’s desired-state model: initialize the supported managed keys (`console`, `downloads`) with empty `hostname` and `tls_secret_ref`, then overlay the configured Terraform values. `oauth` must remain excluded.
> 
> This guarantees that the first create-path PATCH clears `downloads` even when the initial OCM list/GET omits it, rather than making reconciliation depend on read completeness.
> 
> The earlier comment already states that the current implementation clears routes on the first apply **when** OCM returns them, while recommending a fixed-key reset to eliminate that dependency. No update to that comment is needed.

@coderabbitai validate the new changes regarding this discussion. Tell me if the PR is ready to be approved (mind documentation + tests + breaking changes).

amandahla:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== PR metadata and changed files ==='
> gh pr view 1187 --repo terraform-redhat/terraform-provider-rhcs \
>   --json title,state,isDraft,baseRefName,headRefName,mergeable,reviewDecision,files \
>   --jq '{title,state,isDraft,baseRefName,headRefName,mergeable,reviewDecision,files:[.files[].path]}'
> git diff --name-status origin/main...HEAD
> 
> echo '=== Focused current diff ==='
> git diff --find-renames origin/main...HEAD -- \
>   provider/defaultingress/hcp/resource.go \
>   provider/defaultingress/hcp/state.go \
>   provider/defaultingress/component_routes.go \
>   provider/defaultingress/classic/resource.go \
>   subsystem/hcp/default_ingress_resource_test.go \
>   docs/resources/hcp_default_ingress.md
> 
> echo '=== Implementation outlines ==='
> ast-grep outline provider/defaultingress/hcp/resource.go --items all
> ast-grep outline provider/defaultingress/component_routes.go --items all
> ast-grep outline subsystem/hcp/default_ingress_resource_test.go --items all
> 
> echo '=== HCP component-route implementation with context ==='
> rg -n -C 18 \
>   'validHcpComponentRouteKeys|component_routes|ComponentRoutes|resetHcpComponentRoutes|getDefaultIngressBuilder|updateIngress|populateDefaultIngress|ExpandComponentRoute|FlattenComponentRoute|Create\(|Update\(|Read\(' \
>   provider/defaultingress/hcp/resource.go \
>   provider/defaultingress/hcp/state.go \
>   provider/defaultingress/component_routes.go
> 
> echo '=== Classic reset/overlay reference ==='
> rg -n -C 16 \
>   'component_routes|ComponentRoutes|getDefaultIngressBuilder|Hostname\(""\)|TlsSecretRef\(""\)|ExpandComponentRoute' \
>   provider/defaultingress/classic/resource.go
> 
> echo '=== HCP component-route tests and request assertions ==='
> rg -n -C 16 \
>   'component_routes|ComponentRoutes|VerifyJSON|VerifyJQ|MatchJQ|PATCH|patch|downloads|console|hostname|tls_secret_ref' \
>   subsystem/hcp/default_ingress_resource_test.go
> 
> echo '=== HCP documentation ==='
> cat -n docs/resources/hcp_default_ingress.md
> 
> echo '=== Generated documentation/schema references ==='
> rg -n -C 5 \
>   'component_routes|tls_secret_ref|oauth|console|downloads' \
>   docs provider --glob '*.md' --glob '*.go'
> ```
> 
> Length of output: 143734
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== Fixture definitions used by HCP ingress subsystem tests ==='
> rg -n -C 16 \
>   'defaultDay1Template|component_routes|downloads|console' \
>   subsystem/hcp --glob '*.go'
> 
> echo '=== All HCP default-ingress tests, including validation and create-path coverage ==='
> rg -n -C 10 \
>   'It\(|DescribeTable|component_routes|Empty component routes|Incomplete component route|Invalid component route key|Terraform\.Plan|Terraform\.Apply|VerifyJQ|VerifyJSON' \
>   subsystem/hcp/default_ingress_resource_test.go provider/defaultingress/hcp --glob '*_test.go' --glob '*.go'
> 
> echo '=== Empty route object behavior and usages ==='
> rg -n -C 8 \
>   'FlattenComponentRoute\(|ObjectNull\(ComponentRouteAttributeTypes\)|both empty|ComponentRouteAttributeTypes' \
>   provider subsystem --glob '*.go'
> 
> echo '=== Test framework handler-consumption behavior ==='
> rg -n -C 8 \
>   'AppendHandlers|VerifyRequest|unconsumed|handlers.*remaining|Expect.*Handlers|Server.*handlers' \
>   subsystem --glob '*.go'
> 
> echo '=== PR checks and latest commits ==='
> gh pr view 1187 --repo terraform-redhat/terraform-provider-rhcs \
>   --json commits,statusCheckRollup,reviews \
>   --jq '{commits:[.commits[]|{oid,messageHeadline,committedDate}],checks:[.statusCheckRollup[]|{name,status,conclusion,detailsUrl}],reviews:[.reviews[]|{author:.author.login,state,submittedAt,body}]}'
> 
> echo '=== Provider versioning and release-note conventions for additive resource attributes ==='
> fd -t f 'CHANGELOG|changelog|release' . | xargs -r -n1 sh -c 'echo === $0; sed -n "1,160p" "$0"' 
> rg -n -C 5 'breaking change|breaking|component_routes|hcp_default_ingress' \
>   README.md docs .github --glob '*.md' --glob '*.yml' --glob '*.yaml' 2>/dev/null || true
> ```
> 
> Length of output: 202121
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== HCP ingress fixture and test setup ==='
> sed -n '1,125p' subsystem/hcp/default_ingress_resource_test.go
> 
> echo '=== Component-route test names and exact PATCH assertions ==='
> rg -n \
>   '^\s*It\(|VerifyJQ\("\.component_routes\.(console|downloads)\.(hostname|tls_secret_ref)"|Empty component routes|Incomplete component route|Invalid component route key' \
>   subsystem/hcp/default_ingress_resource_test.go
> 
> echo '=== All component-route validation tests in the repository ==='
> rg -n -C 4 \
>   'Empty component routes|Incomplete component route|Invalid component route key|component_routes.*(empty|oauth|tls_secret_ref|hostname)' \
>   provider subsystem --glob '*_test.go'
> 
> echo '=== Current required-check status ==='
> gh pr view 1187 --repo terraform-redhat/terraform-provider-rhcs \
>   --json statusCheckRollup \
>   --jq '.statusCheckRollup[] | select(.name != null) | {name,status,conclusion,detailsUrl}'
> ```
> 
> Length of output: 15631
> 
> `@amandahla` **not ready to approve yet.**
> 
> The agreed implementation is now correct in `provider/defaultingress/hcp/resource.go`: it resets only managed HCP keys (`console`, `downloads`), overlays configured values, and excludes `oauth`. This removes the dependency on the initial OCM read containing every route.
> 
> However, approval should wait for:
> 
> 1. **Create-path regression coverage**
>    
>    * The console-only create test uses an initial list response with _no_ `component_routes`, but it only asserts the configured `console` values. It must also assert:
>      ```json
>      "downloads": {
>        "hostname": "",
>        "tls_secret_ref": ""
>      }
>      ```
>    * Add the second requested create-path case: initial list response contains externally configured `downloads`; a console-only Terraform config must send the same first-PATCH clear payload.
>    * The existing shrink test covers update reconciliation, not either initial-create scenario.
> 2. **Validation coverage**
>    
>    * The validator implements empty-map rejection, `oauth` rejection, and hostname/TLS-pair validation, but there are no HCP tests asserting those plan errors. Add focused tests for `{}`, `oauth`, hostname-only, and TLS-secret-only configurations.
> 3. **Documentation and compatibility disclosure**
>    
>    * The generated docs mention supported keys and OAuth exclusion, but do not document the actual contract:
>      
>      * only `console` and `downloads` are supported;
>      * `component_routes = {}` is invalid;
>      * `hostname` and `tls_secret_ref` must be supplied together;
>      * omitting a managed key—or removing `component_routes`—clears that managed route.
>    * Although the schema addition is additive, the desired-state behavior is **behaviorally breaking** for existing `rhcs_hcp_default_ingress` users: after upgrade, a subsequent apply with `component_routes` omitted can clear out-of-band `console`/`downloads` custom routes. This needs an explicit breaking-change/migration note (and likely a changelog entry), so users can preserve routes by declaring them before upgrading/applying.
> 4. **Verification**
>    
>    * `Pre-push checks` are still running and Red Hat Konflux is queued, so validation is incomplete.
> 
> Once those items are addressed and the required checks pass, the PR should be ready for approval.

@reedcort I think mirroring existing classic tests would be fine:
```
Missing vs Classic (mirror these):

Omit → clear PATCH body — Classic component_routes reset fully (~906 / VerifyJSON empty values ~979). HCP omit test (~347) only checks state null; add empty hostname/tls_secret_ref asserts for console and downloads on the clear PATCH.
Reject null nested route — Classic fails to set single attribute as nil (~641).
Reject both fields empty — Classic fails to set single route as empty (~701).
```

amandahla:
/approve

@olucasfreitas leaving the lgtm to you so the PR wont be merged before you have the chance to check your comments

amandahla:
@reedcort my bad, ""downloads" = {}" is not a valid case here since component_routes is a MapAttribute (ObjectType), I  considered MapNestedAttribute. You can remove the test that is making the pre-push-checks fail.

PR #1209, opened 2026-06-23
ROSAENG-60113 | test(subsystem): add subsystem tests for uncovered datasources

amandahla:
## PR Summary
Add subsystem tests with mock OCM for `rhcs_info`, `rhcs_image_mirrors`, `rhcs_log_forwarders`, and `rhcs_rosa_hcp_operator_roles`.

Four data sources had no subsystem test coverage. Subsystem tests are the primary behavioral gate for provider types that talk to OCM. Without them, plan/apply/read paths for these types are not exercised in CI.

This PR is independent of #1205 (pre-merge gates and subsystem registry). It adds only test files on `main`.

## Related Issues and PRs
- Jira: [ROSAENG-60113](https://issues.redhat.com/browse/ROSAENG-60113)
- Related PR(s): #1205 (subsystem registry gate; separate follow-up)
- Related design/docs: N/A

## Type of Change

- [ ] feat - adds a new user-facing capability.
- [x] test - adds or updates tests only.

## Previous Behavior
No subsystem tests referenced `data "rhcs_info"`, `data "rhcs_image_mirrors"`, `data "rhcs_log_forwarders"`, or `data "rhcs_rosa_hcp_operator_roles"`.

## Behavior After This Change
Subsystem suites exercise the main read paths for those four data sources using mock OCM handlers. Provider behavior is unchanged.

### Preconditions
- Clone with hooks installed (`make install-hooks`)
- Go toolchain per `CONTRIBUTING.md`

### Test Steps
1. Run focused tests:

       go test -count=1 ./subsystem/classic -ginkgo.focus "Info data source"
       go test -count=1 ./subsystem/hcp -ginkgo.focus "Image mirrors data source|Log forwarders data source|ROSA HCP operator roles data source"

2. Run full pre-push checks:

       make pre-push-checks

### Expected Results
- All four new specs pass.
- `make pre-push-checks` completes with no failures.

## Proof of the Fix
- Logs/CLI output:
  - `make pre-push-checks`: 8/8 passed (format, build, generated files, lint, docs-lint, license headers, coverage-changed-files, unit + subsystem tests)
  - Focused `go test` runs for new specs: pass

## Breaking Changes
- [x] No breaking changes

### Breaking Change Details / Migration Plan
N/A

- [ ] I manually tested the change.
- [x] `make build` passes.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for the `rhcs_info` data source validating account and organization attributes.
  * Added test coverage for the `rhcs_image_mirrors` data source with cluster-specific mirror data validation.
  * Added test coverage for the `rhcs_log_forwarders` data source including error handling and log forwarder configuration validation.
  * Added test coverage for the `rhcs_rosa_hcp_operator_roles` data source with operator role and service account attribute validation.

### DIFF
--- a/subsystem/hcp/image_mirrors_data_source_test.go
+++ b/subsystem/hcp/image_mirrors_data_source_test.go
@@ -1,0 +1,56 @@
+// Copyright Red Hat
+// SPDX-License-Identifier: Apache-2.0
+
+package hcp
+
+import (
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2/dsl/core"             // nolint
+	. "github.com/onsi/gomega"                         // nolint
+	. "github.com/onsi/gomega/ghttp"                   // nolint
+	. "github.com/openshift-online/ocm-sdk-go/testing" // nolint
+
+	. "github.com/terraform-redhat/terraform-provider-rhcs/subsystem/framework"
+)
+
+var _ = Describe("Image mirrors data source", func() {
+	It("Can list image mirrors for a cluster", func() {
+		TestServer.AppendHandlers(
+			CombineHandlers(
+				VerifyRequest(http.MethodGet, "/api/clusters_mgmt/v1/clusters/123/image_mirrors"),
+				RespondWithJSON(http.StatusOK, `{
+				  "page": 1,
+				  "size": 1,
+				  "total": 1,
+				  "items": [
+				    {
+				      "id": "mirror-1",
+				      "type": "digest",
+				      "source": "registry.example.com/team",
+				      "mirrors": ["mirror.corp.com/team"],
+				      "creation_timestamp": "2024-01-01T00:00:00Z",
+				      "last_update_timestamp": "2024-01-02T00:00:00Z"
+				    }
+				  ]
+				}`),
+			),
+		)
+
+		Terraform.Source(`
+		  data "rhcs_image_mirrors" "cluster_mirrors" {
+		    cluster_id = "123"
+		  }
+		`)
+		runOutput := Terraform.Apply()
+		Expect(runOutput.ExitCode).To(BeZero())
+
+		resource := Terraform.Resource("rhcs_image_mirrors", "cluster_mirrors")
+		Expect(resource).To(MatchJQ(`.attributes.cluster_id`, "123"))
+		Expect(resource).To(MatchJQ(`.attributes.image_mirrors | length`, 1))
+		Expect(resource).To(MatchJQ(`.attributes.image_mirrors[0].id`, "mirror-1"))
+		Expect(resource).To(MatchJQ(`.attributes.image_mirrors[0].type`, "digest"))
+		Expect(resource).To(MatchJQ(`.attributes.image_mirrors[0].source`, "registry.example.com/team"))
+		Expect(resource).To(MatchJQ(`.attributes.image_mirrors[0].mirrors[0]`, "mirror.corp.com/team"))
+	})
+})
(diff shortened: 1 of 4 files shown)


subsystem/hcp/image_mirrors_data_source_test.go
```diff
@@ -0,0 +1,56 @@
...
+				      "type": "digest",
+				      "source": "registry.example.com/team",
+				      "mirrors": ["mirror.corp.com/team"],
+				      "creation_timestamp": "2024-01-01T00:00:00Z",
+				      "last_update_timestamp": "2024-01-02T00:00:00Z"
+				    }
+				  ]
+				}`),
+			),
+		)
```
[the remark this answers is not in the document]
olucasfreitas: Agree with this — the inconsistency within the same PR is the key signal. `log_forwarders_data_source_test.go` explicitly documents that data source reads run during plan and apply and double-registers the handler. Either that comment is wrong (and should be removed), or this test is under-registered and will flake under certain terraform-plugin-framework versions.

@amandahla can you align the pattern across the three new HCP data source tests? If single registration genuinely works here because the framework only reads once for this data source, then the comment in `log_forwarders` is misleading and should be removed instead.

subsystem/hcp/rosa_hcp_operator_roles_data_source_test.go
```diff
@@ -0,0 +1,122 @@
...
+			),
+		)
+
+		Terraform.Source(`
+		  data "rhcs_rosa_hcp_operator_roles" "operator_roles" {
+		    operator_role_prefix = "terraform-operator"
+		    account_role_prefix  = "TerraformAccountPrefix"
+		  }
+		`)
+		runOutput := Terraform.Apply()
```
olucasfreitas, line 69: Same question as image_mirrors: should this handler be registered twice per the pattern in `log_forwarders_data_source_test.go`? The `sts_credential_requests` endpoint only gets a single `AppendHandlers` call here — if the data source is read during both plan and apply phases, this will flake.

subsystem/hcp/image_mirrors_data_source_test.go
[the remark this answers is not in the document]
amandahla: @coderabbitai @olucasfreitas good catch but lets consider the docs: "Terraform attempts to query data sources during the planning phase, but depending on details in the configuration it may defer reading until the apply phase. When Terraform must defer reading data sources, it indicates it in the plan output."
https://developer.hashicorp.com/terraform/language/data-sources#data-source-behavior

Our subsystem tests use data-only configs with known arguments and no dependencies on managed resources, so Terraform reads during plan and doesn’t need a second read in apply. That’s why one handler is enough, the tests pass, and this matches the existing pattern (groups, cloud_providers, info, etc.).

log_forwarders is different: hcp_test.go registers a RouteToHandler for cluster 123 that returns an empty list, so any subsequent call to that endpoint would silently return wrong data. That’s why that test uses cluster 456 (and duplicates the handler as extra hardening).

subsystem/hcp/rosa_hcp_operator_roles_data_source_test.go
amandahla: I believe it can be solved by same explanation here:
https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1209/changes#r3460902528

olucasfreitas: Makes sense — same rationale as the image_mirrors explanation. No route conflict for `sts_credential_requests` and static config means single handler is correct.

PR #1273, opened 2026-07-22
OCM-00000 | refactor(logging): drop direct terraform-plugin-sdk/v2 dependency

somaz94:
## PR Summary

Removes the vestigial direct dependency on `github.com/hashicorp/terraform-plugin-sdk/v2`. The only remaining use was `helper/logging.LogLevel()` in `logging/logging.go`, which is reimplemented locally with identical semantics.

The provider is built on the terraform-plugin-framework, but still carried `terraform-plugin-sdk/v2` as a direct dependency solely for a single `LogLevel()` call that reads the `TF_LOG` environment variable. That call is trivial to reproduce locally, so the SDK dependency is unnecessary weight in the module graph.

## Related Issues and PRs
- Jira: OCM-00000 (no tracked ticket)
- Fixes: `#1127`
- Related PR(s): N/A
- Related design/docs: N/A

## Type of Change
- [x] refactor - code restructuring with no behavior change.

## Previous Behavior

`logging.New()` derived the log level from `terraform-plugin-sdk/v2/helper/logging.LogLevel()`, keeping the SDK as a direct dependency.

## Behavior After This Change

`logging.New()` uses a local `tfLogLevel()` helper. Behavior is unchanged: an unset `TF_LOG` yields an empty level, a recognized level is matched case-insensitively and upper-cased, and any other non-empty value falls back to `TRACE` (with the same warning log). `terraform-plugin-sdk/v2` and its now-unused transitive `hashicorp/logutils` are dropped from `go.mod` via `go mod tidy`.

### Preconditions
Go toolchain per `go.mod`.

### Test Steps
1. `make test` (runs the new `logging` package unit tests among the suite)
2. `go build ./...`
3. `grep -r "terraform-plugin-sdk/v2" --include="*.go" .` returns no import

### Expected Results
Tests pass, build succeeds, and the SDK is no longer imported anywhere or present in `go.mod`.

## Proof of the Fix
- Logs/CLI output: `make pre-push-checks` passes all 8 steps locally (format-check, build, generated-files, lint, docs-lint, license-check, subsystem-registry, unit+subsystem tests). The `logging` package had no tests before this PR and now has unit coverage for `tfLogLevel()` and `New()` level mapping.

## Breaking Changes
- [x] No breaking changes

- [x] `make install-hooks` has been run in this clone.
- [x] Documentation was added/updated where appropriate. (N/A, no user-facing docs affected)

### Testing (check all that apply; use N/A when not relevant)
- [x] **New or changed validation, plan modifiers, or helpers** — unit tests added in the same package (`logging/logging_test.go`).
- [x] `make check-subsystem-registry` passes.
- [x] I manually tested the change when behavior is user-visible. (N/A, no user-visible behavior change; verified via unit tests and `TF_LOG` level mapping.)

## Summary by CodeRabbit

- **New Features**
  - Logging configuration now reads the `TF_LOG` environment variable directly.
  - Supports case-insensitive `TRACE`, `DEBUG`, `INFO`, `WARN`, and `ERROR` levels.
  - Invalid values display a warning and default to the most detailed logging level.

- **Bug Fixes**
  - Logging behavior is now consistent for unset, valid, and invalid `TF_LOG` values.

- **Tests**
  - Added coverage for log-level normalization and logger configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ unit-test: $(GINKGO)
 	$(GINKGO) run \
 		--succinct \
 		-ldflags="$(ldflags)" \
-		-r provider internal/...
+		-r provider internal/... logging
 
 # Optional local diagnostics — not run by pre-push-checks or CI.
 .PHONY: unit-test-coverage
@@ -138,7 +138,7 @@ unit-test-coverage: $(GINKGO)
 		--cover \
 		--coverprofile coverage.out \
 		-ldflags="$(ldflags)" \
-		-r provider internal/...
+		-r provider internal/... logging
 
 
 .PHONY: test tests
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-validators v0.19.0
 	github.com/hashicorp/terraform-plugin-go v0.31.0
 	github.com/hashicorp/terraform-plugin-log v0.10.0
-	github.com/hashicorp/terraform-plugin-sdk/v2 v2.40.1
 	github.com/onsi/ginkgo/v2 v2.32.0
 	github.com/onsi/gomega v1.42.1
 	github.com/openshift-online/ocm-common v0.0.44
@@ -115,7 +114,6 @@ require (
 	github.com/hashicorp/go-plugin v1.7.0 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
 	github.com/hashicorp/hc-install v0.9.4 // indirect
-	github.com/hashicorp/logutils v1.0.0 // indirect
 	github.com/hashicorp/terraform-exec v0.25.1 // indirect
 	github.com/hashicorp/terraform-json v0.27.3-0.20260213134036-298b8f6b673a // indirect
 	github.com/hashicorp/terraform-registry-address v0.4.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -178,8 +178,6 @@ github.com/hashicorp/hc-install v0.9.4 h1:KKWOpUG0EqIV63Qk2GGFrZ0s275NVs5lKf9N5v
 github.com/hashicorp/hc-install v0.9.4/go.mod h1:4LRYeEN2bMIFfIv57ldMWt9awfuZhvpbRt0vWmv51WU=
 github.com/hashicorp/hcl/v2 v2.24.0 h1:2QJdZ454DSsYGoaE6QheQZjtKZSUs9Nh2izTWiwQxvE=
 github.com/hashicorp/hcl/v2 v2.24.0/go.mod h1:oGoO1FIQYfn/AgyOhlg9qLC6/nOJPX3qGbkZpYAcqfM=
-github.com/hashicorp/logutils v1.0.0 h1:dLEQVugN8vlakKOUE3ihGLTZJRB4j+M2cdTm/ORI65Y=
-github.com/hashicorp/logutils v1.0.0/go.mod h1:QIAnNjmIWmVIIkWDTG1z5v++HQmx9WQRO+LraFDTW64=
 github.com/hashicorp/terraform-exec v0.25.1 h1:PRutYRGM8pixV3B8812NYoBK5O+yuf3qcB/70KFKGiU=
 github.com/hashicorp/terraform-exec v0.25.1/go.mod h1:+izOYrs9sKMQK4OYvGDnrSSJHY/pm4e4eXFqSL2Q5mA=
 github.com/hashicorp/terraform-json v0.27.3-0.20260213134036-298b8f6b673a h1:T7AMR21kjrbeEpN+KhGlyd31XXHsSZF5zg+ivfeYte4=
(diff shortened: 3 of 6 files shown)


amandahla:
@coderabbitai review this PR based on the following criteria:

1. **Issue Resolution:** Verify that this PR correctly fixes the issue outlined in https://github.com/terraform-redhat/terraform-provider-rhcs/issues/1127.
2. **Dependency Cleanup:** Ensure any removed dependencies do not break or negatively impact existing code.
3. **Linter Directives:** Check if all `// nolint` annotations are strictly necessary, or if the underlying code can be refactored to satisfy the linter without suppressing warnings.
4. **Test Coverage:** Confirm unit cover both happy path and edge-case branches.
5. **Code Complexity:** Flag any deeply nested `if` statements or loops (`for`), suggesting guard clauses or early returns instead.
6. **Idiomatic Go (1.26):** Ensure code follows modern Go standards, standard library features, and idioms.

**Formatting Rules:**
- Add inline comments where relevant.
- Keep all comments concise, direct, and actionable.

somaz94:
Thanks for the review @amandahla — both of CodeRabbit's findings were valid. Fixed in `9658e4e`.

**P2 — the new suite was not running.** `make unit-test` is `ginkgo run -r provider internal/...`, an explicit package list, so `logging` was never picked up. Added it to both `unit-test` and `unit-test-coverage`.

Worth noting where the gap actually was: `LINT_TARGETS` already contained `./logging/...`, so the package was being linted but not tested — the asymmetry was only on the test side.

Measured before and after:

```
# before — -r provider internal/...
Ginkgo ran 18 suites in 13.646504875s     # no Logging Suite

# after  — -r provider internal/... logging
Logging Suite - 17/17 specs SUCCESS
Ginkgo ran 19 suites in 34.217616083s
Test Suite Passed
```

**P3 — dead `// nolint` directives.** Removed from `logging/main_test.go`. `.golangci.yml` sets `tests: false`, so they were suppressing nothing. `make lint` still reports `0 issues`.

The push re-triggered the full presubmit set on the new head, including `govulncheck`, which was the only red check before (and is `Required: false`). Local `make pre-push-checks` passes all 8 steps.

amandahla:
/approve
/lgtm
thanks for submitting the PR!

somaz94:
@amandahla Thanks for the `/ok-to-test`. The four rosa e2e jobs failed, but I don't think it's the PR — all four die in `day1-prepare`, before any test code runs, with the same auth error:

```
Get "https://api.stage.openshift.com/api/clusters_mgmt/v1/versions?...":
    can't get access token: invalid_grant: Offline user session not found
```

at `tests/utils/cms/versions.go:54`, each after ~11m. That's the OCM offline token in the CI credential having expired or been revoked, so a `/retest` would just re-run against the same dead token — I've held off for that reason.

Could someone with access rotate the OCM offline token for the presubmits? Happy to `/retest` once it's refreshed.

For reference this PR only drops the direct `terraform-plugin-log` dependency in the logging package; it doesn't touch the e2e paths.

jerichokeyne:
/retest

It should be fixed now. Seems like there was some issue with the environment yesterday, but some other CI jobs have been passing today so it seems to be resolved

somaz94:
Thanks @jerichokeyne — you were right, the environment issue is gone. `rhcs-e2e-setup` now succeeds in 21m59s and `rhcs-e2e-tests` passes in 12m41s, so the `invalid_grant` problem is fully resolved.

The one remaining red is unrelated to it, and I don't think it's this PR either. It fails in **teardown**, not setup or tests:

```
Step rhcs-e2e-setup     succeeded after 21m59s
Step rhcs-e2e-tests     succeeded after 12m41s
Step rhcs-e2e-teardown  failed    after 1h7m8s

[FAIL] Delete cluster [It] DestroyClusterByProfile
  tests/e2e/cluster_destroy_test.go:29
Ran 1 of 151 Specs -- 0 Passed | 1 Failed | 150 Skipped

failed to delete leaked security group '...-default-sg' after 8 attempts:
  still attached to network interfaces
Error: deleting EC2 Subnet: DependencyViolation
```

The same failure is on two other open PRs with the same signature — same spec, same `DependencyViolation`, same setup-pass/tests-pass/teardown-fail shape — and notably the failing HCP leg is a *different* one on each:

| PR | hcp-advanced | hcp-private |
|---|---|---|
| #1273 | fail | pass |
| #1300 | pass | fail |
| #1293 | pass | fail |

Both STS legs tear down cleanly in ~17m on all three, so this looks specific to the HCP cleanup path leaking a default SG onto ENIs. For what it's worth this PR only touches `logging/` plus `go.mod`/`go.sum`/`Makefile`, and none of those symbols appear anywhere in the build log.

Retrying just that leg rather than the whole suite:

/test e2e-presubmits-rosa-hcp-advanced-critical-high-presubmit

(`govulncheck` is also red on all three PRs — `GO-2026-5970` in `x/text` and `GO-2026-6061` in `grpc`, both transitive and untouched here. tide doesn't list it as a blocker.)

amandahla:
/override ci/prow/e2e-presubmits-rosa-hcp-advanced-critical-high-presubmit

